PR #49, opened 2022-01-26
Add missions endpoints

### LINKED ISSUE #35: Set up mission endpoints
In the [API-specification](https://equinor.github.io/flotilla/) the `/mission` and `/mission/{mission_id}` is used to get information about available missions on the asset. The information about available missions should be retrieved from Echo. Set up the endpoints using information from Echo.

vetlek:
Closes #35


backend/tests/test_api/test_missions_api.py
```diff
@@ -0,0 +1,84 @@
...
+    mock_response = Response()
+    mock_response.status_code = request_status_code
+    mock_response.json = lambda: request_json
+    echo_requests_mock = EchoRequestsMock()
+    echo_requests_mock.add_missions_responses([mock_response])
+
+    def get_echo_requests_mock():
+        yield echo_requests_mock
+
+    app.dependency_overrides[get_echo_requests] = get_echo_requests_mock
```
tsundvoll, line 42: Alternatively drop the helper function altogether and just use the EchoRequestsMock directly
```suggestion
    app.dependency_overrides[get_echo_requests] = EchoRequestsMock()
```

backend/tests/test_api/test_missions_api.py
```diff
@@ -0,0 +1,84 @@
...
+    mock_response = Response()
+    mock_response.status_code = request_status_code
+    mock_response.json = lambda: request_json
+    echo_requests_mock = EchoRequestsMock()
+    echo_requests_mock.add_mission_responses([mock_response])
+
+    def get_echo_requests_mock():
+        yield echo_requests_mock
+
+    app.dependency_overrides[get_echo_requests] = get_echo_requests_mock
```
tsundvoll, line 81: Same here, use helper function from echo_requests_mock, or use mock class directly

tsundvoll (changes requested):

backend/src/flotilla/echo/deserializers.py
```diff
@@ -0,0 +1,38 @@
...
+    sensor_type: List = json_reponse["sensorTypes"]
+    inspection_types: List[str] = []
+    if sensor_type:
+        for sensor in sensor_type:
+            try:
+                inspection_types.append(
+                    echo_sensor_to_inspection_type[sensor["sensorTypeKey"]]
+                )
+            except KeyError:
+                continue
```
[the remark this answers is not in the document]
vetlek: Main reason is that we haven't implemented/designed a logger scheme for the backend. Think that should be it's own issue. Might be that we want to notify the user that the mission contains an unsupported inspection type in some way, but that needs to be designed in the OpenAPI-spec.

backend/tests/test_api/test_missions_api.py
```diff
@@ -0,0 +1,84 @@
...
+    expected_status_code: int,
+):
+    mock_response = Response()
+    mock_response.status_code = request_status_code
+    mock_response.json = lambda: request_json
+    echo_requests_mock = EchoRequestsMock()
+    echo_requests_mock.add_missions_responses([mock_response])
+
+    def get_echo_requests_mock():
+        yield echo_requests_mock
```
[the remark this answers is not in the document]
vetlek: That mock is the default one when we only want the default return value. In the case where we want to give certain responses in some tests, we need to give new generator function. Tried to test with using the one in /mocks/echo_requests_mock.py and updating the object it generates without success. Not that familiar with using yield and generators, so there might be a solution.

backend/tests/test_api/test_missions_api.py
vetlek: I believe that it needs some sort of generator function. This is based on FastAPIs example for how to share a database in the different endpoints. The suggestion don't work since the dependency needs to be a callable function. Also tested adding `__call__` to EchoRequestMock and use the suggestion, but I got a 422 status code indicating that fastAPI is not able to handle the dependency.

backend/tests/test_api/test_missions_api.py
vetlek: See comment over.

backend/src/flotilla/echo/deserializers.py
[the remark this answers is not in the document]
tsundvoll: Agreed, reasonable to do this separately. Can you make an issue on it, so it is remembered?

backend/tests/test_api/test_missions_api.py
tsundvoll: My bad, I see the difference now

backend/tests/test_api/test_missions_api.py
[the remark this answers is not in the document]
tsundvoll: I see. I guess this is just how `dependency_overrides` in FastAPI works, so we can keep it like this for now, not a big problem

tsundvoll (approved):
LGTM, good job :+1:

PR #56, opened 2022-01-31
Dockerize frontend

oysand:


.dockerignore
```diff
@@ -0,0 +1,5 @@
+# Git
+.git/
+
+# Node
+frontend/node_modules
```
aestene, line 5: This should be put into the `.gitignore` file which is found in the `flotilla` folder. https://github.com/oysand/flotilla/blob/dockerize-frontend/frontend/.gitignore

aestene: Or this was a `.dockerignore`... 

Should the `.dockerignore` be together with the compose file in the root or together with the dockerfile in frontend?

oysand: This serves two different purposes. This is the .dockerignore

oysand: Just saw the second comment.. I think it makes sense to have it together with the compose file as there will be a second docker container for the backend, but the compose file ties them both together and we can use the same docker ignore for both

docker-compose.yml
```diff
@@ -0,0 +1,6 @@
+version: '3.7'
+services:
+  frontend:
+    build: frontend
+    ports:
+      - '3000:3000'
```
aestene, line 6: We should not run Flotilla on `3000` as that is the port used for ISAR. I'll make an issue to change port.

aestene: See #57

aestene (approved):
LGTM

Minor comments on `.dockerignore` but I think what you have done already might be correct.

PR #59, opened 2022-02-02
Change default port for frontend to 3001

### LINKED ISSUE #57: Change default port for Flotilla frontend
The port should be changed from `3000` as that is the default port of the ISAR application.

Instead port `3001` could be used.
- Update dockerfile for frontend to also use this port.
- Make the frontend application start by default on port `3000`.

NA

aestene:
Closes #57

### DIFF
--- a/frontend/.gitignore
+++ b/frontend/.gitignore
@@ -1,1 +1,2 @@
 build
+node_modules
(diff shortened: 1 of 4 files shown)


frontend/.gitignore
```diff
@@ -1 +1,2 @@
 build
+node_modules
```
oysand, line 2: Add new line to the end of the file

aestene: Done

oysand (approved):
LGTM

PR #63, opened 2022-02-03
Setup start and stop endpoints

### LINKED ISSUE #60: Setup start/stop missions endpoints
The backend should provide the frontend with endpoints for starting and stopping robots/missions.

Endpoints that forwards the request to start a robot with a given mission to the appropriate ISAR endpoint. Similarly for stop commands.

The potential to create harm with these endpoints are greater than the others. Different scope/authorization for making calls to these endpoints should be considered.

vetlek:
Closes #60


backend/src/flotilla/services/isar/requests.py
```diff
@@ -0,0 +1,53 @@
...
+    @abstractmethod
+    def start_mission(self, host: str, port: int, mission_id: int) -> Response:
+        pass
+
+    @abstractmethod
+    def stop(self, host: str, port: int) -> Response:
+        pass
+
+
+class IsarRequests(IsarRequestsInteface):
```
nicholasdalhaug, line 20: Sounds like a list of requests. Is this not just an API? Also a typo in "interface"
```suggestion
class IsarAPI(IsarAPIInterface):
```

backend/src/flotilla/services/request_handler.py
```diff
@@ -0,0 +1,166 @@
...
+    def get(
+        self,
+        url: str,
+        json_body=None,
+        request_timeout: float = settings.REQUEST_TIMEOUT,
+        auth: Optional[tuple] = None,
+        headers: Optional[dict] = None,
+        data: Optional[dict] = None,
+        params: Optional[dict] = None,
+        **kwargs,
```
nicholasdalhaug, line 69: I see we already use `**kwargs`, why not just use that for all?
```suggestion
        **kwargs,
```

backend/tests/mocks/isar_requests_mock.py
```diff
@@ -0,0 +1,57 @@
...
+        self.start_mission_responses: List[Response] = []
+        self.stop_responses: List[Response] = []
+
+        self.default_start_mission_reponse = default_start_mission_response()
+        self.default_stop_response = default_stop_response()
+
+    def start_mission(self, host: str, port: int, mission_id: int) -> Response:
+        if self.start_mission_responses:
+            return self.start_mission_responses.pop(0)
+        return self.default_start_mission_reponse
```
nicholasdalhaug, line 41:
```suggestion
        return self.default_start_mission_response
```

backend/tests/mocks/isar_requests_mock.py
```diff
@@ -0,0 +1,57 @@
...
+    def stop(self, host: str, port: int) -> Response:
+        if self.stop_responses:
+            return self.stop_responses.pop(0)
+        return self.default_stop_response
+
+    def add_start_mission_responses(self, responses: List[Response]) -> None:
+        self.start_mission_responses += responses
+
+    def add_stop_responses(self, responses: List[Response]) -> None:
+        self.stop_responses += responses
```
nicholasdalhaug, line 52: I do not see why we need this functionality in the mock. Is it not enough to mock `start` and `stop`?

nicholasdalhaug (changes requested):
Please tell me if you think I misunderstand something 😁

backend/src/flotilla/api/missions_api.py
```diff
@@ -56,6 +56,10 @@ async def read_missions(
             "model": ProblemDetails,
             "description": "Not found - The requested mission object does not exist",
         },
+        502: {
+            "model": ProblemDetails,
+            "description": "Not found - Error while contacting Echo",
+        },
```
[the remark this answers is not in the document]
vetlek: This is for documentation purposes (So the different responses shows up in the Swagger). Agree that the error messages should be the same though.

backend/src/flotilla/api/robots_api.py
```diff
@@ -30,8 +36,8 @@ def get_db():
     summary="List all robots on the asset.",
     dependencies=[Security(authentication_scheme)],
 )
-async def read_robots(response: Response, db: Session = Depends(get_db)) -> List[Robot]:
-    robots: List[RobotDB] = get_robots(db)
+async def get_robots(response: Response, db: Session = Depends(get_db)) -> List[Robot]:
+    robots: List[RobotDB] = read_robots(db)
     robots_api: List[Robot] = [robot.get_api_robot() for robot in robots]
```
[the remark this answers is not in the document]
vetlek: I think the names should correspond to CRUD operations and the get_* should be reserved for the API-function.

backend/src/flotilla/api/missions_api.py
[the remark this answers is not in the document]
aestene: I agree, the same goes for the other endpoint above. There should be a way to only have the string change in one location.

backend/src/flotilla/settings/settings.py
```diff
@@ -14,6 +14,9 @@ class Settings(BaseSettings):
     ECHO_APP_SCOPE: str = Field(default="", env="ECHO_APP_SCOPE")
     ECHO_API_URL: str = Field(default="", env="ECHO_API_URL")
     INSTALLATION_CODE: str = Field(default="", env="INSTALLATION_CODE")
+    REQUEST_TIMEOUT: float = Field(default=5, env="REQUEST_TIMEOUT")
```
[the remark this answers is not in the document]
vetlek: The default is used if the variable is not in the .env file, so it is not a case of two different defaults.

backend/src/flotilla/api/missions_api.py
```diff
@@ -70,12 +74,14 @@ async def read_single_mission(
     try:
         echo_response: RequestResponse = echo_requests.get_mission(mission_id)
     except RequestException:
-        response.status_code = 404
-        return ProblemDetails(title="Not found - Could not contact echo", status=404)
+        response.status_code = 502
```
aestene, line 77: Instead of writing status codes by writing the int directly an enum should be used. This goes for both in code and documentation decorators.

Use this library: https://docs.python.org/3/library/http.html
image

backend/src/flotilla/services/request_handler.py
vetlek: It is nice to see the different available parameters when you want to use the function.

backend/src/flotilla/api/robots_api.py
```diff
@@ -46,14 +52,124 @@ async def read_robots(response: Response, db: Session = Depends(get_db)) -> List
...
+            "description": "Failed to start - Client is not authorized to start mission",
+        },
+        409: {
+            "model": ProblemDetails,
+            "description": "Failed to start - Conflict with the robot schedule",
+        },
+        404: {
+            "model": ProblemDetails,
+            "description": "Not Found - No robot with given id",
+        },
```
aestene, line 83: Should be sorted according to status number.

backend/src/flotilla/api/robots_api.py
```diff
@@ -46,14 +52,124 @@ async def read_robots(response: Response, db: Session = Depends(get_db)) -> List
...
+)
+async def post_start_robot(
+    response: Response,
+    robot_id: int = Path(None, description=""),
+    mission_id: int = Path(None, description=""),
+    db: Session = Depends(get_db),
+    isar_requests: IsarRequests = Depends(get_isar_requests),
+) -> StartResponse:
+    """### Overview Start a mission with given id using robot with robot id."""
+    robot: RobotDB = read_robot_by_id(db, robot_id)
```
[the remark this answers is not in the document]
aestene: Yes, rather throw an exception in `read_robot_by_id`

backend/tests/mocks/isar_requests_mock.py
vetlek: So you can be able to add your custom responses if you want to test how the api response to a particular response from ISAR

backend/tests/mocks/isar_requests_mock.py
```diff
@@ -0,0 +1,57 @@
...
+    return response
+
+
+class IsarRequestsMock(IsarRequestsInteface):
+    def __init__(self) -> None:
+        self.start_mission_responses: List[Response] = []
+        self.stop_responses: List[Response] = []
+
+        self.default_start_mission_reponse = default_start_mission_response()
+        self.default_stop_response = default_stop_response()
```
aestene, line 36:
```suggestion
        self.default_start_mission_response: Response = default_start_mission_response()
        self.default_stop_response: Response = default_stop_response()
```

backend/src/flotilla/api/robots_api.py
```diff
@@ -46,14 +52,124 @@ async def read_robots(response: Response, db: Session = Depends(get_db)) -> List
...
+        )
+    except RequestException:
+        response.status_code = 502
+        return ProblemDetails(title="Error sending request to ISAR")
+    response_isar_json: dict = response_isar.json()
+    if response_isar.status_code != 200:
+        response.status_code = response_isar.status_code
+        return ProblemDetails(
+            title=f"Could not start mission. Reason: {response_isar_json['message']}"
+        )
```
aestene, line 118: This should not be necessary. If you look at the `RequestHandler` it uses a function `response.raise_for_status()`. This will raise an error for any status code outside the 200 range and as such we should not get to this if statement as it will have been raised as a `RequestException` in `RequestHandler`

aestene (changes requested):
See comments

backend/src/flotilla/api/robots_api.py
vetlek: Should we change this behavior? Seems like we can loose a lot of useful information from using the raise_from_status, like if there is a conflict. Don't know if we would handle it any different, but might want to pass the information about the particular issue?

backend/src/flotilla/services/isar/requests.py
```diff
@@ -0,0 +1,53 @@
...
+        self.credentials: DefaultAzureCredential = get_azure_credentials()
+
+    def start_mission(self, host: str, port: int, mission_id: int) -> Response:
+        token: str = self.credentials.get_token(self.request_scope).token
+        url: str = f"http://{host}:{port}/schedule/start-mission"
+        params: dict = {"ID": mission_id}
+        response: Response = self.request_handler.post(
+            url=url, headers={"Authorization": f"Bearer {token}"}, params=params
+        )
+        return response
```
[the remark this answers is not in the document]
vetlek: Handled since we raise error when status code is not in 200 range

backend/src/flotilla/api/robots_api.py
Christdej, line 83: Remove all responses except 200, we dont need this at the moment

backend/src/flotilla/api/robots_api.py
```diff
@@ -46,14 +52,124 @@ async def read_robots(response: Response, db: Session = Depends(get_db)) -> List
...
+            "description": "Failed to stop - Client is not authorized to stop mission",
+        },
+        408: {
+            "model": ProblemDetails,
+            "description": "Failed to stop - Request Timeout while stopping mission",
+        },
+        502: {
+            "model": ProblemDetails,
+            "description": "Failed to stop - Error while contacting ISAR",
+        },
```
Christdej, line 144: Remove all responses except 200, we dont need this at the moment

Christdej (changes requested):

backend/src/flotilla/api/robots_api.py
aestene: The error message itself is kept as we use `raise error from e`. As such we should only need to log the error to have it available internally. Typically you would like to not send all the error information to the frontend but rather log it in the backend itself. As such I think this should be fine as we are saving the information in the error itself.
https://github.com/equinor/flotilla/pull/63/files#diff-c5827550d464dc4af6f269546e3cd1aae61565d50740f7837449c79f12ab8436R54-R58

backend/src/flotilla/api/robots_api.py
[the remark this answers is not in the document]
vetlek: Leave this to #74 since this needs to be done other places that is not in the scope of this PR.

backend/src/flotilla/api/robots_api.py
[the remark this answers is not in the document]
vetlek: The naming issue is fixed temporarily in this file for now. Added an issue about renaming for all database models #76

backend/src/flotilla/api/robots_api.py
vetlek: Think the try catch around the raise_for_status call should be removed and let the error be handled in the controller.

vetlek: But this should probably be done in a separate issue.

aestene: Sure, that may be more correct for this case.

aestene (approved):
LGTM. Check with @nicholasdalhaug and @Christdej if they are happy with the review before merging.

backend/src/flotilla/settings/settings.py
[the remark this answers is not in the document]
nicholasdalhaug: I do not see why we need all these environment variables, but I can ask that in an other setting.

backend/tests/test_api/test_missions_api.py
```diff
@@ -57,9 +57,9 @@ def get_echo_requests_mock():
 @pytest.mark.parametrize(
     "mission_id, request_status_code, request_json, expected_status_code",
     [
-        (84, 200, mission_success_json, 200),
-        (666, 404, dict(), 404),
-        (54, 200, mission_error_json, 404),
+        (84, HTTPStatus.OK.value, mission_success_json, HTTPStatus.OK.value),
+        (666, HTTPStatus.NOT_FOUND.value, dict(), HTTPStatus.BAD_GATEWAY.value),
+        (54, HTTPStatus.OK.value, mission_error_json, HTTPStatus.NOT_FOUND.value),
     ],
```
nicholasdalhaug, line 63: The first numbers here are pretty magical, where are they from?

nicholasdalhaug (changes requested):
Still one comment from before that is still not answered: [link](https://github.com/equinor/flotilla/pull/63#discussion_r800747191)

backend/src/flotilla/services/isar/requests.py
vetlek: Hmm, don't know if feel like API is any better

nicholasdalhaug (approved):

PR #81, opened 2022-02-09
Add DBModel suffix to db models

### LINKED ISSUE #76: Change name of database models to use DB prefix
There is an overlap of the names between the models used in the API interface and the database. This leads to conflict when both models are used in the same file. A solution to this is to add a DB prefix to the database models to make it clear that the models originated from the database and that all other models is from the API.

This will make the code easier to read and less confusing when both models are used.

No changes

vetlek:
Closes #76


aestene (approved):
LGTM

PR #86, opened 2022-02-11
Add report api

anetteu:


backend/src/flotilla/api/reports_api.py
```diff
@@ -0,0 +1,71 @@
...
+    tags=["Reports"],
+    summary="List all available reports on the asset",
+    dependencies=[Security(authentication_scheme)],
+)
+async def get_reports(
+    db: Session = Depends(get_db),
+) -> List[Report]:
+    """List all available reports on the asset."""
+    db_reports: List[ReportDBModel] = read_reports(db)
+    logger.info(f"{db_reports}")
```
vetlek, line 39: Probably not necessary to log both the reports in the database and the returned reports. Suggest that both are removed until we see the need

vetlek (changes requested):
Looks good. One small change

backend/src/flotilla/api/reports_api.py
anetteu: You are right, this is dev logging that I forgot to remove

vetlek (approved):
LGTM

PR #101, opened 2022-03-14
Improvement - use http exception for request errors

GodVenn:
Make use of FastAPI's HTTPExceptions for exception handling of everything HTTP related.
FastAPI catches these in the framework and returns an appropriate response to the end-user.
This reduces the amount of except clauses we need for all request operations.  

-----
~~This change causes all tests making use of the "TestClient" to fail. Have to look into this~~
The isar and echo service mockers are duplicating behavior that exists in the request handler, such that if a change is made in the request handler, the tests may fail simply because the duplication is no longer up-to-speed.
Instead of duplicating this behavior, a better solution could be to mock the lowest level function call, the "requests.request()" call in the request_handler.


anetteu (approved):
Good work! Is there a reason why the tracebacks are no longer logged?

GodVenn:
> Good work! Is there a reason why the tracebacks are no longer logged?

@anetteu Thanks! The tracebacks are logged when there is an unhandled exception, so logging them when we handle the exceptions may be misleading and would also clutter the log with a lot of useless lines of information. 
It was something I did for consistency, but I don't know what the idea was when it was implemented so if you disagree, I can change it back to logging full stacktraces!

anetteu:
> > Good work! Is there a reason why the tracebacks are no longer logged?
> 
> @anetteu Thanks! The tracebacks are logged when there is an unhandled exception, so logging them when we handle the exceptions may be misleading and would also clutter the log with a lot of useless lines of information. It was something I did for consistency, but I don't know what the idea was when it was implemented so if you disagree, I can change it back to logging full stacktraces!

I agree that the log is cleaner and that logging stacktraces are probably not necessary when the exceptions are handled. I was just curious about your thoughts :smile: 

This is a clear improvement - well done!

PR #114, opened 2022-03-31
Ïntroduce multithreading applications to run api and event handler in seperate treads

oysand:


backend/main.py
```diff
@@ -0,0 +1,31 @@
...
+    threads: List[Thread] = []
+
+    populate_mock_db(session=SessionLocal(), connection=connection, base=Base)
+
+    api_thread: Thread = Thread(
+        target=run_app,
+        name="Flotilla Backend API",
+        daemon=True,
+    )
+
```
anetteu, line 19: remove whiteline

backend/main.py
```diff
@@ -0,0 +1,31 @@
...
+    threads.append(api_thread)
+
+    event_handler_thread = Thread(
```
anetteu, line 22:
```suggestion
    event_handler_thread: Thread = Thread(
```

aestene (approved):
LGTM

aestene:
Suggestion for documentation purposes would be to change the title of the PR as this is more about introducing multithreaded applications than adding an event handler.

PR #124, opened 2022-04-05
Add pagination for certain GET requests

### LINKED ISSUE #102: Add pagination to certain GET requests
When receiving a GET requests for multiple entries we should restrict the amount returned using a pagination system with default parameters that may be overwritten in the query. The parameter for page size should have an upper limit.  
This is in accordance with the RESTful API principles.

This will throttle our API and protect it from being overloaded in the future when we might have a large amount of data being retrieved by these calls.   
Better to implement it now in anticipation as it might be easy to forget in the future as the database will probably grow slowly.

This improvement should make our API more reliable in the long run and possibly prevent downtime.

anetteu:
Resolves #102

### DIFF
--- a/backend/src/flotilla/database/crud.py
+++ b/backend/src/flotilla/database/crud.py
@@ -12,10 +12,15 @@ from flotilla.database.models import (
     ReportStatus,
     RobotDBModel,
 )
+from flotilla.settings import settings
 
 
-def read_robots(db: Session) -> List[RobotDBModel]:
-    robots: List[RobotDBModel] = db.query(RobotDBModel).all()
+def read_robots(
+    db: Session, page: int = settings.PAGE, page_size: int = settings.PAGE_SIZE
+) -> List[RobotDBModel]:
+    robots: List[RobotDBModel] = (
+        db.query(RobotDBModel).offset(page * page_size).limit(page_size).all()
+    )
     return robots
 
 
@@ -31,8 +36,12 @@ def read_robot_by_id(db: Session, robot_id: int) -> RobotDBModel:
     return robot
 
 
-def read_reports(db: Session) -> List[ReportDBModel]:
-    reports: List[ReportDBModel] = db.query(ReportDBModel).all()
+def read_reports(
+    db: Session, page: int = settings.PAGE, page_size: int = settings.PAGE_SIZE
+) -> List[ReportDBModel]:
+    reports: List[ReportDBModel] = (
+        db.query(ReportDBModel).offset(page * page_size).limit(page_size).all()
+    )
     return reports
 
 
@@ -48,8 +57,12 @@ def read_report_by_id(db: Session, report_id: int) -> ReportDBModel:
     return report
 
 
-def read_events(db: Session) -> List[EventDBModel]:
-    events: List[EventDBModel] = db.query(EventDBModel).all()
+def read_events(
+    db: Session, page: int = settings.PAGE, page_size: int = settings.PAGE_SIZE
+) -> List[EventDBModel]:
+    events: List[EventDBModel] = (
+        db.query(EventDBModel).offset(page).limit(page_size).all()
+    )
     return events
 
 
--- a/backend/src/flotilla/database/crud.py
+++ b/backend/src/flotilla/database/crud.py
@@ -13,7 +13,7 @@ from flotilla.database.models import EventDBModel, ReportDBModel, ReportStatus
 T = TypeVar("T", bound=Base)  # Generic type for db models
 
 
-def get_list_paginated(
+def read_list_paginated(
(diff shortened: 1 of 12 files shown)


backend/src/flotilla/database/db.py
```diff
@@ -2,8 +2,9 @@
...
 from sqlalchemy.orm import scoped_session, sessionmaker
 
-# TODO: Move to config
-SQLALCHEMY_DATABASE_URL = "sqlite:///:memory:"
+from flotilla.settings import settings
+
+SQLALCHEMY_DATABASE_URL = settings.SQLALCHEMY_DATABASE_URL
 
 connection = create_engine(
     SQLALCHEMY_DATABASE_URL, connect_args={"check_same_thread": False}
```
vetlek, line 10: Use settings.SQLALCHEMY_DATABASE_URL directly in the create_engine

vetlek (changes requested):
One minor change

backend/src/flotilla/settings/settings.py
```diff
@@ -17,6 +17,11 @@ class Settings(BaseSettings):
     REQUEST_TIMEOUT: float = Field(default=5, env="REQUEST_TIMEOUT")
     ISAR_CLIENT_ID: str = Field(default="", env="ISAR_CLIENT_ID")
     ISAR_APP_SCOPE: str = Field(default="", env="ISAR_APP_SCOPE")
+    SQLALCHEMY_DATABASE_URL: str = Field(
+        default="sqlite:///:memory:", env="SQLALCHEMY_DATABSE_URL"
+    )
+    SKIP: int = Field(default=0, env="SKIP")
+    LIMIT: int = Field(default=100, env="LIMIT")
```
GodVenn, line 24: We should also have a MAX_LIMIT to prevent the user from overriding the pagination alltogether

backend/src/flotilla/database/crud.py
```diff
@@ -47,8 +54,10 @@ def read_report_by_id(db: Session, report_id: int) -> ReportDBModel:
     return report
 
 
-def read_events(db: Session) -> List[EventDBModel]:
-    events: List[EventDBModel] = db.query(EventDBModel).all()
+def read_events(
+    db: Session, skip: int = settings.SKIP, limit: int = settings.LIMIT
+) -> List[EventDBModel]:
+    events: List[EventDBModel] = db.query(EventDBModel).offset(skip).limit(limit).all()
```
GodVenn, line 60: Consider changing the query parameter names to the more intuitive "page" & "pageSize"/"count"

GodVenn: The input should also be validated, for example the skip/page and lmit/count should both be integers >= 0, and as mentioned in another commend the count/limit <= MAX_LIMIT

backend/src/flotilla/database/crud.py
```diff
@@ -47,8 +56,12 @@ def read_report_by_id(db: Session, report_id: int) -> ReportDBModel:
     return report
 
 
-def read_events(db: Session) -> List[EventDBModel]:
-    events: List[EventDBModel] = db.query(EventDBModel).all()
+def read_events(
+    db: Session, page: int = settings.PAGE, page_size: int = settings.PAGE_SIZE
+) -> List[EventDBModel]:
+    events: List[EventDBModel] = (
+        db.query(EventDBModel).offset(page).limit(page_size).all()
```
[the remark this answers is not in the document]
GodVenn: I found a way to generalize the pagination functionality, we can look at it when you have time

backend/src/flotilla/database/crud.py
```diff
@@ -1,68 +1,37 @@
...
-
-def read_reports(db: Session) -> List[ReportDBModel]:
-    reports: List[ReportDBModel] = db.query(ReportDBModel).all()
-    return reports
-
-
-def read_report_by_id(db: Session, report_id: int) -> ReportDBModel:
-    report: Optional[ReportDBModel] = (
-        db.query(ReportDBModel).filter(ReportDBModel.id == report_id).first()
+def get_list_paginated(
```
vetlek, line 16:
```suggestion
def read_list_paginated(
```

backend/src/flotilla/database/crud.py
```diff
@@ -105,7 +74,7 @@ def create_event(
 
 
 def remove_event(db: Session, event_id: int) -> None:
```
vetlek, line 76: Not this PR, but this should be `delete_event` to follow CRUD name convention

vetlek (commented):
Some minor things

backend/src/flotilla/database/crud.py
anetteu: This function is named `remove_event` because it is used in another function called `delete_event`

vetlek (approved):
LGTM

PR #129, opened 2022-04-11
Add check for conflicting events before creation

### LINKED ISSUE #95: Add check for time conflicts for the post event endpoint
Currently there is no check in the post event endpoint if an event is already scheduled in the time-slot that the user want to schedule the event. This should be avoided, and the user should be notified that an event is already scheduled for that time period.  

This will change the existing functionality by making it possible for the backend to reject posted events in case of conflicts.

No change.

vetlek:
Closes #95

### DIFF
--- a/backend/src/flotilla/api/events_api.py
+++ b/backend/src/flotilla/api/events_api.py
@@ -1,3 +1,4 @@
+from datetime import timedelta
 from http import HTTPStatus
 from logging import getLogger
 from typing import List
@@ -12,6 +13,7 @@ from flotilla.database.crud import (
     create_event,
     read_event_by_id,
     read_events,
+    read_events_by_robot_id_and_time_span,
     remove_event,
 )
 from flotilla.database.db import get_db
@@ -21,8 +23,7 @@ logger = getLogger("api")
 
 router = APIRouter()
 
-NOT_FOUND_DESCRIPTION = "Not Found - No event with given id"
-INTERNAL_SERVER_ERROR_DESCRIPTION = "Internal Server Error"
+DEFAULT_EVENT_DURATION = timedelta(hours=1)
 
 
 @router.get(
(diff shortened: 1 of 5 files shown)


tsundvoll (changes requested):

tsundvoll:
I think this looks good. When my comments above are solved I will approve.

backend/src/flotilla/api/events_api.py
```diff
@@ -62,11 +63,24 @@ async def post_event(
...
+        overlapping_events: List[Event] = read_events_by_robot_id_and_time_span(
+            db=db,
+            robot_id=event_request.robot_id,
+            start_time=event_request.start_time,
+            end_time=end_time,
+        )
+        if overlapping_events:
+            raise HTTPException(
+                status_code=HTTPStatus.CONFLICT.value,
+                detail=f"Conflict with already existing event in the same time period. Events with id: {','.join(str(event.id) for event in overlapping_events)}",
```
GodVenn, line 76: Will all users be authorized to see all events for a given robot?
If not, this might be giving out too much information, maybe the first statement is enough?
I'm not sure, open for discussion

backend/src/flotilla/api/events_api.py
```diff
@@ -62,11 +63,24 @@ async def post_event(
 ) -> Event:
     """Add a new event to the robot schedule"""
     try:
+        end_time = event_request.start_time + DEFAULT_EVENT_DURATION
```
GodVenn, line 66: Will we always use the default duration? Maybe the user should have an event_duration optional filed to this post endpoint so they can define it themselves?

vetlek: Currently a default value is enough in my opinion. I don't think the user should be able to set the event_duration, but rather be based on the number of tags and/or based on data from previous missions with the same id. Can make a card for setting the event_duration based on number of tags.

tsundvoll (approved):
LGTM :+1:

PR #137, opened 2022-04-20
Add endpoint to create and enable robots

### LINKED ISSUE #122: Add endpoints to create, activate and deactivate robots
In the case of a new robot with corresponding ISAR instance being added there should be an option for an admin user to add this to the database through an endpoint in the API. In addition it should be possible to activate and deactivate robots that should not be considered in use on the facility.

This will be new functionality.

The authorization for an admin role has not been implemented yet such that a first iteration of this issue will be to enable the endpoint without admin access. Keep in mind that admin restrictions should be added later.

aestene:
Closes #122

### DIFF
--- a/backend/src/flotilla/api/events_api.py
+++ b/backend/src/flotilla/api/events_api.py
@@ -142,7 +142,6 @@ async def delete_event(
     dependencies=[Security(authentication_scheme)],
 )
 async def get_event(
-    response: Response,
     db: Session = Depends(get_db),
     event_id: int = Path(None, description=""),
 ) -> Event:
--- a/backend/src/flotilla/api/events_api.py
+++ b/backend/src/flotilla/api/events_api.py
@@ -4,6 +4,7 @@ from logging import getLogger
 from typing import List
 
 from fastapi import APIRouter, Body, Depends, HTTPException, Path, Response, Security
+from flotilla_openapi.models.error import Error
 from flotilla_openapi.models.event import Event
 from flotilla_openapi.models.event_request import EventRequest
 from pytest import Session
@@ -34,6 +35,14 @@ DEFAULT_EVENT_DURATION = timedelta(hours=1)
             "model": List[Event],
             "description": "Request successful.",
         },
+        HTTPStatus.UNAUTHORIZED.value: {
+            "model": Error,
+            "description": "Unauthorized",
+        },
+        HTTPStatus.NOT_FOUND.value: {
+            "model": Error,
+            "description": "Not Found",
+        },
     },
     tags=["Events"],
     summary="Lookup events",
@@ -58,6 +67,18 @@ async def get_events(
     "/events",
     responses={
         HTTPStatus.CREATED.value: {"model": Event, "description": "Request successful"},
+        HTTPStatus.UNAUTHORIZED.value: {
+            "model": Error,
+            "description": "Unauthorized",
+        },
+        HTTPStatus.NOT_FOUND.value: {
+            "model": Error,
+            "description": "Not Found",
+        },
+        HTTPStatus.CONFLICT.value: {
+            "model": Error,
+            "description": "Conflict",
+        },
     },
     tags=["Events"],
     summary="Create new event",
@@ -105,6 +126,14 @@ async def post_event(
     "/events/{event_id}",
     responses={
         HTTPStatus.NO_CONTENT.value: {"description": "Event successfully deleted"},
+        HTTPStatus.UNAUTHORIZED.value: {
+            "model": Error,
+            "description": "Unauthorized",
+        },
+        HTTPStatus.NOT_FOUND.value: {
+            "model": Error,
+            "description": "Not Found",
+        },
     },
     tags=["Events"],
     summary="Delete event with specified id",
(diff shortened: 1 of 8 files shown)


backend/src/flotilla/api/events_api.py
```diff
@@ -33,9 +34,20 @@
             "model": List[Event],
             "description": "Request successful.",
         },
+        HTTPStatus.UNAUTHORIZED.value: {
+            "model": Error,
```
vetlek, line 38: The description of return codes can be misleading since the Error model is not returned when we throw the HTTPException. Might be wrong

aestene: What is returned is an `HTTPException` which leads to a dictionary showing this

```
{
    "detail": "some text here"
}
```

Which is what the `Error` object indicates. Not certain what else to put there as you can't indicate an `HTTPException`...

backend/tests/test_api/test_robots_api.py
```diff
@@ -15,6 +19,84 @@ def test_get_robots(test_app: FastAPI):
...
+                "host": "localhost",
+                "port": "This is not an integer and should cause validation error",
+                "enabled": True,
+                "status": "available",
+                "capabilities": ["Image", "ThermalImage"],
+            },
+        ),
+    ],
+)
+def test_create_robot_throws_exception(
```
vetlek, line 91: Thin the name can be a bit misleading. I expect that the test throws an exception and that we use pytest.raises when the name of the test is ...throws_exception. Perhaps calling it something with invalid input is more representative

backend/tests/test_api/test_robots_api.py
```diff
@@ -15,6 +19,84 @@ def test_get_robots(test_app: FastAPI):
         assert len(response.json()) == 2
 
 
+def test_create_robot_success(test_app: FastAPI, session: Session):
+    body: dict = {
+        "id": 0,
+        "name": "test_robot",
```
vetlek, line 25: Comment here since it is most obvious here. I think it is a bit weird that we give an id in the POST /robots request when this is not used in the database and thus is not valid. I think we should consider different robot models for the POST /robots endpoint and the GET /robots... endpoints to better represent what information that is actually used. I also think it is unnecessary to publish host and port in the GET /robots endpoints, which would require different models for POST /robots anyway.

backend/src/flotilla/database/crud.py
```diff
@@ -31,6 +34,64 @@ def read_robot_by_id(db: Session, robot_id: int) -> RobotDBModel:
...
+
+def _add_capabilities_to_robot(robot: RobotDBModel, capabilities: List[str]) -> None:
+    for capability in capabilities:
+        if not InspectionType.has_value(capability):
+            raise HTTPException(
+                status_code=HTTPStatus.BAD_REQUEST.value,
+                detail="Invalid format for robot capabilities",
+            )
+
+        CapabilityDBModel(robot=robot, capability=InspectionType(capability))
```
vetlek, line 76: Does this work? Thought the CapabililityDBModel wanted robot_id and not the whole robot. Some sort of sqlalchamy magic?

backend/src/flotilla/api/robots_api.py
```diff
@@ -139,3 +226,40 @@ async def post_stop_robot(
...
+    robot_id: int = Path(None, description=""),
+    enable: bool = Query(None, description=""),
+    db: Session = Depends(get_db),
+):
+    robot: RobotDBModel = read_robot_by_id(db=db, robot_id=robot_id)
+    robot.enabled = enable
+
+    db.commit()
+
+    return {"detail": "Success"}
```
vetlek, line 265: Why don't we use the Success model here?

backend/src/flotilla/api/robots_api.py
```diff
@@ -41,20 +60,70 @@ async def get_robots(db: Session = Depends(get_db)) -> List[Robot]:
...
+            host=robot.host,
+            port=robot.port,
+            enabled=robot.enabled,
+            capabilities=robot.capabilities,
+        )
+    except HTTPException as e:
+        logger.exception(f"An error occurred while creating a robot: {e.detail}")
+        raise
+
+    return {"robot_id": robot_id}
```
vetlek, line 101: Why list a model in the description when it is not used here? Guess they are the same, but a bit weird

vetlek (changes requested):
Think we should use separate robot models in the API for POST and GET

backend/tests/test_api/test_robots_api.py
aestene: Renamed to `test_create_robot_fails_with_invalid_input`

backend/tests/test_api/test_robots_api.py
aestene: I agree! I have seen this before but forgot to look into it.

aestene: I agree! I have seen this before but forgot to look into it.

As for publishing the host and port through the GET endpoints I don't really see why not? It is static metadata for the robot that is associated with the object. It is not required for the frontend to know but I don't think we should design the data part of the API to just serve the specific frontend we have now.

backend/src/flotilla/database/crud.py
aestene: Yes, it does work. However, I found an issue when attempting to do the same in some locations in the mock database setup. I'll experiment a bit more first.

backend/tests/test_api/test_robots_api.py
aestene: If we wish to create a different model for GET and POST we need to make adjustments to the openapi specification too.

What do you think about accepting this for now and setting this change as an issue for the workshop next week? It would be nice to have closed this PR before the workshop to avoid merge conflicts etc. as it changes quite a lot of things.

vetlek: Thats fine

aestene: #149

vetlek (approved):

PR #151, opened 2022-04-25
Event handler functionality

oysand:


anetteu:
@oysand I suggest that we close this as we are no longer developing the python backend.

anetteu:
The implementation of event handler functionality supporting dotnet is done in #196. Closing this issue as we are not doing more development on the Python backend.

PR #181, opened 2022-04-29
Set up unit testing in dotnet

GodVenn:
Closes #172

PR #182, opened 2022-05-02
Set up formatting in dotnet

### LINKED ISSUE #168: Set up formatting

GodVenn:
Closes #168 
~~Included #181 to avoid too many conflicts, will rebase once #181 is merged~~

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,509 @@
+# Docs: https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference
+
+# top-most EditorConfig file
+root = true
+
+# Don't use tabs for indentation.
+[*]
+indent_style = space
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+# Code files
+[*.{cs,csx,vb,vbx}]
+indent_size = 4
+charset = utf-8-bom
+
+# Xml project files
+[*.{csproj,vbproj,vcxproj,vcxproj.filters,proj,projitems,shproj}]
+indent_size = 4
+
+# Xml config files
+[*.{props,targets,ruleset,config,nuspec,resx,vsixmanifest,vsct,xml,stylecop}]
+indent_size = 2
+
+# JSON files
+[*.{json,json5}]
+indent_size = 2
+
+# YAML Files
+[*.{yml,yaml}]
+indent_size = 2
+
+#### .NET Coding Conventions ####
+[*.{cs,csx,cake,vb}]
+
+# Organize usings
+dotnet_separate_import_directive_groups = false
+dotnet_sort_system_directives_first = true
+
+# Avoid "this." and "Me." if not necessary
+dotnet_style_qualification_for_event = false : warning
+dotnet_style_qualification_for_field = false : warning
+dotnet_style_qualification_for_method = false : warning
+dotnet_style_qualification_for_property = false : warning
+
+# Use language keywords instead of framework type names for type references
+dotnet_style_predefined_type_for_locals_parameters_members = true : warning
+dotnet_style_predefined_type_for_member_access = true : warning
+
+# Parentheses preferences
+dotnet_style_parentheses_in_arithmetic_binary_operators = always_for_clarity:warning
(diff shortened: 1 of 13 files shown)


.editorconfig
```diff
@@ -0,0 +1,506 @@
...
+dotnet_diagnostic.MA0091.severity = warning    # MA0091: Sender should be 'this' for instance events
+dotnet_diagnostic.MA0092.severity = warning    # MA0092: Sender should be 'null' for static events
+dotnet_diagnostic.MA0093.severity = warning    # MA0093: EventArgs should not be null
+dotnet_diagnostic.MA0094.severity = warning    # MA0094: A class that provides CompareTo(T) should implement IComparable<T>
+dotnet_diagnostic.MA0095.severity = warning    # MA0095: A class that implements IEquatable<T> should override Equals(object)
+dotnet_diagnostic.MA0096.severity = warning    # MA0096: A class that implements IComparable<T> should also implement IEquatable<T>
+dotnet_diagnostic.MA0097.severity = warning    # MA0097: A class that implements IComparable<T> or IComparable should override comparison operators
+dotnet_diagnostic.MA0098.severity = suggestion # MA0098: Use indexer instead of extension methods
+dotnet_diagnostic.MA0099.severity = warning    # MA0099: Use Explicit enum value instead of 0
+MA0053.public_class_should_be_sealed = false
```
Christdej, line 506: Add newline

.editorconfig
```diff
@@ -0,0 +1,506 @@
+# Docs: https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference
```
Christdej, line 1: Are all these options needed? Seems like many of they are default, just feel like it is a bit overwhelming. But if this is what is needed to have good formatting and linting, sure.

backend/README.md
```diff
@@ -14,4 +14,26 @@ To unit test the backend, run the following command in the backend folder:
...
+
+We use [dotnet format](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-format)
+to format and verify code style in backend based on the 
+[C# coding conventions](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions).  
+
+Dotnet format is included in the .NET6 SDK.
+
+To check the formating, run 
+```
+cd backend
```
Christdej, line 31: dont we need to be in the backend/api folder?

Christdej: This also goes for the `dotnet run` command above and for test you need to be in backend/test no?

backend/api/api.csproj
```diff
@@ -14,4 +15,8 @@
     <PackageReference Include="Swashbuckle.AspNetCore" Version="6.2.3" />
   </ItemGroup>
 
+  <ItemGroup>
+    <EditorConfigFiles Remove="C:\Users\tlne\source\repos\flotilla\backend\api\.editorconfig" />
```
Christdej, line 19: This seems like something that should not be here?

Christdej (changes requested):

backend/README.md
GodVenn: No, the solution file gives dotnet the info it needs, so we can run both commands from the backend folder

backend/api/api.csproj
GodVenn: Yes, will fix

.editorconfig
GodVenn: I am not sure right now which ones are still default, but in my opinion they are nicely organized and have expressive names such that it will be easy to modify should we find that we have different needs or preferences in the future.
It also gives us an overview of all the options that we have available for controlling the code style and formatting rules

backend/README.md
Christdej: Ah, that doesnt work for VSCode (maybe I need an extension or something), but Im fine with that

GodVenn: It fails to run the command in the backend directory? I thought the solution file would give the information even outside visual studio

Christdej (approved):
Fix the broken pipeline then you can merge

PR #194, opened 2022-05-10
Add mission controller endpoints

aestene:
The mission controller endpoints use the EchoService to communicate
with Echo and retrieve missions. Currently, missions are served from
Echo but will eventually be added to the database.


backend/api/Services/EchoService.cs
```diff
@@ -0,0 +1,182 @@
...
+            var uri = ConstructGetMissionsRequestUri();
+
+            var response = await client.GetAsync(uri);
+            if (!response.IsSuccessStatusCode) throw new MissionNotFoundException($"Failed to retrieve missions from Echo: {response.ReasonPhrase}");
+
+            var echoMissions = await response.Content.ReadFromJsonAsync<List<EchoMissionResponse>>();
+
+            if (echoMissions is null) throw new JsonException("Failed to deserialize missions from Echo");
+
+            var missions = ProcessEchoMissions(echoMissions);
```
aestene, line 53:
```suggestion
            var missions = ProcessEchoMissions(echoMissions?);
```

backend/api/Models/Enums/InspectionType.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        return sensorType switch
+        {
+            "Picture" => Image,
+            "ThermicPicture" => ThermalImage,
+            "Audio" => Audio,
+            _ => Image,
+        };
+    }
+}
```
Christdej, line 21: Convert to Enum, string conversion is taken care automatically by:
`builder.Services.AddControllers().AddJsonOptions(options => options.JsonSerializerOptions.Converters.Add(new JsonStringEnumConverter()));`

backend/api/Program.cs
```diff
@@ -1,16 +1,35 @@
...
+});
+
 builder.Services.AddDbContext<FlotillaDbContext>(options => options.UseInMemoryDatabase("flotilla"));
+
+builder.Services.AddSingleton<DefaultAzureCredential>();
+
+builder.Services.AddSingleton<KeyVaultService>();
+
 builder.Services.AddScoped<RobotService>();
+
```
Christdej, line 28: No need for empty line between services we have made imo, but I like the empty line between different types of services

backend/api/Services/EchoService.cs
```diff
@@ -0,0 +1,182 @@
...
+using Api.Models;
+using Azure.Core;
+using Azure.Identity;
+
+namespace Api.Services
+{
+    public class EchoService
+    {
+        private readonly ILogger<EchoService> _logger;
+
```
Christdej, line 13: No need for empty line between private class properties imo, open for discussion

backend/api/appsettings.Development.json
```diff
@@ -1,7 +1,15 @@
...
-    "ClientSecret": "Fill in here or preferrably in ASP.NET Secret Manager"
+    "ClientSecret": "Fill in here or preferably in ASP.NET Secret Manager"
+  },
+  "KeyVault": {
+    "VaultUri": "https://eqrobotkeyvault.vault.azure.net/"
+  },
+  "Echo": {
+    "EchoClientId": "bf0b2569-e09c-42f0-8095-5a52a873eb7b",
+    "EchoAppScope": ".default",
+    "EchoApiUrl": "https://echohubapi.equinor.com/api"
```
Christdej, line 12: Dont think prefixing with Echo is necessary, but I have no strong opinion about this, agreed it may increase readability in some cases

backend/api/Services/KeyVaultService.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        try
+        {
+            KeyVaultSecret keyVaultSecret = await _secretClient.GetSecretAsync(secretName);
+            return keyVaultSecret.Value;
+        }
+        catch
+        {
+            throw;
+        }
```
Christdej, line 22: No need to catch if we just rethrow, `try/catch` can probably be removed

Christdej (changes requested):

backend/api/Services/EchoService.cs
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        public async Task<IList<Mission>> GetMissions()
+        {
+            var accessToken = await AcquireAccessToken();
+            ConfigureRequest(accessToken);
+            var uri = ConstructGetMissionsRequestUri();
+
+            var response = await client.GetAsync(uri);
+            if (!response.IsSuccessStatusCode) throw new MissionNotFoundException($"Failed to retrieve missions from Echo: {response.ReasonPhrase}");
```
anetteu, line 47: This is a very long line, is it better to split it? I like short if-oneliners, but im not sure if i like the long ones.

backend/api/Services/EchoService.cs
```diff
@@ -0,0 +1,182 @@
...
+
+            var echoMissions = await response.Content.ReadFromJsonAsync<List<EchoMissionResponse>>();
+
+            if (echoMissions is null) throw new JsonException("Failed to deserialize missions from Echo");
```
anetteu, line 51: Is there a good reason for using `is` instead of `==` here? I cannot see that it makes a big difference for `null`, but we have used `==` previously and i prefer consistency.

backend/api/Services/EchoService.cs
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        public async Task<Mission> GetMission(int missionId)
+        {
+            var accessToken = await AcquireAccessToken();
+            ConfigureRequest(accessToken);
+            var uri = ConstructGetMissionRequestUri(missionId);
+
+            var response = await client.GetAsync(uri);
+            if (!response.IsSuccessStatusCode) throw new MissionNotFoundException($"Failed to retrieve mission with ID: {missionId} from Echo: {response.ReasonPhrase}");
```
anetteu, line 67: A very long line. See previous similar comment.

backend/api/Services/EchoService.cs
```diff
@@ -0,0 +1,182 @@
...
+
+            var echoMission = await response.Content.ReadFromJsonAsync<EchoMissionResponse>();
+
+            if (echoMission is null) throw new JsonException("Failed to deserialize mission from Echo");
```
anetteu, line 71: See previous similar comment about `==` vs `is`.

backend/api/Controllers/MissionController.cs
```diff
@@ -0,0 +1,84 @@
...
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    public async Task<ActionResult<IList<Mission>>> GetMissions()
+    {
+        try
+        {
+            var missions = await _echoService.GetMissions();
+
```
anetteu, line 41:
```suggestion
```

Just because I think c# produces enough lines and I think it looks just as clean without a whiteline here. Open for discussion.

backend/api/Controllers/MissionController.cs
```diff
@@ -0,0 +1,84 @@
...
+            return Ok(missions);
+        }
+        catch (Exception e) when (e is MissionNotFoundException || e is JsonException)
+        {
+            _logger.LogError(e, "Missions not found");
+
```
anetteu, line 47:
```suggestion
```

Just because I think c# produces enough lines and I think it looks just as clean without a whiteline here. Open for discussion.

backend/api/appsettings.Development.json
aestene: Fair, since it is within the `Echo` section.

backend/api/Controllers/MissionController.cs
```diff
@@ -0,0 +1,84 @@
...
+            return NotFound();
```
anetteu, line 48: Do we want to add a comment here? Something like:

```
NotFound("$Missions not found: {error}", e)
```

Is it necessary with both logging and returning a not found? I think we should do this similarly for all controllers to keep a readable log.

backend/api/Controllers/MissionController.cs
```diff
@@ -0,0 +1,84 @@
...
+            var mission = await _echoService.GetMission(missionId);
+
+            return Ok(mission);
+        }
+        catch (Exception e) when (e is MissionNotFoundException || e is JsonException)
+        {
+            _logger.LogError(e, "Mission not found");
+
+            return NotFound();
+        }
```
anetteu, line 82: See previous comments about logging and whiteline.

backend/api/Controllers/MissionController.cs
```diff
@@ -0,0 +1,84 @@
...
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    public async Task<ActionResult<Mission>> GetMission([FromRoute] int missionId)
+    {
+        try
+        {
+            var mission = await _echoService.GetMission(missionId);
+
```
anetteu, line 74:
```suggestion
```

Just because I think c# produces enough lines and I think it looks just as clean without a whiteline here. Open for discussion.

backend/api/Services/KeyVaultService.cs
aestene: I've added a logging statement to show which secret failed to be retrieved. Will give us a good warning if something isn't working.

backend/api/Services/EchoService.cs
anetteu: Agree

backend/api/Models/Enums/InspectionType.cs
aestene: Fixed

backend/api/Services/EchoService.cs
aestene: Personally I much prefer `is` for checking for `null` values. There is no difference in functionality but I think it reads better.

backend/api/Controllers/MissionController.cs
aestene: What is added to the `NotFound` statement is not logged to the application log, it's sent to the user. So the log statement is what we wish to log in our application while the `NotFound` statement is what we wish to tell the user. 

Do we want to provide the full stack trace to the user? I would think it's more useful to log the stack trace in the application and provide a not found to the user.

aestene: As for `NotFound()` vs ``NotFound("Missions not found") I think they give about the same level of information if you're not adding the error message. I would rather have an empty `NotFound()` maybe than have to maintain specific strings throughout the application for different types of `NotFound()`?

GodVenn: I agree with your points @aestene . I would just point out that if its not clear "what" was NotFound, then it might be worth adding it as a parameter, for example when we call robots/{id}/missions/{id}, We should clarify whether it was the robot or the mission that was not found

anetteu: Sounds good, @aestene and @GodVenn. I think we should make sure that the logging is similar across the controllers, but that can be handled in #199.

backend/api/Services/EchoService.cs
GodVenn: The "is" operator is preferred for checking for null in C# because it guarantees that no user-overloaded method of "==" or "!=" is being used.

backend/api/Database/Models/Mission.cs
```diff
@@ -0,0 +1,24 @@
...
+        [MaxLength(128)]
+        [Required]
+        public string Name { get; set; }
+
+        [Url]
+        [Required]
+        public Uri URL { get; set; }
+
+        [Required]
+        public IList<Tag> Tags { get; set; }
```
Christdej, line 22: missing `virtual` keyword

backend/api/Database/Models/Tag.cs
```diff
@@ -0,0 +1,26 @@
...
+        [MaxLength(128)]
+        [Required]
+        public string TagId { get; set; }
+
+        [Url]
+        [Required]
+        public Uri URL { get; set; }
+
+        [Required]
+        public IList<InspectionType> InspectionTypes { get; set; }
```
Christdej, line 23: missing `virtual` keyword

backend/api/Models/EchoMissionResponse.cs
```diff
@@ -0,0 +1,71 @@
...
+using System.Text.Json.Serialization;
+
+namespace Api.Models
+{
+    public class EchoMissionResponse
+    {
+        [JsonPropertyName("robotPlanId")]
+        public int Id { get; set; }
+
+        [JsonPropertyName("installationCode")]
```
anetteu, line 11: Is it clear what an installationcode is?

Christdej (changes requested):

backend/api/Models/EchoMissionResponse.cs
aestene: In terms of an EchoResponse maybe?

Alternatively we could call it something like `FacibilityShortName`

Christdej (approved):
LGTM

PR #203, opened 2022-05-12
Run csharpier to format all files

aestene:


Christdej (changes requested):
Should we add this to a pipeline? If not every PR is going to have a lot of changes in files not related to the PR.

aestene:
> Should we add this to a pipeline? If not every PR is going to have a lot of changes in files not related to the PR.

Yes, that's the plan but I wanted to see if people wanted to use this first.

GodVenn:
Does this make use of the .editorconfig file? Or does it have its own rules?

aestene:
> Does this make use of the .editorconfig file? Or does it have its own rules?

I think it has its own set of "standard rules". We're following the default now.

GodVenn:
I like the formatting it suggested, although we can also get this by using dotnet format, changing the .editorconfig file.
Do you know if the formatting is based on current C# best practices?
And does it also detect naming conventions, use of 'var' and other more code-related issues?

aestene:
> I like the formatting it suggested, although we can also get this by using dotnet format, changing the .editorconfig file. Do you know if the formatting is based on current C# best practices? And does it also detect naming conventions, use of 'var' and other more code-related issues?

Haven't found any specifics on those things but I like the philosophy they follow: https://prettier.io/docs/en/option-philosophy.html

aestene:
image

GodVenn:
Sounds like a good philosophy, I'm up for trying it!Do we have to make some changes to dotnet format for us to still have rules on naming and such or is it good as-is?

aestene:
I ran these commands https://csharpier.com/docs/IntegratingWithLinters
Although I don't know what effect they have. For now it seems to be working together so I suggest just trying it and see if we get issues and then fix them.

I can add a section to the readme for csharpier. I have added it to the pipeline now.

Christdej (approved):
GJ

PR #208, opened 2022-05-13
Add initial mqtt project

### LINKED ISSUE #207: Add MQTT Client to backend
We need an MQTT client to listen for updates from ISAR in Flotilla.

An MQTT client that triggers C# events in the flotilla API, which then has callback functions that act based on the information from the MQTT client.

GodVenn:
Closes #207


backend/api/MQTT/MessageModels/IsarMission.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Text.Json.Serialization;
```
oysand, line 1: Should the IsarMission topics be defined in Isar and not "duplicated" in Flotilla?

GodVenn: Yes, it would be great if there was a way to somehow "import" these models, I made an [issue](#211 ) on this for future exploration.

backend/api/MQTT/MqttService.cs
```diff
@@ -0,0 +1,194 @@
...
+
+            string topic = messageReceivedEvent.ApplicationMessage.Topic;
+            var messageType = MqttTopics.TopicsToMessages.GetItemByTopic(topic);
+            if (messageType is null)
+            {
+                _logger.LogError("No message class defined for topic '{topicName}'", topic);
+                return;
+            }
+
+            switch (messageType)
```
oysand, line 90: Can we raise an event based on the specific topic rather than grouping them all, then sorting it?

GodVenn: I don't think so but i will look into this.

backend/api/MQTT/MqttService.cs
```diff
@@ -0,0 +1,194 @@
...
+                    f =>
+                    {
+                        f.WithTopic(topic);
+                    }
+                )
+                .Build();
+            _mqttClient.SubscribeAsync(mqttSubscribeOptions.TopicFilters);
+        }
+
+        private void OnIsarTopicReceived<T>(string content) where T : MqttMessage
```
oysand, line 157: Could "T" be spelled out for more readability?

GodVenn: T is the standard C# convention for the "Type" in generic methods, I think it's best to follow this

backend/api/MQTT/MqttService.cs
```diff
@@ -0,0 +1,194 @@
...
+        {
+            Instance = this;
+            _logger = logger;
+            var mqttFactory = new MqttFactory();
+            _mqttClient = mqttFactory.CreateManagedMqttClient();
+
+            var mqttConfig = config.GetSection("Mqtt");
+
+            string username = mqttConfig.GetValue<string>("Username");
+            string password = "default";
```
oysand, line 41: Should this be collected from the environment in some manner?

backend/api/Services/ReportService.cs
```diff
@@ -50,5 +66,27 @@ public async Task<IEnumerable<Report>> ReadAll()
                 report => report.Id.Equals(id, StringComparison.Ordinal)
             );
         }
+
+        private void OnMissionUpdate(object? sender, MqttReceivedArgs mqttArgs)
+        {
+            var mission = (IsarMission)mqttArgs.Message;
+            _logger.LogInformation(
+                "{time} - Mission {id} updated",
```
oysand, line 74: Is there some parallell to fstring that can be used so that we can write {mission.Timestamp} instad of {time}

backend/api/MQTT/MqttService.cs
```diff
@@ -0,0 +1,194 @@
...
+            RegisterCallbacks();
+
+            string[] topics = mqttConfig.GetSection("Topics").Get<string[]>();
+            foreach (string topic in topics)
+            {
+                SubscribeToTopic(topic);
+            }
+        }
+
+        protected override async Task ExecuteAsync(CancellationToken stoppingToken)
```
oysand, line 64: What is the purpose of this? Does it need to be a background service?

backend/api/MQTT/MqttService.cs
GodVenn: Definetily! Waiting for the keyvault

oysand (changes requested):
See comments

backend/api/Services/ReportService.cs
GodVenn: Yes, but the logger recommends this format, not sure why.
`$"{mission.Timestamp - Mission etc"` would be the equivalent

backend/api/MQTT/MqttService.cs
GodVenn: The purpose of what? Not sure what you are referring to

backend/api/Services/ReportService.cs
oysand: Then we can probably assume that the logger recommends it for a reason and follow that.

backend/api/MQTT/MqttService.cs
GodVenn: No support for this in the MQTT library unfortunately

backend/api/MQTT/MqttService.cs
```diff
@@ -0,0 +1,184 @@
...
+
+            var topics = mqttConfig.GetSection("Topics").Get<List<string>>();
+            SubscribeToTopics(topics);
+        }
+
+        protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+        {
+            _logger.LogInformation("MQTT client STARTED");
+            await _mqttClient.StartAsync(_options);
+            while (!stoppingToken.IsCancellationRequested) { }
```
oysand, line 65: Maybe await a task in this rather than a while

oysand (approved):

backend/api/MQTT/MqttService.cs
GodVenn: Found an elegant way to do this [here](https://medium.com/@cilliemalan/how-to-await-a-cancellation-token-in-c-cbfc88f28fa2) and implemented it!

oysand (approved):
LGTM

oysand (changes requested):
There is a bug that the report service that subscribe on the events do not exist when the mqtt messages arrives. This needs to be adressed.

aestene:
The last commit message should be in imperative so change it to "Change log level from info to debug"

backend/api/MQTT/MqttService.cs
```diff
@@ -0,0 +1,186 @@
...
+using Api.MQTT.MessageModels;
+using Api.Utilities;
+using MQTTnet;
+using MQTTnet.Client.Connecting;
+using MQTTnet.Client.Disconnecting;
+using MQTTnet.Client.Options;
+using MQTTnet.Client.Receiving;
+using MQTTnet.Extensions.ManagedClient;
+
+namespace Api.MQTT
```
aestene, line 12: I believe abbreviations above three letters should have lower case e.g. `Mqtt`. This has been used in the class name, should it also be used for the folder and namespace names?

GodVenn: Agreed

backend/api/MQTT/MqttService.cs
```diff
@@ -0,0 +1,186 @@
...
+        {
+            Instance = this;
+            _logger = logger;
+            var mqttFactory = new MqttFactory();
+            _mqttClient = mqttFactory.CreateManagedMqttClient();
+
+            var mqttConfig = config.GetSection("Mqtt");
+
+            string username = mqttConfig.GetValue<string>("Username");
+            string password = config.GetValue<string>("mqtt-broker-password");
```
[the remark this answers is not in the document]
GodVenn: The keyvault is defined in the config (appsettings.Development.json), currently set to flotilla keyvault.
1. Yes, good point
2. Yes, I can change it so that the app will only use keyvault if ASP.NET environment is set to production. We can always manually set this locally if we want to test the keyvault during development.

GodVenn:
> The last commit message should be in imperative so change it to "Change log level from info to debug"

It is already in imperative?

backend/api/MQTT/MqttService.cs
```diff
@@ -0,0 +1,186 @@
...
+
+            switch (messageType)
+            {
+                case Type type when type == typeof(IsarMission):
+                    OnIsarTopicReceived<IsarMission>(content);
+                    break;
+                case Type type when type == typeof(IsarTask):
+                    OnIsarTopicReceived<IsarTask>(content);
+                    break;
+                case Type type when type == typeof(IsarStep):
```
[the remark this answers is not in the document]
GodVenn: Beautiful!

backend/api/MQTT/MqttService.cs
```diff
@@ -0,0 +1,186 @@
...
+                        messageType.Name
+                    );
+                    break;
+            }
+        }
+
+        private void OnConnected(MqttClientConnectedEventArgs obj)
+        {
+            _logger.LogInformation("Successfully connected.");
+        }
```
aestene, line 112: I would suggest including some more information in this log message just because it can be useful. For example adding which host and port you have connected to.

aestene: I assume that may be found in the `MqttClientConnectedEventArgs`

GodVenn: Agreed

backend/api/MQTT/MqttService.cs
```diff
@@ -0,0 +1,186 @@
...
+
+        private void OnConnectingFailed(ManagedProcessFailedEventArgs obj)
+        {
+            _logger.LogWarning(
+                "Couldn't connect to MQTT broker. Exception: {message}\n"
+                    + "      Retrying in {time}s",
+                obj.Exception.Message,
+                _reconnectDelay.Seconds
+            );
+        }
```
aestene, line 122:
```suggestion
        private void OnConnectionFailed(ManagedProcessFailedEventArgs obj)
        {
            _logger.LogWarning(
                "Couldn't connect to MQTT broker. Exception: {message}\n"
                    + "      Retrying in {time}s",
                obj.Exception.Message,
                _reconnectDelay.Seconds
            );
        }
```
Small wording improvement. As far as I can tell this is not an override and I think the standard term is `Connection` rather than `Connecting`

backend/api/MQTT/MqttService.cs
```diff
@@ -0,0 +1,186 @@
...
+
+        private void OnDisconnected(MqttClientDisconnectedEventArgs obj)
+        {
+            _logger.LogInformation("Successfully disconnected.");
+        }
```
aestene, line 127: Same as the above, add some of the context that is probably available in the parameter.

backend/api/MQTT/MqttService.cs
aestene: I see below that `Connecting` is indeed used by the library. I will retract my statement then.

backend/api/MQTT/MqttService.cs
```diff
@@ -0,0 +1,186 @@
...
+
+            var type = typeof(T);
+            try
+            {
+                var raiseEvent = type switch
+                {
+                    _ when type == typeof(IsarMission) => MqttIsarMissionReceived,
+                    _ when type == typeof(IsarTask) => MqttIsarTaskReceived,
+                    _ when type == typeof(IsarStep) => MqttIsarStepReceived,
+                    _
```
aestene, line 169: Should be possible to use the same style for a switch on type as mentioned in the previous comment?

GodVenn: It's not possible in either, because we are not checking the **type** of the variable, we are checking the **value** of the variable, which is a type.

backend/api/MQTT/MqttService.cs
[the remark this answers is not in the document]
GodVenn: Adressed issues, please see updated code

aestene:
> > The last commit message should be in imperative so change it to "Change log level from info to debug"
> 
> It is already in imperative?

Should not be prefixed by "MQTT - "

backend/api/appsettings.Development.json
```diff
@@ -19,5 +19,13 @@
...
-  "AllowedOrigins": "http://localhost:3001"
+  "AllowedOrigins": "http://localhost:3001",
+  "Mqtt": {
+    "Host": "localhost",
+    "Port": 1883,
+    "Username": "mosquitto",
+    "Topics": [
+      "#"
+    ]
+  }
```
aestene, line 30: Have you seen these errors? I don't think subscribing to a wildcard is a very good idea. 
image

Should be mentioned that in my case this is a freshly started broker which seems to have some memory. Maybe there are some builtin topics too?

backend/api/appsettings.Development.json
```diff
@@ -19,5 +19,13 @@
...
   "AllowedHosts": "*",
-  "AllowedOrigins": "http://localhost:3001"
+  "AllowedOrigins": "http://localhost:3001",
+  "Mqtt": {
+    "Host": "localhost",
+    "Port": 1883,
+    "Username": "mosquitto",
+    "Topics": [
+      "#"
+    ]
```
aestene, line 29:
```suggestion
    "Topics": [
      "isar/+/mission",
      "isar/+/task",
      "isar/+/step"
    ]

```
I would suggest this instead.

aestene:
Fixed issue in ISAR that caused the `MissionId` to be `null` when deserializing. https://github.com/equinor/isar/pull/271

backend/api/Program.cs
```diff
@@ -45,10 +47,15 @@
...
-);
+// Adds keyvault as config option if environment is set to production
+// This can be changed locally in the "Properties/launchSettings.json" file
+if (builder.Environment.IsProduction())
+{
+    builder.Configuration.AddAzureKeyVault(
+        new Uri(builder.Configuration.GetSection("KeyVault")["VaultUri"]),
+        new DefaultAzureCredential()
+    );
+}
```
aestene, line 58: I'm not certain this is the preferred solution. Mainly because we're now using a different specific system for secrets (secret manager) locally when it doesn't really solve our problem which is to run the application without secrets in an open source fashion. It might be this is not really that easy with this application as it was with ISAR. 

I suggest we go back to having the keyvault as default also in development. Another error that occurs with this change is that there are no keyvaults if you use the suggestes launch.json change to production as there is no appsettings for production yet.

aestene: Easier to use keyvault also for local development then and only require the user to manage one secret.

backend/api/appsettings.Development.json
aestene: Fixed by using more specific topics.

aestene (approved):
LGTM

aestene:
I have resolved some of the later issues @oysand. Take a look and see what you think now, thought we should merge this rather than wait for next week. 

I don't know what the bug you mentioned before was but I assume @GodVenn resolved it with his earlier commit?

PR #210, opened 2022-05-18
Add keyvault as config default

GodVenn:
This enables us to use the keyvault through the config like we would read any other config variable and simplifies the project structure.


oysand (changes requested):
Looks good, only need a default dev key vault set up. Could this be referred to in the readme?

oysand (approved):
LGTM

PR #222, opened 2022-05-24
Fix no id required to create

### LINKED ISSUE #198: Should not be needed to provide ID when creating event / report / robot
**Describe the bug**
Ids are database generated and should not be necessary to provide during api post calls

**To Reproduce**
Go to swagger and try to create event / report / robot without Id and watch the error

**Expected behavior**
Should be able to create event / report / robot without providing Id

aestene:
Closes #198


anetteu (approved):
LGTM

PR #224, opened 2022-05-26
Update report on events

### LINKED ISSUE #144: Add topic listener process
A topic listener process (insert better name here) should be added in order to listen to communication from ISAR / the robot over MQTT or through API calls if necessary to update the current status of executing missions, ISAR state and robot availability. 

This process should listen to relevant topics and update objects (reports and robots) in the database when new information is received.

Although there is no required solution it is most likely another thread being spun up at startup in `main`. Which listens to predefined topics. Keep in mind that events on topics will lead to this thread modifying the database so thread safe thinking should be important.

Not affected.

### LINKED ISSUE #212: Add database models for Task and Step to reflect ISAR/Echo
ISAR uses Tasks and Steps to represent the structure of the mission, similar to the structure from Echo. This structure should be used in Flotilla as well, were each Report contains a list of Tasks and each Task contain several steps. The database models and API models should be updated to support this structure. Relevant data (time, status, files etc) should be attached to the Task and Step.

Database will better represent the structure in connected services.

No change

aestene:
Closes #212
Closes #144

### DIFF
--- a/backend/api/Database/Models/IsarStep.cs
+++ b/backend/api/Database/Models/IsarStep.cs
@@ -1,4 +1,4 @@
-using System.ComponentModel.DataAnnotations;
+﻿using System.ComponentModel.DataAnnotations;
 using System.ComponentModel.DataAnnotations.Schema;
 
 # nullable disable
@@ -43,22 +43,39 @@ namespace Api.Models
         Cancelled
     }
 
+    public static class IsarStepStatusMethods
+    {
+        public static IsarStepStatus FromString(string status) =>
+            status switch
+            {
+                "successful" => IsarStepStatus.Successful,
+                "not_started" => IsarStepStatus.NotStarted,
+                "in_progress" => IsarStepStatus.InProgress,
+                "failed" => IsarStepStatus.Failed,
+                "cancelled" => IsarStepStatus.Cancelled,
+                _
+                  => throw new ArgumentException(
+                      $"Failed to parse report status {status} as it's not supported"
+                  )
+            };
+    }
+
     public enum StepType
     {
         DriveToPose,
         TakeImage,
         TakeThermalImage,
-        Audio
+        RecordAudio
     }
 
     public class SelectStepType
     {
-        public static StepType From(string sensorType)
+        public static StepType FromSensorTypeAsString(string sensorType)
         {
             return sensorType switch
             {
                 "DriveToPose" => StepType.DriveToPose,
-                "Audio" => StepType.Audio,
+                "RecordAudio" => StepType.RecordAudio,
                 "TakeImage" => StepType.TakeImage,
                 "TakeThermalImage" => StepType.TakeThermalImage,
                 _ => StepType.TakeImage,
@@ -75,7 +92,7 @@ namespace Api.Models
 
     public class SelectInspectionType
     {
-        public static InspectionType From(string sensorType)
+        public static InspectionType FromSensorTypeAsString(string sensorType)
         {
             return sensorType switch
             {
--- a/backend/api/Database/Models/IsarStep.cs
+++ b/backend/api/Database/Models/IsarStep.cs
@@ -4,7 +4,7 @@ using System.ComponentModel.DataAnnotations.Schema;
 # nullable disable
 namespace Api.Models
 {
-    public class Step
+    public class IsarStep
     {
         [Key]
         [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
@@ -15,12 +15,12 @@ namespace Api.Models
         public string IsarStepId { get; set; }
 
         [Required]
-        public virtual Task task { get; set; }
+        public virtual IsarTask Task { get; set; }
 
         public string TagId { get; set; }
 
         [Required]
-        public StepStatus StepStatus { get; set; }
+        public IsarStepStatus StepStatus { get; set; }
 
         [Required]
         public StepType StepType { get; set; }
@@ -34,7 +34,7 @@ namespace Api.Models
         public string FileLocation { get; set; }
     }
 
-    public enum StepStatus
+    public enum IsarStepStatus
     {
         Successful,
         InProgress,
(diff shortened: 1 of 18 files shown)


backend/api/Database/Models/IsarStep.cs
```diff
@@ -1,39 +1,69 @@
...
+    public enum IsarStepStatus
+    {
+        Successful,
+        InProgress,
+        NotStarted,
+        Failed,
+        Cancelled
+    }
+
+    public enum StepType
```
aestene, line 46: Not very happy about the "duplication" of `StepType` and `InspectionType` here but not certain what else to do.

aestene:
There is one issue which I'm uncertain how should be addressed.

When starting a mission in ISAR from Flotilla the response from ISAR contains the mission details which Flotilla then writes to the database as a `Report` object which contains `IsarTask` and `IsarStep` as lists. However, the state machine in ISAR has started executing the current mission before Flotilla has written to the database. Thus ISAR is sending status updates for mission, task and step which Flotilla will not recognize because it can't find those IDs in the database yet as the write operation is not complete. 

Not entirely certain how this should be resolved.

aestene:
There is one issue which I'm uncertain how should be addressed.

When starting a mission in ISAR from Flotilla the response from ISAR contains the mission details which Flotilla then writes to the database as a `Report` object which contains `IsarTask` and `IsarStep` as lists. However, the state machine in ISAR has started executing the current mission before Flotilla has written to the database. Thus ISAR is sending status updates for mission, task and step which Flotilla will not recognize because it can't find those IDs in the database yet as the write operation is not complete. 

Not entirely certain how this should be resolved.

oysand:
I think a better way to solve this issue is to move the misssion planner from ISAR to Flotilla so that Flotilla provides the mission to ISAR. That way we can toggle if ISAR should use a local mission planner or be used in combination with Flotilla. We can then write the mission to the database before sending it to ISAR

oysand:
> There is one issue which I'm uncertain how should be addressed.
> 
> When starting a mission in ISAR from Flotilla the response from ISAR contains the mission details which Flotilla then writes to the database as a `Report` object which contains `IsarTask` and `IsarStep` as lists. However, the state machine in ISAR has started executing the current mission before Flotilla has written to the database. Thus ISAR is sending status updates for mission, task and step which Flotilla will not recognize because it can't find those IDs in the database yet as the write operation is not complete.
> 
> Not entirely certain how this should be resolved.

I think a better way to solve this issue is to move the misssion planner from ISAR to Flotilla so that Flotilla provides the mission to ISAR. That way we can toggle if ISAR should use a local mission planner or be used in combination with Flotilla. We can then write the mission to the database before sending it to ISAR

aestene:
> > There is one issue which I'm uncertain how should be addressed.
> > When starting a mission in ISAR from Flotilla the response from ISAR contains the mission details which Flotilla then writes to the database as a `Report` object which contains `IsarTask` and `IsarStep` as lists. However, the state machine in ISAR has started executing the current mission before Flotilla has written to the database. Thus ISAR is sending status updates for mission, task and step which Flotilla will not recognize because it can't find those IDs in the database yet as the write operation is not complete.
> > Not entirely certain how this should be resolved.
> 
> I think a better way to solve this issue is to move the misssion planner from ISAR to Flotilla so that Flotilla provides the mission to ISAR. That way we can toggle if ISAR should use a local mission planner or be used in combination with Flotilla. We can then write the mission to the database before sending it to ISAR

I have mentioned this in a new issue #226. It's easier to solve in a different PR and to just include this functionality as it is now. It works somewhat for turtlebot, but there are some unfortunate sideffects.

GodVenn (commented):
LGTM, just one renaming suggestion

backend/api/Database/Models/IsarStep.cs
```diff
@@ -0,0 +1,115 @@
...
+        Successful,
+        InProgress,
+        NotStarted,
+        Failed,
+        Cancelled
+    }
+
+    public static class IsarStepStatusMethods
+    {
+        public static IsarStepStatus From(string status)
```
GodVenn, line 48:
```suggestion
        public static IsarStepStatus FromString(string status)
```

backend/api/Database/Models/IsarStep.cs
```diff
@@ -0,0 +1,115 @@
...
+    {
+        DriveToPose,
+        TakeImage,
+        TakeThermalImage,
+        Audio
+    }
+
+    public class SelectStepType
+    {
+        public static StepType From(string sensorType)
```
GodVenn, line 80:
```suggestion
        public static StepType FromSensorTypeAsString(string sensorType)
```

backend/api/Database/Models/IsarStep.cs
```diff
@@ -0,0 +1,115 @@
...
+    public enum InspectionType
+    {
+        Image,
+        ThermalImage,
+        Audio
+    }
+
+    public class SelectInspectionType
+    {
+        public static InspectionType From(string sensorType)
```
GodVenn, line 102:
```suggestion
        public static InspectionType FromSensorTypeAsString(string sensorType)
```

GodVenn (commented):
Some naming suggestions for methods

oysand (approved):
LGTM

backend/api/Database/Models/IsarStep.cs
aestene: Done

backend/api/Database/Models/IsarStep.cs
aestene: Done

backend/api/Database/Models/IsarStep.cs
aestene: Done

backend/api/Database/Models/IsarStep.cs
```diff
@@ -4,36 +4,66 @@
...
+        Failed,
+        Cancelled
+    }
+
+    public enum StepType
+    {
+        DriveToPose,
+        TakeImage,
+        TakeThermalImage,
+        Audio
```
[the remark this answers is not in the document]
aestene: Done

GodVenn (approved):
LGTM

PR #228, opened 2022-06-08
Authentication for mqtt

### LINKED ISSUE #93: Setup Access Control List for the MQTT-broker
The security of the broker can be improved by using authorization on the different topics. The mosquitto broker offers the opportunity to use an Access Control List (ACL) to perform authorization. In ACL you can specify which topics a certain user can publish/subscribe to. A brief introduction can be found [here](https://www.chirpstack.io/project/guides/mqtt-authentication/). Different users (with passwords) should be generated for the different roles (streaming of telemetry, status from isar, flotilla backend, flotilla frontend).

This will improve the security of the broker and will have potential effects to all clients of the broker, both publishers and subscribers.

This should reduce the risk of unauthorized publishing to the broker and leak of information (listening to certain topics).

### LINKED ISSUE #127: Add improved authentication to Isar-Flotilla MQTT
Extend authentication method with role-based access.

User (Isar) will have to have an app registration which is added to an access group for this communication link.

Mitigate spoof data being fed to Flotilla from malicious actor.

GodVenn:
Closes #93 
closes #127

### DIFF
--- a/backend/api/MQTT/MqttService.cs
+++ b/backend/api/MQTT/MqttService.cs
@@ -1,4 +1,5 @@
-﻿using System.Text.Json;
+﻿using System.Text;
+using System.Text.Json;
 using Api.Mqtt.Events;
 using Api.Mqtt.MessageModels;
 using Api.Utilities;
@@ -81,7 +82,11 @@ namespace Api.Mqtt
                 return;
             }
 
-            _logger.LogDebug("Topic: {topic} - Message recieved: \n{payload}", topic, content);
+            _logger.LogInformation(
+                "Topic: {topic} - Message recieved: \n{payload}",
+                topic,
+                content
+            );
 
             switch (messageType)
             {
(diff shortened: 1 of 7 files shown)


backend/api/MQTT/MqttService.cs
```diff
@@ -1,4 +1,5 @@
-﻿using System.Text.Json;
+﻿using System.Text;
+using System.Text.Json;
```
oysand, line 2: Do we need to say that we use .Json as well when we say using System.Text?

GodVenn: Yes, we do

backend/api/MQTT/MqttService.cs
```diff
@@ -147,18 +152,34 @@ private void RegisterCallbacks()
...
+            sb.AppendLine("Mqtt service subscribing to the following topics:");
+            topics.ForEach(
+                topic =>
+                {
+                    topicFilters.Add(new MqttTopicFilter() { Topic = topic });
+                    sb.AppendLine(topic);
+                }
+            );
+            _logger.LogInformation("{topicContent}", sb.ToString());
+            _mqttClient.SubscribeAsync(topicFilters).Wait();
```
oysand, line 165: Does this do the same as await function()? Why not await?

broker/mosquitto/config/access_control
```diff
@@ -0,0 +1,8 @@
+user admin
+topic readwrite #
+
+user isar
+topic readwrite isar/#
+
+user flotilla
+topic read isar/#
```
oysand, line 8: Add newline

oysand (approved):
LGTM

PR #233, opened 2022-06-10
Bug mqtt infinite retries

### LINKED ISSUE #229: MQTT service retries infinitely
**Describe the bug**
The MQTT service in the backend will attempt to reconnect an infinite amount of times.
After a certain number of tries, it should give up and fail, or we might never know that something is wrong unless watching the logs.

GodVenn:
Closes #229 
The current implementation causes flotilla backend to simply stop if the broker cannot be reached within the attempt limit.
This can be disabled by changing the config variable `ShouldFailOnMaxRetries`.
This was done because the backend is reliant on being connected to the broker for it to function as intended, and if it can't reach the broker, which is on the same docker compose network, then this implies that something is very wrong.
Not sure about this, though, and I'm very open for discussion.

### DIFF
--- a/backend/api/MQTT/MqttService.cs
+++ b/backend/api/MQTT/MqttService.cs
@@ -1,4 +1,6 @@
-﻿using Api.Mqtt.Events;
+﻿using System.Text;
+using System.Text.Json;
+using Api.Mqtt.Events;
 using Api.Mqtt.MessageModels;
 using Api.Utilities;
 using MQTTnet;
@@ -7,8 +9,6 @@ using MQTTnet.Client.Disconnecting;
 using MQTTnet.Client.Options;
 using MQTTnet.Client.Receiving;
 using MQTTnet.Extensions.ManagedClient;
-using System.Text;
-using System.Text.Json;
 
 namespace Api.Mqtt
 {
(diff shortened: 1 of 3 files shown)


backend/api/MQTT/MqttService.cs
```diff
@@ -115,25 +122,54 @@ private void OnConnected(MqttClientConnectedEventArgs obj)
...
+            string errorMsg =
+                "Failed to connect to MQTT broker. Exception: " + obj.Exception.Message;
+
+            if (_reconnectAttempts >= _maxRetryAttempts)
+            {
+                _logger.LogError("{errorMsg}\n      Exceeded max reconnect attempts.", errorMsg);
+
+                if (_shouldFailOnMaxRetries)
+                {
+                    _logger.LogError("Stopping MQTT client due to critical failure");
```
oysand, line 139: Should we crash the entire backend or just the mqtt service if this is reached?

GodVenn: If we crash the mqtt service, the backend also crashes.
We could configure ASP.NET to not crash on the crash of a Hosted Service, but the hosted service will never be restarted so that doesn't seem like a good idea.

oysand (changes requested):
When running the backend in docker, the retries are not initiated and it does the following over and over: 
image

The retry fix should work in docker as well

backend/api/MQTT/MqttService.cs
oysand: The backend still do not stop when the retries are used up.

GodVenn:
> When running the backend in docker, the retries are not initiated and it does the following over and over: image
> 
> The retry fix should work in docker as well

This looks like you are running an earlier version of the build in docker, have you run the docker compose with the --build flag?

PR #239, opened 2022-06-16
Implement custom robot cards using styled components

MortFred:


aestene (approved):
LGTM

frontend/src/components/MissionOverview/MissionCard.tsx
```diff
@@ -0,0 +1,5 @@
+import { Card, Typography } from '@equinor/eds-core-react'
+
+export function MissionCard() {
+    return <Card>hello</Card>
```
oysand, line 4: Remove hello as this makes it so it says three hello three times below all three of the cards.

MortFred: This is expected behaviour

PR #260, opened 2022-06-28
Feature tls for mqtt communication

### LINKED ISSUE #232: Configure TLS connection for MQTT clients/broker
The current connection between the MQTT broker and it's clients are not encrypted.
This means the password and usernames used for connecting are sent in clear text, and is vulnerable to interception.
To mitigate this, we should use TLS to encrypt the communication.

This will not change functionality, but will increase security.

Reduce the risk of leaking information about the statuses of robots to unauthorized 3rd parties.

GodVenn:
Closes #232 

Also added an `appsettings.Production` file so that we can set the app to production and it won't crash due to lacking variables.


Christdej (approved):
LGTM

PR #266, opened 2022-07-04
Use sqlite inmemory database

Christdej:
Change test / dev database to SQLite. 
Add possibility to connect to remote SQL database.
Update tests and usage of database in tests

### DIFF
--- a/backend/api/Database/Models/Report.cs
+++ b/backend/api/Database/Models/Report.cs
@@ -12,7 +12,7 @@ namespace Api.Database.Models
         public string Id { get; set; }
 
         [Required]
-        public string RobotId { get; set; }
+        public Robot Robot { get; set; }
 
         [MaxLength(128)]
         [Required]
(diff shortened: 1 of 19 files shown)


backend/api/Database/Models/Report.cs
```diff
@@ -12,7 +12,7 @@ public class Report
         public string Id { get; set; }
 
         [Required]
-        public string RobotId { get; set; }
+        public Robot Robot { get; set; }
```
GodVenn, line 15: Why was this changed back to Robot? Are we certain this behaves as expected? 
For example, when GETing a report object through the API, the Robot field will now always be null.

GodVenn (commented):
Looks good, but 4 of the tests fail locally for me, due to the 'concurrency exception'.
I also added a commit to move the database config from the Program.cs file to its own service extension to keep the Program.cs file light

GodVenn:
Not sure if we should merge with the change `RobotId` -> `Robot`, want to know why this was done.
Suggesting holding off merge until @Christdej is back from vacations since this is not urgent.

GodVenn (approved):
Looks great to me

Issue #263, opened 2022-06-30
Flotilla should create robot automatic on connecting to ISAR


Christdej:
**Describe the new feature you would like to see**
When an instance of ISAR is connected to flotilla, a robot should be automatically added to the database fetching relevant information from ISAR

Christdej:
We can also consider storing "ISAR_ROBOT_ID" or something similar that is generated by ISAR, not the database in flotilla. Today this "ID" is the name of the robot, i.e. William, not a good ID

Issue #235, opened 2022-06-13
Update ScheduledMission status upon completed mission


oysand:
**Describe the improvement you would like to see**
When missions are completed the ScheduledMission object status should be updated to Completed/Successfull/Failed.

**How will this change existing functionality?**
The status of ScheduledMissions are currently Pending, then Started. Now it would have a status that indicate if the mission have been finished.

**How will this improvement affect the current Threat Model?**
Should not affect the threat model.

PR #270, opened 2022-07-07
Update status when mission completes

### LINKED ISSUE #234: Update robot status
When missions are started and finished the robot status should change between busy and available.

When the robot is in a mission it should say busy in the database so that the ScheduledMissionEventHandler does not try to schedule the next mission.

This should not effect the threat model.

anetteu:
This PR does a few things:

- Cleans up some bugs and clearfies the use of missionId/EchoMissionId/IsarMissionId
- Allows for reading robots from the database by name. What is the robot.name in flotilla is the robotId in ISAR.

A bit confusing with the robots and missions ids and names? ISAR and Flotilla should exchange and align the IDs to make sure they talk about the same items. Further, the fetching operation of the echo mission should possibly be moved from flotilla to ISAR. See #263  and https://github.com/equinor/isar/issues/293.

Resolves #234 #235 :
- Updates the robots status to Busy when the mission is started, and updates it back to Available when the mission status is reported as completed from ISAR.
- Deletes the scheduledMission object when the mission status id reported as completed from ISAR. In the issue, it is suggested to update the status of the scheduledMission to completed, but I suggest deleting it s the information is kept in the report.


backend/api/Controllers/RobotController.cs
```diff
@@ -193,6 +193,8 @@ public async Task<ActionResult<Robot>> PostRobot([FromBody] Robot robot)
         try
         {
             var report = await _isarService.StartMission(robot, echoMissionId);
+            robot.Status = RobotStatus.Busy;
+            await _robotService.Update(robot);
```
GodVenn, line 197: We might need to verfify from ISAR that the mission was successful in starting before setting robot to busy.

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -61,6 +66,18 @@ private async void OnMissionUpdate(object? sender, MqttReceivedArgs mqttArgs)
                     mission.Status,
                     mission.RobotId
                 );
+            if (mission.Status == "completed")
```
GodVenn, line 68: Use proper string comparison pattern "string.Equals(otherString, StringComparison.OrdinalIgnoreCase)" (This should probably be case insensitive)

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -61,6 +66,18 @@ private async void OnMissionUpdate(object? sender, MqttReceivedArgs mqttArgs)
...
+                _logger.LogInformation("Mission with ISAR mission id {id} is completed by the robot {name}", mission.MissionId, mission.RobotId);
```
GodVenn, line 69: Based on the previous commits, im confused if this is now ISAR mission ID or Echo mission ID ?

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -61,6 +66,18 @@ private async void OnMissionUpdate(object? sender, MqttReceivedArgs mqttArgs)
...
+                var robot = await _robotService.ReadByName(mission.RobotId);
+                if (robot == null)
+                {
+                    _logger.LogWarning("Could not find robot with name {id}. The robot status is not updated.", mission.RobotId);
```
GodVenn, line 73: I think maybe this should be an error

backend/api/EventHandlers/ScheduledMissionEventHandler.cs
```diff
@@ -72,14 +72,17 @@ private async Task<bool> StartScheduledMission(ScheduledMission scheduledMission
             {
                 var report = await _isarService.StartMission(robot: scheduledMission.Robot, echoMissionId: scheduledMission.EchoMissionId);
                 _logger.LogInformation("Started mission {id}", scheduledMission.Id);
+                await _scheduledMissionService.Delete(scheduledMission.Id);
```
GodVenn, line 75: Should the scheduled mission be deleted upon completion instead? In case it fails, it might be nice to be able to reschedule it, using the same instance

backend/api/Controllers/RobotController.cs
anetteu: fixed

MortFred (commented):
LGTM

Issue #247, opened 2022-06-21
Flotilla assumes only one instance of ISAR connected


GodVenn:
**Describe the bug**
flotilla currently operates from the assumption of only one connected ISAR instance, with a single mission running at any given time.
This is not necessarily the case as flotilla will be connected to several robots and thus several instances of ISAR.

**To Reproduce**
Attempt to control more than one robots from flotilla.

**Expected behavior**
The different ISAR instances should be handled based on the robot ID's related to flotilla.

aestene:
The solution could be to collect the ISAR URL and port from the database instead of storing it as a variable `isarUrl` within the `IsarService`

PR #276, opened 2022-07-13
Update robot when ISAR connects

anetteu:
Solves #247 and solves #263

This PR does two things:

1. Makes sure a that there is one robot in the database corresponding to the connected instances of ISAR. Currently, the `name` of the robot, stored as `robot.Name` in flotilla and `robotId` in ISAR is the unique identifier. The `host` and `port` for each robot is fetched from ISAR. Whether assigning `host` and `port` should be the responsibility of ISAR and flotillla is up for discussion. 

1. Allows for several ISAR instances /robots to be connected and monitored to flotilla. By making sure that each robot is assigned to an ISAR instance and that it has the updated ISAR api connection info `robot.Host` and `robot.Port`, the general `_isarUri` config is removed and a url is generated for each robot.

The robot in the database should also be disabled when ISAR disconnects. See #278.

### DIFF
--- a/backend/api/Controllers/RobotController.cs
+++ b/backend/api/Controllers/RobotController.cs
@@ -220,7 +220,13 @@ public class RobotController : ControllerBase
     [ProducesResponseType(StatusCodes.Status500InternalServerError)]
     public async Task<ActionResult<IsarStopMissionResponse>> StopMission([FromRoute] string robotId)
     {
-        var response = await _isarService.StopMission();
+        var robot = await _robotService.ReadById(robotId);
+        if (robot == null)
+        {
+            _logger.LogWarning("Could not find robot with id={id}", robotId);
+            return NotFound();
+        }
+        var response = await _isarService.StopMission(robot);
         if (!response.IsSuccessStatusCode || response.Content is null)
         {
             _logger.LogError("Could not stop mission on robot: {robotId}", robotId);
@@ -332,11 +338,17 @@ public class RobotController : ControllerBase
     [ProducesResponseType(StatusCodes.Status403Forbidden)]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
     [ProducesResponseType(StatusCodes.Status500InternalServerError)]
-    public async Task<ActionResult<string>> PauseMission()
+    public async Task<ActionResult<string>> PauseMission([FromRoute] string robotId)
     {
+        var robot = await _robotService.ReadById(robotId);
+        if (robot == null)
+        {
+            _logger.LogWarning("Could not find robot with id={id}", robotId);
+            return NotFound();
+        }
         try
         {
-            var response = await _isarService.PauseMission();
+            var response = await _isarService.PauseMission(robot);
             string? responseContent = await response.Content.ReadAsStringAsync();
             string? isarResponse = JsonSerializer.Deserialize<string>(responseContent);
             return Ok(isarResponse);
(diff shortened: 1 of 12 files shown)


anetteu:
Need to update tests, but please have a look at the general implementation

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -42,6 +43,28 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
             await stoppingToken;
         }
 
+        private async void OnIsarConnect(object? sender, MqttReceivedArgs mqttArgs)
+        {
+            var isarRobot = (IsarConnectMessage)mqttArgs.Message;
+            var robot = await _robotService.ReadByName(isarRobot.RobotId);
+            _logger.LogError("ISAR CONNECT");
```
GodVenn, line 50: Error?

backend/api/Services/RobotService.cs
```diff
@@ -20,6 +20,24 @@ public async Task<Robot> Create(Robot newRobot)
             return newRobot;
         }
 
+        public async Task<Robot> Create(string name, string host, int port)
```
GodVenn, line 23: I guess this raises a question to this approach - When creating a robot from the ISAR MQTT message, we don't have enough information.
Maybe this points to the possibility that we should not automatically create robots in database, and that this should be done by users through the frontend, as it is a rare and important operation.

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -42,6 +43,28 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
+                await _robotService.Create(isarRobot.RobotId, isarRobot.Host, isarRobot.Port);
+                _logger.LogInformation("Isar instance for robot {name} is connected. Robot is created and added to the database.", isarRobot.RobotId);
+            }
+            else
+            {
+                robot.Host = isarRobot.Host;
+                robot.Port = isarRobot.Port;
+                robot.Enabled = true;
+                await _robotService.Update(robot);
+                _logger.LogInformation("Isar instance for robot {name} is connected. Robot is enabled and host and port is updated.", isarRobot.RobotId);
```
GodVenn, line 62: Maybe also log the Flotilla `robotId` and the updated host and port

backend/api/MQTT/MessageModels/IsarConnect.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Text.Json.Serialization;
+
+namespace Api.Mqtt.MessageModels
+{
+#nullable disable
+    public class IsarConnectMessage : MqttMessage
```
GodVenn, line 6: Why no timestamp here, like in ISAR?

GodVenn (commented):
Pretty clean, but I have some doubts to the automatic creation of robots

backend/api/Services/RobotService.cs
anetteu: You are right. Lets log an error instead of creating a new robot when an isar instance connects and there is no robot with the same name/id in the database

backend/api/Controllers/RobotController.cs
```diff
@@ -332,11 +338,17 @@ public async Task<ActionResult<IList<VideoStream>>> GetVideoStreams([FromRoute]
     [ProducesResponseType(StatusCodes.Status403Forbidden)]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
     [ProducesResponseType(StatusCodes.Status500InternalServerError)]
-    public async Task<ActionResult<string>> PauseMission()
+    public async Task<ActionResult<string>> PauseMission([FromRoute] string robotId)
```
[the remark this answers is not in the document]
GodVenn: Oups, my bad, this is just wrong, I blame the early morning

backend/api/Services/IsarService.cs
```diff
@@ -78,15 +77,19 @@ public async Task<Report> StartMission(Robot robot, string echoMissionId)
             return await _reportService.Create(report);
         }
 
-        public async Task<HttpResponseMessage> StopMission()
+        public async Task<HttpResponseMessage> StopMission(Robot robot)
         {
-            var builder = new UriBuilder($"{_isarUri}/schedule/stop-mission");
+            string host = robot.Host;
+            if (host == "0.0.0.0") host = "localhost";
+            var builder = new UriBuilder($"http://{host}:{robot.Port}/schedule/stop-mission");
```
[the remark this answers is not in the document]
GodVenn: Fixed

GodVenn (approved):
LGTM, Good approach on using the OnConnect for MQTT to handle this!

PR #277, opened 2022-07-13
Feature - Send mission to ISAR as Json

### LINKED ISSUE #271: Fetch missions from echo to send as json to ISAR
ISAR will generalize its missions to receive them in json format rather than receiving ID's on what and where to fetch them (see [#293](https://github.com/equinor/isar/issues/293)).
This means that Flotilla now has to support a local mission planner (reading from file) and fetching missions from Echo.
Both of these missions need to have methods to convert them to the mission format ISAR expects.

An interface `IMissionPlanner` describing the methods expected from a mission service, amongst them should be a conversion to an `IsarMission` class that represents the mission format in ISAR so that we can send them as json to ISAR.
This PR should include at least one implementation of the `IMissionPlanner` interface, `EchoMissionPlanner`.

GodVenn:
Closes #271 
Please review this commit-by-commit by clicking on `Files changed`-tab -> changes from -> `top (first) commit`, and then clicking through them using the arrow navigation.
It will make the pull request much more comprehensable

This Pull request also stabilizes tests by adding interfaces for proper mocking, re-creating database on each test, and properly utilizing the Dispose function on test classes to reset the state before each test.

It also 'fixes' our event handlers, by adding a base class that makes sure events are unsubscribed when the object is disposed. This issue was a memory leak, and was detected during creation of tests.

### DIFF
--- a/backend/api.test/Services/ScheduledMissionService.cs
+++ b/backend/api.test/Services/ScheduledMissionService.cs
@@ -1,25 +1,33 @@
-﻿using System.Linq;
+﻿using System;
+using System.Linq;
 using System.Threading.Tasks;
+using Api.Database.Context;
 using Api.Database.Models;
 using Api.Services;
 using Xunit;
 
-namespace Api.Test
+namespace Api.Test.Services
 {
     [Collection("Database collection")]
-    public class ScheduledMissionServiceTest
+    public class ScheduledMissionServiceTest : IDisposable
     {
-        private readonly DatabaseFixture _fixture;
+        private readonly FlotillaDbContext _context;
 
         public ScheduledMissionServiceTest(DatabaseFixture fixture)
         {
-            _fixture = fixture;
+            _context = fixture.NewContext;
+        }
+
+        public void Dispose()
+        {
+            _context.Dispose();
+            GC.SuppressFinalize(this);
         }
 
         [Fact]
         public async Task ReadAll()
         {
-            var scheduledMissionService = new ScheduledMissionService(_fixture.Context);
+            var scheduledMissionService = new ScheduledMissionService(_context);
             var scheduledMissions = await scheduledMissionService.ReadAll();
 
             Assert.True(scheduledMissions.Any());
@@ -28,7 +36,7 @@ namespace Api.Test
         [Fact]
         public async Task Read()
         {
-            var scheduledMissionService = new ScheduledMissionService(_fixture.Context);
+            var scheduledMissionService = new ScheduledMissionService(_context);
             var scheduledMissions = await scheduledMissionService.ReadAll();
             var firstScheduledMission = scheduledMissions.First();
             var scheduledMissionById = await scheduledMissionService.ReadById(firstScheduledMission.Id);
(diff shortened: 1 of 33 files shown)


backend/api/Controllers/RobotController.cs
```diff
@@ -180,7 +184,7 @@ public async Task<ActionResult<Robot>> PostRobot([FromBody] Robot robot)
     [ProducesResponseType(StatusCodes.Status500InternalServerError)]
     public async Task<ActionResult<Report>> StartMission(
```
MortFred, line 185: Seems strange that we ask for a report here and not a mission object.
Not something to refactor now, but I am commenting so I don't forget it during discussion.

backend/api/appsettings.Development.json
```diff
@@ -14,7 +14,7 @@
     }
   },
   "AllowedHosts": "*",
-  "AllowedOrigins": "http://localhost:3001",
+  "AllowedOrigins": "http://localhost:3000",
```
MortFred, line 17: Why? I think we need 3001 for the frontend to be able to connect to Flotilla backend?

backend/api.test/Services/ScheduledMissionService.cs
```diff
@@ -31,7 +31,7 @@ public async Task Read()
             var scheduledMissionService = new ScheduledMissionService(_fixture.Context);
             var scheduledMissions = await scheduledMissionService.ReadAll();
             var firstScheduledMission = scheduledMissions.First();
-            var scheduledMissionById = await scheduledMissionService.Read(firstScheduledMission.Id);
+            var scheduledMissionById = await scheduledMissionService.ReadById(firstScheduledMission.Id);
```
MortFred, line 42: Read by which Id? IsarMissionId, EchoMissionId, reportId, missionId

backend/api/Services/ReportService.cs
```diff
@@ -4,9 +4,38 @@
...
+
+        public abstract Task<IsarStep?> ReadIsarStepById(string isarStepId);
+
+        public abstract Task<bool> UpdateMissionStatus(string isarMissionId, ReportStatus reportStatus);
+
+        public abstract Task<bool> UpdateTaskStatus(string isarTaskId, IsarTaskStatus taskStatus);
+
+        public abstract Task<bool> UpdateStepStatus(string isarStepId, IsarStep.IsarStepStatus stepStatus);
+    }
+
```
MortFred, line 35: I could use an explanation on this programming pattern.
Are we planning to use several implementations of this interface?

MortFred:
Good work! :)

backend/api/appsettings.Development.json
GodVenn: My bad, i mixed it up with the ISAR port

backend/api/Controllers/RobotController.cs
GodVenn: I tend to agree with this, the report can then be looked up based on the mission ID if needed

backend/api.test/Services/ScheduledMissionService.cs
GodVenn: I think it's clear from the context that this is the ScheduledMissionService and so it will be the ScheduledMissionId, this is following the standard CRUD naming of functions

backend/api/Services/ReportService.cs
GodVenn: Yes, sure!
The use interfaces for services is recommended for scalability (In which case they should be in their own files, of course), they also serve as a good overview/documentation of the intended key features of a service.
In this case I implemented them because they greatly reduce the need for mocking of code in the test project, which is a nice bonus.

Issue #273, opened 2022-07-12
Consider to log less of the regular status updates


anetteu:
**Describe the improvement you would like to see**
The status updates with battery level is now sent every second, and these messages are logged in the flotilla backend. Is it necessary to send this information every second? If so, can the logging be reduced to avoid spamming the backend log. 

**How will this change existing functionality?**
Possibly a bit less frequent update on battery status and less logging. The frequency on status updates can be discussed - pose, mission status, and battery level does not necessarily need the same update frequency. 

**How will this improvement affect the current Threat Model?**
No effect

PR #281, opened 2022-07-14
Remove dev logging

anetteu:
This error logging was just ment for dev as there is a of logging happening and hard to find those lines you are looking for. Reducing the logging of mqtt messages and frequent status updates should be considered to get a more readable log with less spam. See #273


GodVenn (approved):
LGTM

PR #288, opened 2022-07-20
Fix all references to echo mission ID to int

GodVenn:
After a couple of rebases, the change of the echo mission ID was not complete.
This PR fixes the bugs related to this.


MortFred (approved):
LGTM

PR #297, opened 2022-07-25
Add asset select functionality

### LINKED ISSUE #293: Enable selection of assets as an option
The user should be able to choose and asset in the frontend, and only get relevant map- and robot data for that asset.

Save asset as a global variable in the frontend, and send this variable in API calls to backend

By coordinating asset selection to user access scope, we can limit the amount of data a user can see, thus improving safety.

MortFred:
Closes #293


GodVenn (approved):
LGTM

PR #302, opened 2022-07-27
Disable robot if ISAR connection is lost

### LINKED ISSUE #255: Handle HttpRequestException when connecting to ISAR
**Describe the bug**
When making a request to ISAR through IsarService and the ISAR instance is unavailable we get an HttpRequestException which is unhandled.

**To Reproduce**
- Start Flotilla
- Use any endpoint towards ISAR when ISAR is not running

**Expected behavior**
Exception should be handled and reported to the user.

**Screenshots**
image

### LINKED ISSUE #278: Robot should be disabled when ISAR instance disconnects
The robot in the database should be disabled by setting `robot.Enabled = False` when the corresponding ISAR instance is disconnected.

This can be triggered by either flotilla checking with ISAR instances that is connected, or ISAR publishing a message when disconnecting.

No change.

sondreo:
Resolves #278 by disabling a robot in the database whenever an API POST request using `IsarService` throws an `HttpRequestException`, indicating that the corresponding ISAR instance is unavailable. At once, the robot status is also set to `Offline`.

The `HttpRequestException` is caught for all endpoints involving `IsarService` and the error reported, which fixes #255.

### DIFF
--- a/backend/api/Controllers/RobotController.cs
+++ b/backend/api/Controllers/RobotController.cs
@@ -214,34 +214,46 @@ public class RobotController : ControllerBase
             }
 
             _logger.LogError(e, "Error getting mission from Echo");
-            return new StatusCodeResult(StatusCodes.Status502BadGateway);
+            return StatusCode(StatusCodes.Status502BadGateway, $"{e.Message}");
         }
         catch (JsonException e)
         {
-            _logger.LogError(e, "Error deserializing mission from Echo");
-            return new StatusCodeResult(StatusCodes.Status500InternalServerError);
+            string message = "Error deserializing mission from Echo";
+            _logger.LogError(e, "{message}", message);
+            return StatusCode(StatusCodes.Status500InternalServerError, message);
         }
 
+        Report report;
         try
         {
-            var report = await _isarService.StartMission(robot, echoMissionId, new IsarMissionDefinition(echoMission));
-            robot.Status = RobotStatus.Busy;
-            await _robotService.Update(robot);
-            return Ok(report);
+            report = await _isarService.StartMission(robot, echoMissionId, new IsarMissionDefinition(echoMission));
         }
         catch (HttpRequestException e)
         {
-            _logger.LogError(e, "Error connecting to isar while starting mission");
+            string message = "Error connecting to ISAR while starting mission";
+            _logger.LogError(e, "{message}", message);
             robot.Enabled = false;
             robot.Status = RobotStatus.Offline;
             await _robotService.Update(robot);
-            return new StatusCodeResult(StatusCodes.Status502BadGateway);
+
+            return StatusCode(StatusCodes.Status502BadGateway, message);
         }
-        catch (Exception e)
+        catch (MissionException e)
         {
-            _logger.LogError(e, "Error while starting isar mission");
-            throw;
+            _logger.LogError(e, "Error while starting ISAR mission");
+            return StatusCode(StatusCodes.Status502BadGateway, $"{e.Message}");
         }
+        catch (JsonException e)
+        {
+            string message = "Error while processing of the response from ISAR";
+            _logger.LogError(e, "{message}", message);
+            return StatusCode(StatusCodes.Status500InternalServerError, message);
+        }
+
+        robot.Status = RobotStatus.Busy;
+        await _robotService.Update(robot);
+
+        return Ok(report);
     }
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


backend/api/Controllers/RobotController.cs
```diff
@@ -258,23 +266,34 @@ public async Task<ActionResult<IsarStopMissionResponse>> StopMission([FromRoute]
...
+                    return new StatusCodeResult(statusCode);
 
+                return new StatusCodeResult(StatusCodes.Status502BadGateway);
+            }
+
+            string? responseContent = await response.Content.ReadAsStringAsync();
+            var isarResponse = JsonSerializer.Deserialize<IsarStopMissionResponse>(responseContent);
+            return Ok(isarResponse);
+        }
+        catch (HttpRequestException e)
```
GodVenn, line 289: Reduce the try-clause to only contain the call that is being tested

GodVenn (commented):
Good job!
Not really related to this issue, but it seems as if there is a difference between the endpoints in how much handling is left to the IsarService. Ideally this should be standardized between them so that the seperation between controller and service is clear, maybe this could be included in this PR?

backend/api/Controllers/RobotController.cs
sondreo: Done.

backend/api/Services/IsarService.cs
```diff
@@ -56,7 +56,7 @@ public async Task<Report> StartMission(Robot robot, int echoMissionId, IsarMissi
             if (isarMissionResponse is null)
             {
                 _logger.LogError("Could not read content from mission");
-                throw new JsonException("Failed to deserialize mission from Isar");
+                throw new JsonException("Failed to deserialize mission from ISAR");
```
GodVenn, line 59: These two messages should match

backend/api/Services/IsarService.cs
```diff
@@ -79,11 +79,34 @@ public async Task<Report> StartMission(Robot robot, int echoMissionId, IsarMissi
...
+                _logger.LogError("Could not read content from mission");
+                throw new MissionException("Could not read content from mission");
+            }
+
+            var isarMissionResponse =
+                await response.Content.ReadFromJsonAsync<IsarStopMissionResponse>();
+            if (isarMissionResponse is null)
+            {
+                _logger.LogError("Could not read content from mission");
+                throw new JsonException("Failed to deserialize mission from ISAR");
```
GodVenn, line 106: These two messages should match

backend/api/Services/IsarService.cs
```diff
@@ -98,8 +121,21 @@ public async Task<HttpResponseMessage> PauseMission(Robot robot)
...
+                throw new MissionException("Could not read content from mission");
+            }
 
-            return response;
+            var isarMissionResponse =
+                await response.Content.ReadFromJsonAsync<HttpResponseMessage>();
+            if (isarMissionResponse is null)
+            {
+                _logger.LogError("Could not read content from mission");
+                throw new JsonException("Failed to deserialize mission from ISAR");
```
GodVenn, line 135: These two messages should match

backend/api/Controllers/RobotController.cs
```diff
@@ -214,26 +214,43 @@ public async Task<ActionResult<Robot>> PostRobot([FromBody] Robot robot)
...
 
             _logger.LogError(e, "Error getting mission from Echo");
-            return new StatusCodeResult(StatusCodes.Status502BadGateway);
+            return StatusCode(StatusCodes.Status502BadGateway, $"{e.Message}");
         }
         catch (JsonException e)
         {
             _logger.LogError(e, "Error deserializing mission from Echo");
-            return new StatusCodeResult(StatusCodes.Status500InternalServerError);
+            return StatusCode(StatusCodes.Status500InternalServerError, $"{e.Message}");
```
GodVenn, line 222: I think in this case (the case of json errors), the exception message should not be passed to the user as it won't help them understand what went wrong and we are not sure what it might contain. We could instead simply pass a message saying that something went wrong during the processing of the response from ISAR.
The same goes for the other endpoints

backend/api/Controllers/RobotController.cs
```diff
@@ -214,26 +214,43 @@ public async Task<ActionResult<Robot>> PostRobot([FromBody] Robot robot)
...
+        }
+        catch (HttpRequestException e)
+        {
+            _logger.LogError(e, "Error connecting to ISAR while starting mission");
+            robot.Enabled = false;
+            robot.Status = RobotStatus.Offline;
             await _robotService.Update(robot);
-            return Ok(report);
+
+            return StatusCode(StatusCodes.Status502BadGateway, $"{e.Message}");
```
GodVenn, line 237: Maybe here we could return our own message since we know what the issue is (ISAR is offline)
Same for other endpoints

GodVenn (commented):
Really nice! I like the approach, very clean and understandable now.
I just have some comments on what we return as the message, let me know what you think.

backend/api/Controllers/RobotController.cs
sondreo: Agreed that the exception message might be generic and not necessarily helpful to users. I reworded the messages relating to `JsonException` for the relevant endpoints.

backend/api/Controllers/RobotController.cs
sondreo: Good idea!

GodVenn (approved):
LGTM, Great job!

PR #303, opened 2022-07-27
Pose

### LINKED ISSUE #295: Add Pose to Robot model, and update it with MQTT
Every robot should have a pose object, such that its pose can be published to frontend.
The pose should be updated from Isar.

Add the Pose in backend database, and link to Robot.
There should be a default pose initialized.
Connect to Mqtt topic.

More information will be exposed in frontend.

MortFred:
Closes #295


backend/api/Database/Models/Robot.cs
```diff
@@ -51,12 +53,24 @@ public string IsarUri
                 return $"http://{host}:{Port}";
             }
         }
+        public Robot()
+        {
+            Name = "defaultId";
+            Model = "defaultModel";
+            SerialNumber = "defaultSerialNumber";
+            Status = RobotStatus.Busy;
```
GodVenn, line 61: Maybe default should be "offline" since it is disabled

backend/api/Services/RobotService.cs
```diff
@@ -24,6 +24,16 @@ public RobotService(FlotillaDbContext context)
             _context = context;
         }
 
+        private IQueryable<Robot> GetRobotObject()
```
GodVenn, line 27: Consider renamin to more descriptive name like `GetRobotsWithSubModels`

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -245,5 +247,38 @@ private async void OnBatteryUpdate(object? sender, MqttReceivedArgs mqttArgs)
...
+
+            robotPose.Position.X = isarPose.position.x;
+            robotPose.Position.Y = isarPose.position.y;
+            robotPose.Position.Z = isarPose.position.z;
+
+            robotPose.Orientation.X = isarPose.orientation.x;
+            robotPose.Orientation.Y = isarPose.orientation.y;
+            robotPose.Orientation.Z = isarPose.orientation.z;
+            robotPose.Orientation.W = isarPose.orientation.w;
+        }
```
GodVenn, line 281: I think this belongs on the IsarPose class as a method `ToRobotPose()` returning a `Pose`, rather than in the MQTT event handler

GodVenn (commented):
Nice!

backend/api/EventHandlers/MqttEventHandler.cs
MortFred: I agree, however we need to send in the `Pose robotPose` argument, since if we let `ToRobotPose()` just return a new pose, then it will be missing the correct id that is mapped in the databae. 
Hence, we need to send in the database robot object and copy the values directly.
We can change this when we achieve saving of pose through JSON serialization

GodVenn: Ah, that makes sense, good catch.

.gitignore
```diff
@@ -14,3 +14,7 @@ frontend/build/
 # Backend
 backend/*/bin
 backend/*/obj
+
+#Broker
+broker/mosquitto/config/certs/server-key.pem
+*.DS_Store
```
GodVenn, line 20: newline

GodVenn (approved):
Good job!:)
Just add the newline and I approve.

PR #307, opened 2022-07-29
Radix deployment

### LINKED ISSUE #300: CI & CD to Radix for Backend, Broker and Frontend components
We should complete and follow up on our Continuous Integration and Deployment to the Radix cluster in order to facilitate manual end-to-end testing as a quick and easy process as well as making it available to our users at an early stage, giving them more options to be involved in the process.

A CI/CD pipeline deploying to Radix (This can all be configured in Radix) which is fully functional (backend,broker and frontend) and which gives us alerts for errors.

All efforts to facilitate continuous testing and monitoring increase integrity and by consequence security.

GodVenn:
Closes #300


radixconfig.yaml
```diff
@@ -0,0 +1,43 @@
...
+  environments:
+    - name: dev
+      build:
+        from: main
+    - name: qa
+    - name: prod
+  components:
+    - name: frontend
+      src: frontend
+      dockerfileName: Dockerfile_Radix
```
oysand, line 18: Should it be called DockerfileRadix without the _?

GodVenn: Maybe, i was thinking that since a normal dockerfile is simply called `Dockerfile`, it would make it clearer what this was if I divided it more clearly like this

oysand: Could we then just call it Radix as Dockerfile is implied by dockerfileName?

oysand: Just realized that it points to the file Dockerfile_Radix... The way I have seen it earlier have been to call it Dockerfile.radix?

GodVenn: Will change!

radixconfig.yaml
```diff
@@ -0,0 +1,43 @@
...
+  environments:
+    - name: dev
+      build:
+        from: main
+    - name: qa
+    - name: prod
+  components:
+    - name: frontend
+      src: frontend
+      dockerfileName: Dockerfile.Radix
```
oysand, line 18: Do we need a seperate dockerfile for the radix or can we use/adapt the existing dockerfile for the frontend?

oysand: Or can I assume that is what is explained at line 8 in the file?

GodVenn: You may assume that yes

radixconfig.yaml
GodVenn: fixed

frontend/src/config.ts
```diff
@@ -10,5 +10,5 @@ export const config = {
     BACKEND_URL: GetEnvVariable('REACT_APP_BACKEND_URL'),
     FRONTEND_URL: GetEnvVariable('REACT_APP_FRONTEND_URL'),
     AD_CLIENT_ID: GetEnvVariable('REACT_APP_AD_CLIENT_ID'),
-    AD_TENANT_ID: GetEnvVariable('REACT_APP_AD_TENANT_ID'),
+    AD_TENANT_ID: '3aa4a235-b6e2-48d5-9195-7fcf05b459b0',
```
oysand, line 13: Keep as env variable?

PR #311, opened 2022-08-02
Configure environment variables for frontend

### LINKED ISSUE #306: Frontend doesn't have environment-specific variables
**Describe the bug**
We currently have no support for environment-specific variables for the frontend.
This will only work as is as long as we only have one environment active, but needs resolving before we can move further.

**To Reproduce**
Try to change all components to a different environment (-> different URL's, different Azure AD clients)
Behold how we need to hardcode the frontend because we don't have a flexible solution

**Expected behavior**
Environment variables being dependent on the current environment in the frontend (could be read from NODE_ENV for example.)

GodVenn:
Closes #306 
This PR configures an environment-variable setup which allows for environment specific variables to be defined and used.
It does, however, also complicate the process of adding new environment variables: They now need to be defined in 3 files:
`run_nginx.sh`
`index.html`
`config.ts`
(and workflow `frontend_lint_and_test.yml` for tests)


frontend/Dockerfile
```diff
@@ -3,14 +3,32 @@ FROM node:current-slim as build
 WORKDIR /app
 
 COPY . .
+
+# The local .env is copied for local testing, this is misleading as it will not be included when deployed
+RUN rm -f .env
```
oysand, line 8: Put in dockerignore

oysand (approved):
LGTM

PR #318, opened 2022-08-08
Fullscreen video

oysand:


frontend/src/components/VideoStream/VideoSection.tsx
```diff
@@ -0,0 +1,37 @@
...
+    height: 32px;
+    opacity: 0.75;
+`
+
+Icon.add({ fullscreen })
+
+export function VideoSection({ videoStream, toggleFullScreenModeFunction }: VideoSectionProps) {
+    return (
+        <StyledVideoSection>
+            <img src={videoStream.url} width="100%" height="100%" alt="robot video stream" />
```
GodVenn, line 31: Maybe include the `videostream.name` in the `alt` property?

frontend/src/components/VideoStream/VideoStreamWindow.tsx
```diff
@@ -1,50 +1,53 @@
...
+                isOpen={fullScreenMode}
+                onRequestClose={() => {
+                    setFullScreenMode(false)
+                }}
+            >
+                <VideoCard variant="default" style={{ boxShadow: tokens.elevation.sticky }}>
+                    <Typography variant="h5">{videoStream.name}</Typography>
+                    <VideoSection videoStream={videoStream} toggleFullScreenModeFunction={toggleFullScreenMode} />
+                </VideoCard>
+            </VideoFullScreen>
```
GodVenn, line 50: Won't this display both video cards (although one will be on top) since there is no check for the `fullScreenMode` on the normal video card?

GodVenn (commented):
Please see comments

frontend/src/components/VideoStream/VideoSection.tsx
oysand: Agreed

frontend/src/components/VideoStream/VideoStreamWindow.tsx
oysand: That is true, decided to have all video streams as small in the background as described in the UX design: 
image
Should it be much to handle all the video streams at once we could disable the ones not in full screen mode, but i propose that we add this if it proves to be an issue later

GodVenn: Sounds good!

GodVenn (approved):
LGTM

PR #321, opened 2022-08-10
Update upcomingMissions at beginning of loop

eivindsjovold:


backend/api/EventHandlers/ScheduledMissionEventHandler.cs
```diff
@@ -47,6 +46,7 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
         {
             while (!stoppingToken.IsCancellationRequested)
             {
+                _upcomingScheduledMissions = await _scheduledMissionService.ReadByStatus(ScheduledMissionStatus.Pending);
```
GodVenn, line 49: This fixes the current issue because this is the only place where the missions are used, but what if `_upcomingScheduledMissions` is referenced anywhere else in the class later? 
We won't have a guarantee it is updated then.
Suggesting to move the database call `await _scheduledMissionService.ReadByStatus(ScheduledMissionStatus.Pending)` to the `_upcomingScheduledMissions` parameter itself like an [Expression bodied member](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/expression-bodied-members)

GodVenn (approved):
LGTM

aestene:
Please remove the punctuation "." at the end of some of your commits for conformity :D

PR #322, opened 2022-08-10
Use remote database for staging

aestene:
This pull request does the following
- Implement use of remote SQL database in staging environment
- Enable distribution of database contexts to background services for parallel database operations

Note that everything in the `Migrations` folder is autogenerated.


backend/api/Configurations/CustomServiceConfigurations.cs
```diff
@@ -10,15 +10,18 @@ public static class CustomServiceConfigurations
...
+            IConfiguration configuration
         )
         {
-            string sqlConnectionString = configuration.GetSection("Database").GetValue<string>("ConnectionString");
-            if (isDevelopment)
+            bool useInMemoryDatabase = configuration
+                .GetSection("Database")
+                .GetValue<bool>("UseInMemoryDatabase");
+
+            if (useInMemoryDatabase)
```
aestene, line 20: This is the only real change in this file, the rest is formatting.

backend/api/Program.cs
```diff
@@ -69,7 +69,7 @@
 var app = builder.Build();
 
 // Configure the HTTP request pipeline.
-if (app.Environment.IsDevelopment())
+if (app.Environment.IsDevelopment() || app.Environment.IsStaging())
```
GodVenn, line 72: Maybe this should be for all environments? I am assuming we want swagger for documentation in Production as well

backend/api/appsettings.Staging.json
```diff
@@ -0,0 +1,41 @@
...
+      "isar/+/step",
+      "isar/+/battery",
+      "isar/+/pose"
+    ],
+    "MaxRetryAttempts": 15,
+    "ShouldFailOnMaxRetries": true
+  },
+  "Database": {
+    "UseInMemoryDatabase": false,
+    "ConnectionString": ""
```
GodVenn, line 39: Should this be here?

backend/api/Database/Models/Pose.cs
```diff
@@ -8,6 +8,7 @@ public class Orientation
     {
         [Key]
         [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
+        [Required]
```
GodVenn, line 11: The `Key` attribute will make the property be non-null in the database always, adding the `required` attribute does not change anything database-wise, but complicates our code because it forces us to specify separate `query models` to avoid errors when receiving models without explicit ID's through the API.
Unless there was a specific other reason for adding this attribute, I suggest removing it and thus removing the need for more "duplicate" model classes.
(Same for the other model classes)

GodVenn (commented):
Looks good! Please see comments.

backend/api/Program.cs
aestene: I think we will change this part slightly as we evolve. There will probably be separate sections for each environment so I would say we just add it when we need it.

backend/api/appsettings.Staging.json
aestene: Yes, it's retrieved from the keyvault. It's collected from keyvault when using the staging environment as that will use the azure database.

backend/api/Database/Models/Pose.cs
aestene: I'm not certain this would remove the need for the query models we are currently using. If we don't use those models the user friendliness of the endpoints will be worse as you'll have to remove multiple sections. For example ID and some others for specific models.

As for the database issue me and @Christdej had an issue which we solved by setting required as the database was throwing an error specifically linked to that property in that model. I can test again to see if the issue is still there.

backend/api/appsettings.Staging.json
GodVenn: But if it's retrieved from keyvault it's not set in this file so maybe having it there is misleading as we really should never fill in the value?

aestene: My understanding was that having it here is what caused it to be retrieved automatically from the keyvault but maybe I've misunderstood the functionality?

GodVenn: Ah, no, it doesn't have to be there, the keyvault is loaded during startup and when we retrieve values later it simply looks in the keyvault as one of the lookup sources, it doesn't have to be defined here

backend/api/Database/Models/Pose.cs
aestene: More specifically the problem was related to the automatically generated first migration not working when setting up the database. If I remove the required option now it works fine but I'm not certain it would work if we built the database from scratch again. Not entirely certain what to do about it to be honest.

GodVenn: Agreed to leave it for now and maybe get back to it in the future 👍

anetteu (approved):
Good work!

PR #333, opened 2022-08-24
Use downstream api for calls and authentication to ISAR

### LINKED ISSUE #332: Add bearer token to ISAR request for new mission
Currently the `StartMission` function does not add a bearer token which authenticates to the ISAR API. This means Flotilla is unable to start a mission in ISAR if ISAR has authentication turned on. Flotilla should pass the bearer token from its own service principal which should have access to the ISAR service principle.

Functionality will be the same but Flotilla can use ISAR with authentication turned on.

Will be the same as authentication is already in place.

GodVenn:
Closes #332


aestene (approved):
LGTM

PR #354, opened 2022-09-20
Task table

### LINKED ISSUE #334: Finish task table in frontend
The task table from the misssion page should be filled with information, both when the mission is scheduled and when the mission is ongoing.

image

Need to slightly refactor backend APIs. Then, we can call the mission endpoint to get this data.

More information will become available in frontend.

MortFred:
Closes #334


tsundvoll (commented):
Looks good to me, but I don't have that much insights here yet, so I recommend someone else having a look.

Created issue #356 regarding the difference between a ISAR task and an Echo tag / task.

frontend/src/components/MissionOverview/UpcomingMissionCard.tsx
```diff
@@ -26,13 +27,23 @@ const StyledMissionStartTime = styled.div`
...
+            variant="default"
+            style={{ boxShadow: tokens.elevation.raised }}
+            onClick={routeChange}
+        >
             <HorizontalContent>
                 <Checkbox />
                 <Typography variant="h6">Mission name</Typography>
                 <MissionStartTimeDisplay mission={mission} />
-                <Typography>Tags: 17</Typography>
+                <Typography>Tags: {mission.plannedTasks.length}</Typography>
```
GodVenn, line 46: The time showed on the cards is wrong.
The time that is sent from backend is in the UTC timezone, so we should either specify this in the card or convert to the local timezone.

See difference in time in screenshots:

Frontend:
image

From Backend (in UTC):
image

GodVenn, line 46: Also, a bit confusing to say "Tags: 1" and then the mission containts 2 tasks (because there are two inspections on the same tag)

GodVenn (commented):
Nice job, I like the table, just have a few comments.
Also, in the future, maybe the mission overview should be one row per TAG and the rows are expandable to include information on each inspection for the TAG?

frontend/src/api/ApiCaller.tsx
```diff
@@ -84,14 +92,6 @@ export class BackendAPICaller {
...
     }
 
-    async getReports(asset: string): Promise<Report[]> {
-        const path: string = 'reports?assetCode=' + asset
-        const result = await this.GET<Report[]>(path).catch((e) => {
-            throw new Error(`Failed to GET /${path}: ` + e)
-        })
-        return result.body
-    }
-
```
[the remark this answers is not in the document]
MortFred: We don't have the need for a get(all)Missions just yet, so I thought we could add the API call when it is needed.

frontend/src/components/MissionOverview/UpcomingMissionCard.tsx
MortFred: Created issue #357

MortFred: Changed to "Tasks: 2"

PR #365, opened 2022-09-22
Update setup script

aestene:


GodVenn (approved):
LGTM!
The only thing changed in the setup is that you can cancel each setup independently of each other?

Issue #367, opened 2022-09-23
Cancel the mission the mission if the placement of the robot cannot be determined.


oysand:
**Describe the improvement you would like to see**
If the mission points cannot be planned with both placement of the robot and the location of the equipment the mission should fail before being sent to ISAR.

**How will this change existing functionality?**
Currently when there is not found a position to place the robot we just return a dummypose at the origin.
![image](https://user-images.githubusercontent.com/32652118/191986301-06950134-cdab-4ab4-b165-764c9c29fb78.png)

**How will this improvement affect the current Threat Model?**
No change

PR #369, opened 2022-09-23
Implement predefined pose workaround

oysand:


oysand:
Handling of the case where the TAG is not present in the workaround file is to be addressed in #367.

oysand:
> Handling of the case where the TAG is not present in the workaround file is to be addressed in #367.

Implemented in this PR

backend/api/Utilities/TagPositioner.cs
```diff
@@ -13,10 +14,24 @@ public class TagPositioner
...
+
+            var predefinedPose = predefinedPoses.Find(x => x.Tag == tag.TagId);
+
+            if (predefinedPose == null)
+                throw new RobotPositionNotFoundException($"Could not find tag {tag.TagId} in the workaround file");
+
+            IsarPosition position = new(predefinedPose.Pose.Position.X, predefinedPose.Pose.Position.Y, predefinedPose.Pose.Position.Z, predefinedPose.Pose.Frame);
+            IsarOrientation orientation = new(predefinedPose.Pose.Orientation.X, predefinedPose.Pose.Orientation.Y, predefinedPose.Pose.Orientation.Z, predefinedPose.Pose.Orientation.W, predefinedPose.Pose.Frame);
+
+            return new IsarPose(position, orientation, predefinedPose.Pose.Frame);
```
GodVenn, line 33: Suggesting implementing a constructor for `IsarPose` taking a `PredefinedPose` as argument instead of this.

backend/api/Controllers/RobotController.cs
```diff
@@ -257,6 +257,12 @@ public async Task<ActionResult<Robot>> PostRobot([FromBody] CreateRobotQuery rob
             _logger.LogError(e, "{message}", message);
             return StatusCode(StatusCodes.Status500InternalServerError, message);
         }
+        catch (RobotPositionNotFoundException e)
+        {
+            string message = "A suitable robot position could not be found";
```
GodVenn, line 262: "... for one or more of the desired tags"
Also this means that the entire mission will fail if one tag is missing, is this desired functionality?

PR #371, opened 2022-09-26
Development branch

oysand:


backend/api/Utilities/TagPositioner.cs
```diff
@@ -13,10 +14,24 @@ public class TagPositioner
...
+
+            var predefinedPose = predefinedPoses.Find(x => x.Tag == tag.TagId);
+
+            if (predefinedPose == null)
+                throw new RobotPositionNotFoundException($"Could not find tag {tag.TagId} in the workaround file");
+
+            IsarPosition position = new(predefinedPose.Pose.Position.X, predefinedPose.Pose.Position.Y, predefinedPose.Pose.Position.Z, predefinedPose.Pose.Frame);
+            IsarOrientation orientation = new(predefinedPose.Pose.Orientation.X, predefinedPose.Pose.Orientation.Y, predefinedPose.Pose.Orientation.Z, predefinedPose.Pose.Orientation.W, predefinedPose.Pose.Frame);
+
+            return new IsarPose(position, orientation, predefinedPose.Pose.Frame);
```
GodVenn, line 33: Suggesting implementing a constructor for `IsarPose` taking a `PredefinedPose` as argument instead of this.

backend/api/Controllers/RobotController.cs
```diff
@@ -257,6 +257,12 @@ public async Task<ActionResult<Robot>> PostRobot([FromBody] CreateRobotQuery rob
             _logger.LogError(e, "{message}", message);
             return StatusCode(StatusCodes.Status500InternalServerError, message);
         }
+        catch (RobotPositionNotFoundException e)
+        {
+            string message = "A suitable robot position could not be found";
+            _logger.LogError(e, "{message}", message);
+            return StatusCode(StatusCodes.Status500InternalServerError, message);
```
GodVenn, line 264: "... for one or more of the desired tags"
Also this means that the entire mission will fail if one tag is missing, is this desired functionality?

GodVenn (commented):
See comments

backend/api/Controllers/RobotController.cs
oysand: Ideally we'd like to know which points are supported when creating the mission. This will be possible with the echo mission planner once the database for the robot positions are in place. Then we can enforce that all Ready echo missions only contain valid tags. I suggest that we await that implementation and accept that the entire mission will fail in the meantime.

GodVenn: Agreed

backend/api/Utilities/TagPositioner.cs
oysand: Agreed

GodVenn (approved):
LGTM

PR #375, opened 2022-09-28
Filter away Echo missions that are not ready

oysand:


oysand:
Older Echo missions still have the status "Sent to Robot". I propose to only support the newer missions tagged as "Ready".

PR #377, opened 2022-09-30
Pause, start, stop mission, continuation

eivindsjovold:


frontend/src/components/MissionOverview/OngoingMissionCard.tsx
```diff
@@ -27,13 +31,77 @@ export function OngoingMissionCard({ mission }: MissionProps) {
...
+                    <Icon
+                        name="stop"
+                        style={{ color: tokens.colors.interactive.primary__resting.rgba }}
+                        size={32}
+                        onClick={() => handleClick(ControlButton.Stop)}
+                    />
+                </>
+            )
+        }
+        return (
```
oysand, line 96: Should we not be able to stop the mission without pausing it first?

oysand:
#381 Handle the updates of upcoming missions. If this is merged before this PR the paused missions should be included in the periodic update. If this PR is merged first it should be updated in #381

oysand:
The design also lists control buttons on the single mission view. Should we add this as a separate issue?

frontend/src/api/ApiCaller.tsx
```diff
@@ -125,6 +126,27 @@ export class BackendAPICaller {
         })
         return result.body
     }
+
+    async pauseMission(robotId: string) {
+        const path: string = 'robots/' + robotId + '/pause'
+        await this.POST(path, '').catch((e) => {
+            throw new Error(`Failed to POST /${path}: ` + e)
```
oysand, line 133: If the control button post fails, should we force an update of the mission status? For instance if pause fail, then the pause button should be the option rather than the play button?

oysand: This could also be related to #381, but require a faster update rate

oysand (changes requested):
Overall looks good, see comments

frontend/src/components/MissionOverview/OngoingMissionCard.tsx
```diff
@@ -27,13 +30,58 @@ export function OngoingMissionCard({ mission }: MissionProps) {
...
             </HorisontalContent>
         </StyledMissionCard>
     )
 }
+
+function MissionControlButtons({mission}: MissionProps) {
+    const [status, setStatus] = useState(MissionStatus.Ongoing);
+    enum ControlButton {Pause, Stop, Play}
+    
+    const handleClick = (button: ControlButton) => {
```
GodVenn, line 57: Perhaps, rather than setting the status immediately in the frontend, this should only ever display the actual status of the mission coming from the robot?
We could implement a loading symbol for when the operator pushes a button while waiting for the backend to confirm that the action has been processed.

frontend/src/components/MissionOverview/OngoingMissionCard.tsx
```diff
@@ -31,57 +32,76 @@ export function OngoingMissionCard({ mission }: MissionProps) {
...
+            )
         }
         return (
-            <Icon name="pause_circle" style={{ color: tokens.colors.interactive.primary__resting.rgba}} size={32} onClick={() => handleClick(ControlButton.Pause)}/>
+            <Icon
+                name="pause_circle"
+                style={{ color: tokens.colors.interactive.primary__resting.rgba }}
+                size={32}
+                onClick={() => handleClick(ControlButton.Pause)}
+            />
```
GodVenn, line 102: Maybe we should group these Icons into named components "PauseButton", "PlayButton" and "StopButton"?

frontend/src/components/MissionOverview/OngoingMissionView.tsx
```diff
@@ -20,13 +21,22 @@ const OngoingMissionSection = styled.div`
...
             setOngoingMissions(missions)
         })
+        apiCaller.getMissionsByStatus(MissionStatus.Paused).then((pausedMissions) => {
+            setPausedMissions(pausedMissions)
+        })
     }, [])
+    useEffect(() => {
+        const missions: Mission[] = ongoingMissions.concat(pausedMissions)
+        setMissionsToDisplay(missions)
+    }, [ongoingMissions, pausedMissions])
```
GodVenn, line 37: All this is a symptom of the missing "filter functionality" in the backend. It should be possible to query the backend with a filter on several mission statuses, and thus it should not be necessary to do this double collection and concat and stuff

GodVenn: But for this PR this is a good solution!

GodVenn (commented):
Please see comments

frontend/src/components/MissionOverview/OngoingMissionCard.tsx
MortFred: I think it is natural to have a two step process here, since canceling a mission shouldn't be done on a whim. During the user test, no users had any problem finding out how to stop the mission

frontend/src/components/MissionOverview/OngoingMissionCard.tsx
MortFred: Probably a good idea

MortFred: Created an issue on it #386

frontend/src/api/ApiCaller.tsx
MortFred: Will be fixed by having the frontend show the true backend mission state, as described in #386

eivindsjovold:
Closed by #428

PR #381, opened 2022-10-05
Periodically poll backend for updates

oysand:


oysand:
If #377 is merged before this, paused missions needs to be inculded in the periodic update in this PR.

LeikvollE (approved):

PR #395, opened 2022-10-12
Update task table periodicly

oysand:


frontend/src/components/Pages/MissionPage.tsx
```diff
@@ -34,6 +34,20 @@ export function MissionPage() {
...
+        const id = setInterval(() => {
+            if (missionId) {
+                apiCaller.getMissionById(missionId).then((mission) => {
+                    setSelectedMission(mission)
+                    apiCaller.getVideoStreamsByRobotId(mission.robot.id).then((streams) => {
+                        setVideoStreams(streams)
+                    })
+                })
+            }
+        }, 1000)
```
tsundvoll, line 47: I see a magic number, can we first assign this to a const with an explaining name and then use the const here instead?

frontend/src/components/Pages/MissionPage.tsx
```diff
@@ -34,6 +34,20 @@ export function MissionPage() {
...
+    useEffect(() => {
+        const id = setInterval(() => {
+            if (missionId) {
+                apiCaller.getMissionById(missionId).then((mission) => {
+                    setSelectedMission(mission)
+                    apiCaller.getVideoStreamsByRobotId(mission.robot.id).then((streams) => {
+                        setVideoStreams(streams)
+                    })
+                })
+            }
```
tsundvoll, line 46: This is duplicated code from the above `useEffect`. Is it possible to reuse something here?

tsundvoll (changes requested):

tsundvoll (approved):
LGTM, check for conflicts before rebasing

PR #415, opened 2022-10-17
Fix robot images

### LINKED ISSUE #403: Add picture of robots in front-end
Add picture of the robots under 'Robot status' in front-end.

Show a picture of each robot.
image

Does not affect.

aestene:
Closes #403 

image


backend/api/Database/Models/RobotModel.cs
```diff
@@ -0,0 +1,12 @@
...
+{
+    public enum RobotModel
+    {
+        TaurobInspector,
+        TaurobOperator,
+        ExR2,
+        Turtlebot,
+        AnymalX,
+        AnymalD,
+    }
```
Christdej, line 11: Should we add a "default" or generic "robot" case? If we want to test something or use the isar-robot repo which is none of these robots.

Christdej (changes requested):

backend/api/Database/Models/RobotModel.cs
aestene: I decided to use the `Turtlebot` as the default since that is our simulator and the one used for testing rather than have a default name `Robot` or similar.

aestene: So `Turtlebot` is used in mocks and tests etc.

Christdej: Yeah, agreed

Christdej (changes requested):
Make a commit with the DB migrations in same PR?

Christdej (approved):

PR #421, opened 2022-10-18
Add robot name to mission views

oysand:


tsundvoll (approved):
Does it look good to me? Yes :)

PR #424, opened 2022-10-18
Hide offline robots from view

aestene:


oysand:
Are we sure we want the offline robots to be hidden?

eivindsjovold (approved):
LGTM

oysand:
I'd argue we want to hide robots not associated to the asset

GodVenn:
> I'd argue we want to hide robots not associated to the asset

I tend to agree with this

eivindsjovold:
Could alternatively be user choice to hide offline robots.

aestene:
I'm leaning towards a solution where we have offline and they may be showed. Let's say a robot is in operation but currently not powered on so ISAR is unable to communicate with it. Then it would be good to show a robot as offline. 

In addition to this we should add another status which is `Inactive` or `Deactivated` etc. which is used to indicate a robot is not considered interesting for the frontend but of course it should not be deleted from the database.

Thoughts @eivindsjovold, @GodVenn, @oysand?

aestene:
See also this issue: https://github.com/equinor/flotilla/issues/426

oysand:
> I'm leaning towards a solution where we have offline and they may be showed. Let's say a robot is in operation but currently not powered on so ISAR is unable to communicate with it. Then it would be good to show a robot as offline.
> 
> In addition to this we should add another status which is `Inactive` or `Deactivated` etc. which is used to indicate a robot is not considered interesting for the frontend but of course it should not be deleted from the database.
> 
> Thoughts @eivindsjovold, @GodVenn, @oysand?

Agreed, we could have a status Deactivated or something like that

aestene:
This pull request has automatically been marked as stale as there has been no activity for 30 days. Remove stale label or comment or this will be closed in 7 days.

PR #435, opened 2022-10-20
Setup application insights

### LINKED ISSUE #434: Log from backend to Azure Application Insights
We need a way to monitor the use of our application and maintain it long-term.
A good way to do this is to use Azure Application Insights, which also manages logs.

Link the current logging to Application Insights (should be easy to configure in ASP.NET)

Will give us an overview of who and how our application is being used as well as give us somewhere to store and analyze our logs.

GodVenn:
Closes #434

### DIFF
--- a/backend/api/api.csproj
+++ b/backend/api/api.csproj
@@ -12,6 +12,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.20.0" />
     <PackageReference Include="Azure.Extensions.AspNetCore.Configuration.Secrets" Version="1.2.2" />
     <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="6.0.4" />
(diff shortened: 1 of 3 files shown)


backend/api/api.csproj
```diff
@@ -12,6 +12,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.20.0" />
     <PackageReference Include="Azure.Extensions.AspNetCore.Configuration.Secrets" Version="1.2.2" />
```
Christdej, line 16: Shouldnt a section be added to the appsettings.json aswell?
"ApplicationInsights": {
    "ConnectionString": "Copy connection string from Application Insights Resource Overview"
  }

Christdej (changes requested):

backend/api/api.csproj
GodVenn: It's fetched automatically from the keyvault

Christdej: Cool

Christdej (approved):

PR #436, opened 2022-10-21
Fix readme

tsundvoll:


GodVenn (approved):
LGTM

backend/README.md
```diff
@@ -49,7 +49,7 @@ See [Using the “--env-file” option](https://docs.docker.com/compose/environm
 To run the backend in docker, run the following command in the root folder of flotilla:
 
 ```
-docker compose up backend --build
+docker-compose up --build backend
```
oysand, line 52: Whether we use docker compose or docker-compose depends on if we have a newer version of docker compose

tsundvoll: Ah, I didn't know. I see we are using the `docker-compose` format in the front-page README of flotilla, so at least we should be consistent here. Any thoughts on which format we should choose?

tsundvoll: Let's go for the old format for now, and we can update then all later if we want.

PR #444, opened 2022-10-31
Change time format to Norwegian

EinarUeland:


frontend/src/components/MissionOverview/UpcomingMissionCard.tsx
```diff
@@ -113,7 +114,13 @@ function MissionStartTimeDisplay({ mission }: MissionDisplayProps) {
     return (
         <StyledMissionStartTime>
             <Typography>Date: {format(new Date(mission.startTime), 'dd. MMM')}</Typography>
-            <Typography>Time: {format(new Date(mission.startTime), 'hh:mm')}</Typography>
+            <Typography>
+                Time:{' '}
+                {new Date(mission.startTime).toLocaleTimeString('no-NO', {
```
EinarUeland, line 119: This could also use the "default time", but chose to use Norwegian

GodVenn (commented):
Is it possible to get the timezone based on where the frontend is being called from, i.e. where the user is located?
Or maybe the timezone should be based on the location of the plant?
This might make sense to change later when we get more info from the fusion context, maybe they have timezones.

EinarUeland:
> Is it possible to get the timezone based on where the frontend is being called from, i.e. where the user is located? Or maybe the timezone should be based on the location of the plant? This might make sense to change later when we get more info from the fusion context, maybe they have timezones.

frontend/src/components/MissionOverview/UpcomingMissionCard.tsx
GodVenn: Good choice

EinarUeland:
> Is it possible to get the timezone based on where the frontend is being called from, i.e. where the user is located? Or maybe the timezone should be based on the location of the plant? This might make sense to change later when we get more info from the fusion context, maybe they have timezones.

The default is to use local clock. In this case I have specified "Norwegian time".  There are pros and cons to specifying time-zone or letting the users system determine it. For now I think using Norwegian makes sense.

Also - I just checked another Equinor app. Here the correct time is displayed in the planning window though a digital clock. I think this is a decent method to avoid all clock errors and is something we might consider for later.

I'm also noting that currently the time-field does not give much sense as it just displays the time the mission was created in.

aestene:
This pull request has automatically been marked as stale as there has been no activity for 30 days. Remove stale label or comment or this will be closed in 7 days.

PR #447, opened 2022-11-03
Remove single quotes from server key

Christdej:
Fix where in some cases server key is created with single quotes


GodVenn (approved):
LGTM

PR #455, opened 2022-12-08
Display robot pose in Frontend

### LINKED ISSUE #304: Present Robot Pose in Frontend
The robot pose should be presented to the user when entering the robot view page.
If only for debug purposes in the beginning.

Add a box in mission/robot page that list the pose.
Should also edit isar_robot to publish random pose values, so that we can verify changes.

More data will be exposed to the user.

tsundvoll:
Resolves  #304

### DIFF
--- a/frontend/src/components/MapPosition/MapPositionView.tsx
+++ b/frontend/src/components/MapPosition/MapPositionView.tsx
@@ -1,0 +1,76 @@
+import { Typography } from '@equinor/eds-core-react'
+import { Mission } from 'models/Mission'
+import { useEffect, useState } from 'react'
+import styled from 'styled-components'
+
+interface MissionProps {
+    mission?: Mission
+}
+
+const MapPositionSection = styled.div`
+    display: flex;
+    flex-wrap: wrap;
+    gap: 2rem;
+`
+
+const StyledPlaceholder = styled.div`
+    display: flex;
+    box-sizing: border-box;
+    flex-direction: row;
+    align-items: flex-start;
+    padding: 24px;
+    gap: 8px;
+
+    border: 1px solid #dcdcdc;
+    border-radius: 4px;
+
+    flex: none;
+    order: 1;
+    align-self: stretch;
+    flex-grow: 1;
+`
+
+function NoPositionPlaceholder() {
+    return (
+        <StyledPlaceholder>
+            <Typography variant="h4" color="disabled">
+                No robot position available
+            </Typography>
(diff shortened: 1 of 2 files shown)


frontend/src/components/MapPosition/MapPositionSection.tsx
```diff
@@ -0,0 +1,34 @@
...
+    gap: 1rem;
+`
+
+interface MissionProps {
+    mission?: Mission
+}
+
+export function MapPositionSection({ mission }: MissionProps) {
+    var text
+    if (mission && mission.robot.pose)
```
oysand, line 17: Should we wrap this in a 
useEffect(()={
updateText
}, []) 
to ensure that it only runs once? alternately 

useEffect(()={
updateText
}, [changingPoseVariable])

tsundvoll: Good question, let us think more about this.

frontend/src/components/MapPosition/MapPositionSection.tsx
```diff
@@ -0,0 +1,34 @@
...
+    display: grid;
+    grid-column: 1/ -1;
+    gap: 1rem;
+`
+
+interface MissionProps {
+    mission?: Mission
+}
+
+export function MapPositionSection({ mission }: MissionProps) {
```
oysand, line 15: Should we add the apiCaller and a new endpoint to get the position? That way each component calls for what it needs and decides how often it wants the information. I see that this also applies to the task table. So if we want this we should also add an issue for it

eivindsjovold (changes requested):

frontend/src/components/MapPosition/MapPositionSection.tsx
tsundvoll: As it is written now it works the same way as the tasks update. I would say that is fine for now, and will create an issue to do this through an API call directly as suggested in the next comment.

frontend/src/components/MapPosition/MapPositionSection.tsx
tsundvoll: That seems like a good approach. I have created issue #458 on this.

tsundvoll:
I think this approach works for now and we can take it in if you agree. New issue https://github.com/equinor/flotilla/issues/458 on doing this in a better way, also for other components.

eivindsjovold (approved):
:100:

PR #466, opened 2022-12-19
Add placeholder for missing videostream

### LINKED ISSUE #460: Add placeholder when videostream is missing
Similar to what is done with `No ongoing missions` placeholder, add a placeholder when the videostream is missing.

MortFred:
Closes #460

### DIFF
--- a/frontend/src/components/VideoStream/VideoSection.tsx
+++ b/frontend/src/components/VideoStream/VideoSection.tsx
@@ -1,8 +1,10 @@
 import { Button, Card, Icon } from '@equinor/eds-core-react'
-import { fullscreen } from '@equinor/eds-icons'
+import { fullscreen, videocam_off } from '@equinor/eds-icons'
 import styled from 'styled-components'
 
 import { VideoStream } from 'models/VideoStream'
+
+Icon.add({ fullscreen, videocam_off })
 
 interface VideoSectionProps {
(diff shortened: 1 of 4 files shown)


frontend/src/components/VideoStream/VideoSection.tsx
```diff
@@ -23,15 +25,37 @@ const FullscreenButton = styled(Button)`
...
+                )}
+            {!ValidateVideoStream(videoStream) && <NoVideoPlaceholder name="videocam_off" />}
         </StyledVideoSection>
     )
 }
+
+function ValidateVideoStream(videoStream: VideoStream) {
+    const stream = new Image()
+    stream.src = videoStream.url
+    if (stream.naturalHeight === 0) {
```
GodVenn, line 57: Is this a safe check for validating video stream?

MortFred: Yes, if the image is available then it must return naturalHeight > 0.
However, I don't have a video stream stack set up on my setup, so I won't be able to check for edge cases.
If we find some, we can always augment the validate function was my thinking.

GodVenn: Okidoki

GodVenn (approved):
LGTM

PR #478, opened 2022-12-29
Add error handling callback for login

GodVenn:
Added a callback function to handle (log) errors during login. 
For example if the wrong scope is requested, we will now see this in the console.


oysand (approved):
LGTM

PR #486, opened 2023-01-05
Functionality to choose mission start time

Eddasol:


eivindsjovold (approved):
Would prefer a different time-picker(not AM/PM), but I support using what EDS has. LGTM :)

PR #492, opened 2023-01-06
Require asset

Eddasol:
The variable assetString didn't update correctly. Solved by getting asset value directly


frontend/src/components/Pages/FrontPage/MissionOverview/UpcomingMissionView.tsx
```diff
@@ -93,19 +92,17 @@ export function UpcomingMissionView() {
...
-
     useEffect(() => {
         const id = setInterval(() => {
-            apiCaller.getEchoMissions(assetString).then((missions) => {
-                const mappedEchoMissions: Map<string, EchoMission> = mapEchoMissionToString(missions)
-                setEchoMissions(mappedEchoMissions)
-            })
+            const installationCode = sessionStorage.getItem('assetString')
+            if (!installationCode || installationCode == '') {
+                setEchoMissions(mapEchoMissionToString([]))
```
tsundvoll, line 99: Is this possible to do?
```suggestion
                setEchoMissions([])
```

Eddasol: this does not work, but changing to new Map<String, EchoMission> instead

MortFred (approved):

PR #493, opened 2023-01-06
Fix delayed render of "schedule mission"-button

Eddasol:
Instead of waiting for echoMissions and robotOptions to load before rendering the buttons, the "schedule mission" button render in a disabled state while waiting for the parameters to load


phhuseboe (approved):
LGTM

PR #506, opened 2023-01-13
Update robot based on availability in MQTT messages

aestene:
https://github.com/equinor/isar/issues/341


backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -42,7 +46,7 @@ public override void Subscribe()
 
         public override void Unsubscribe()
         {
-            MqttService.MqttIsarConnectReceived -= OnIsarConnect;
+            MqttService.MqttIsarConnectReceived -= OnIsarAvailability;
```
oysand, line 49: Should MqttIsarConnectReceived be renamed since it is a more continuous update?

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -55,35 +59,152 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
+                _logger.LogInformation(
+                    "Received message from new ISAR instance with robot name {name}. Adding new robot to database.",
+                    isarAvailability.RobotName
                 );
-                return;
+
+                var robotQuery = new CreateRobotQuery()
+                {
+                    Name = isarAvailability.RobotName,
+                    Model = RobotModel.Turtlebot,
```
oysand, line 77: The robot model should not be hard coded, should come from the robot or some other place?

backend/api/MQTT/MessageModels/IsarAvailability.cs
```diff
@@ -0,0 +1,49 @@
+using System.Text.Json.Serialization;
+using Api.Controllers.Models;
+using Api.Database.Models;
+
+namespace Api.Mqtt.MessageModels
+{
+#nullable disable
+    public class IsarAvailabilityMessage : MqttMessage
```
oysand, line 8: This seem to have much more values than I would assosiate with an AvialabilityMessage. Should we limit this to a update on just the status? The other values seem more to be set on the initial connect

backend/api/EventHandlers/MqttEventHandler.cs
oysand: Just saw line 39, rename accordingly

backend/api/MQTT/MessageModels/IsarAvailability.cs
aestene: The functionality is twofold in that it signals the availability of the ISAR instance but also enables Flotilla to discover new ISAR instances on the network. Thus all the information required to instantiate a robot object in the database must be available when the message is received. A suggestion from @Christdej (https://github.com/equinor/isar/pull/390) was to use an API call back to the ISAR instance to ask for more details instead. Currently undecided on which approach to go for.

backend/api/EventHandlers/MqttEventHandler.cs
aestene: Solved by including robot model in ISAR as an enum which is then translated to an enum in Flotilla.

backend/api/MQTT/MessageModels/IsarAvailability.cs
aestene: Ended up splitting in two topics with different event listeners for those two.

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -55,33 +60,160 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
+                        && videoStreamQuery.Url != existingVideoStream.Url
+                    )
+                        return true;
+                }
+            }
+
+            return false;
+        }
+
+        private async void OnTimedEvent(string robotName)
```
oysand, line 195: OnTimedOutEvent?

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -55,33 +60,161 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
+                robot = await RobotService.Update(robot);
+                _logger.LogInformation("Updated robot {name} in database", robot.Name);
+            }
+        }
+
+        private bool RobotHasSignificantChange(Robot robot, IsarRobotInfoMessage isarRobotInfo)
+        {
+            if (robot.Host != isarRobotInfo.Host || robot.Port != isarRobotInfo.Port)
+            {
+                _logger.LogInformation("Halleluja");
```
GodVenn, line 170: Not sure about this log message, are you in a good place @aestene ?

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -55,33 +60,161 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
+                );
+                _isarConnectionTimers[isarRobotStatus.RobotName].Reset();
+            }
             else
             {
-                robot.Host = isarRobot.Host;
-                robot.Port = isarRobot.Port;
-                robot.Enabled = true;
-                await RobotService.Update(robot);
+                var timer = new System.Timers.Timer(10000);
```
GodVenn, line 95: Should probably add this timeout to config instead of magic number in code

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -55,33 +60,161 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
+                robot.Status
+            );
+
+            if (_isarConnectionTimers.ContainsKey(isarRobotStatus.RobotName))
+            {
+                _logger.LogInformation(
+                    "Reset connection timer for robot {name}",
+                    isarRobotStatus.RobotName
+                );
+                _isarConnectionTimers[isarRobotStatus.RobotName].Reset();
```
GodVenn, line 91: Are we certain that "stop" and "start" of timer resets the timer's clock?

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -1,4 +1,5 @@
-﻿using Api.Database.Models;
+﻿using Api.Controllers.Models;
+using Api.Database.Models;
```
GodVenn, line 2:
```suggestion
using Api.Database.Models;
using System.Timers;
```

GodVenn: Add namespace for timers

backend/api/EventHandlers/MqttEventHandler.cs
GodVenn, line 170: Quite some repeating code here, is it possible to make this one method, both updating if needed and alerting us of it in the same method?

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -20,6 +21,8 @@ public class MqttEventHandler : EventHandlerBase
         private IRobotService RobotService =>
             _scopeFactory.CreateScope().ServiceProvider.GetRequiredService<IRobotService>();
 
+        private readonly Dictionary<string, System.Timers.Timer> _isarConnectionTimers = new();
```
GodVenn, line 24: I think this belongs in its own EventHandler, it's not related to MQTT (receiving messages and emitting events) and should instead listen to the MQTT events and act accordingly but as a distinct service.
Maybe an "IsarConnectionMonitor" as its own EventHandler?

GodVenn (commented):
An important chunk of work! 
I have some thoughts and prayers, let me know what you think

backend/api/EventHandlers/MqttEventHandler.cs
aestene: That is indeed on the "to be removed" list somewhere in the back of my head :D Some debugging was required.

backend/api/EventHandlers/MqttEventHandler.cs
aestene: Fairly certain yes. I found it slightly counterintuitive too but some googling told me that was the way to solve it. 

And it works in practise.

backend/api/EventHandlers/MqttEventHandler.cs
aestene: I have been looking quite a bit at this previously and it is the checking of the video stream object that has me confused. I agree of course but haven't found a neater way yet.

backend/api/EventHandlers/MqttEventHandler.cs
aestene: 'Timer' is an ambiguous reference between 'System.Timers.Timer' and 'System.Threading.Timer' [api]csharp(CS0104)

For this reason I have to specfically import in this manner and thus the import statement on the top is not required.

backend/api/EventHandlers/MqttEventHandler.cs
GodVenn: Ok, sounds good

backend/api/EventHandlers/MqttEventHandler.cs
GodVenn: Ah ok, makes sense!

backend/api/EventHandlers/MqttEventHandler.cs
aestene: I am leaving this as it is even though I know it's not the prettiest thing I have produced... Suggestions for how to change it would be welcome.

backend/api/EventHandlers/MqttEventHandler.cs
aestene: Implemented

GodVenn (approved):
LGTM

PR #513, opened 2023-01-19
Save asset value in Mission objects

Eddasol:


frontend/src/api/ApiCaller.tsx
```diff
@@ -128,10 +128,16 @@ export class BackendAPICaller {
         return result.content
     }
     async postMission(echoMissionId: number, robotId: string, startTime: Date) {
+        const assetCode = sessionStorage.getItem('assetString')
```
Eddasol, line 131: Is it OK to get asset like this or should it be given as an argument in postMission instead?

oysand:
There is an AssetCode in backend/api/Database/Models/Mission.cs. Could this be utilized?

oysand (approved):
LGTM

PR #514, opened 2023-01-19
Fix homepage bug in package.json

phhuseboe:


phhuseboe:
Would this solution cause trouble for the Aurora pipeline?

tsundvoll (approved):

PR #549, opened 2023-02-03
Feature add recent mission section

### LINKED ISSUE #545: Add 'Recent Missions' section to front page
A Recent mission section where we display a list of missions completed (both failed and succesful).
The section can be a scrollable box the right of ongoing missions.
The request to get all missions should be paginated in the future but for now we can load all and scroll I think.

The recent mission card can contain less information than the pending missions, and will act more as a link to the missions.
Each mission should contain a symbol for how they ended (failed, aborted, success, etc), their name and time of completion.
They should be sorted by time of completion.

This is added to prioritized in milestone because it is related to #509 

It wont

GodVenn:
Closes #545 

Changed name to 'Past Missions' because there is no guarantee they are recent.

### DIFF
--- a/frontend/src/api/ApiCaller.tsx
+++ b/frontend/src/api/ApiCaller.tsx
@@ -88,6 +88,14 @@ export class BackendAPICaller {
         return result.content
     }
 
+    async getMissions(): Promise<Mission[]> {
+        const path: string = 'missions'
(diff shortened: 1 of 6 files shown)


Eddasol (approved):
LGTM

PR #553, opened 2023-02-06
Map name bug

GodVenn:
Wrong config was merged


Christdej:
I think we can keep this configuration for Development. That is, we let MQTT fail after a few tries, and that is okey.

GodVenn:
> I think we can keep this configuration for Development. That is, we let MQTT fail after a few tries, and that is okey.

Agree. Want to change the "try count" to 5 though, to give MQTT a chance to connect if it's slow.

Christdej (approved):

PR #557, opened 2023-02-06
Remove wrong keyword in deploy workflow

tsundvoll:
`login-server` was from an example and it looks like it is replaced by the `registry` keyword and hence can be removed.

Got this from the logs:
`Warning: Unexpected input(s) 'login-server', valid inputs are ['registry', 'username', 'password', 'ecr', 'logout']`


GodVenn (approved):
LGTM

PR #559, opened 2023-02-07
Feature dismiss error message

### LINKED ISSUE #555: Add acknowledge button on error messages on front page
Add acknowledge button on error message. Error message should disappear when the acknowledge button is pressed. 

Current solution:
image

Front page will get a cleaner look when error messages can be acknowledged and removed.

GodVenn:
Closes #555

### DIFF
--- a/frontend/src/components/Pages/FrontPage/MissionOverview/FailedMissionAlertView.tsx
+++ b/frontend/src/components/Pages/FrontPage/MissionOverview/FailedMissionAlertView.tsx
@@ -1,23 +1,38 @@
-import { Button, Card, Typography } from '@equinor/eds-core-react'
+import { Button, Card, Icon, Typography } from '@equinor/eds-core-react'
 import { tokens } from '@equinor/eds-tokens'
 import { useApi } from 'api/ApiCaller'
-import { differenceInMinutes } from 'date-fns'
 import { Mission, MissionStatus } from 'models/Mission'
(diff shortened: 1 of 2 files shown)


frontend/src/components/Pages/FrontPage/MissionOverview/FailedMissionAlertView.tsx
```diff
@@ -47,17 +59,39 @@ export function FailedMissionAlertView({ refreshInterval }: RefreshProps) {
 
     const apiCaller = useApi()
     const [missionsToDisplay, setMissionsToDisplay] = useState<Mission[]>([])
+    const [dismissedMissions, setDismissedMissions] = useState<Mission[]>([])
+    const [recentFailedMissions, setRecentFailedMissions] = useState<Mission[]>([])
```
Eddasol, line 63: Would it be enough with just failedMissionsToDisplay?

GodVenn:
Will change dismissed missions to be added to session storage for "persistent data per user"

Eddasol (commented):
Suggest changing such that all failed missions since last dismiss is shown and not just last 10 minutes

GodVenn:
> Suggest changing such that all failed missions since last dismiss is shown and not just last 10 minutes

I would argue that's what we have the "Past mission" section for and that this alert is meant as a "right now"-alert?

Eddasol (approved):
LGTM

PR #561, opened 2023-02-08
Rotated stream

Eddasol:


backend/api/Database/Models/VideoStream.cs
```diff
@@ -23,5 +23,8 @@ public class VideoStream
         [MaxLength(64)]
         [Required]
         public string Type { get; set; }
+
+        [MaxLength(64)]
+        public bool ShouldRotate { get; set; }
```
oysand, line 28: Should we specify the degree to rotate here?

oysand: Should it be required?

Eddasol: We could, but currently we only support 270 deg rotation. Would renaming to ShouldRotate270 be better?

Eddasol: I don't think it should be required as this is only relevant for some robots

frontend/src/components/Pages/MissionPage/VideoStream/VideoStreamWindow.tsx
```diff
@@ -30,33 +58,97 @@ interface VideoStreamWindowProps {
...
     }
 
     return (
         <VideoCard variant="default" style={{ boxShadow: tokens.elevation.raised }}>
-            <StyledVideoSection>{videoPlayer}</StyledVideoSection>
+            <div onDoubleClick={turnOnFullScreen}>{videoPlayer}</div>
             <Typography variant="h5">{videoStream.name}</Typography>
         </VideoCard>
     )
 }
```
Eddasol, line 114: Split the file here

frontend/src/components/Pages/MissionPage/VideoStream/VideoStreamWindow.tsx
```diff
@@ -30,33 +58,97 @@ interface VideoStreamWindowProps {
...
+        <FullscreenButton color="secondary" onClick={turnOnFullScreen}>
+            <Icon name="fullscreen" size={32} />
+        </FullscreenButton>
+    )
+
     if (IsValidOvenPlayerType({ videoStream })) {
-        videoPlayer = <VideoPlayerOvenPlayer videoStream={videoStream} />
+        if (videoStream.shouldRotate) {
+            videoPlayer = (
+                <StyledVideoSection style={{ width: '10rem' }}>
```
oysand, line 82: Create a new styled componenent from StyledVideoSection

backend/api/Database/Models/VideoStream.cs
oysand: Agreed, should also state CW or CCW

GodVenn (approved):
LGTM

PR #573, opened 2023-02-15
Present duration estimate in frontend

Eddasol:


backend/api/Database/Models/Mission.cs
```diff
@@ -58,7 +58,7 @@ public MissionStatus MissionStatus
 
         public DateTimeOffset EndTime { get; set; }
 
-        public TimeSpan EstimatedDuration { get; set; }
+        public int EstimatedDuration { get; set; }
```
Christdej, line 61: We dont want to change this to int i think, if you need the hours or minutes see:
https://learn.microsoft.com/en-us/dotnet/api/system.timespan?view=net-7.0

frontend/src/components/Pages/MissionPage/MissionHeader/MissionHeader.tsx
```diff
@@ -27,18 +27,33 @@ interface MissionHeaderProps {
...
+        remainingTime = estimatedDuration + ' minutes'
     } else {
-        remainingTime = 'Not available'
+        if (mission.endTime && mission.startTime) {
+            usedTime = differenceInMinutes(new Date(mission.endTime), new Date(mission.startTime))
+            usedTime = usedTime > 0 ? usedTime : 0
+        } else {
+            usedTime = 0
+        }
+        remainingTime = 'N/A'
```
MortFred, line 56: Maybe this can be put into another function, and then called from MissionHeader?
Personally a fan describing the components present in the misison header in this function, and then leave the nitty-gritty details to sub-functions.

backend/api/Database/Models/Mission.cs
```diff
@@ -103,6 +104,27 @@ public static MissionStatus MissionStatusFromString(string status)
...
+            foreach (var plannedTask in PlannedTasks)
+            {
+                numberOfTags += plannedTask.Inspections.Count;
+                var currentPosition = plannedTask.Pose.Position;
+                distance += Math.Abs(currentPosition.X - prevPosition.X) + Math.Abs(currentPosition.Y - prevPosition.Y);
+                prevPosition = currentPosition;
+            }
+            int estimate = (int)(Divide((decimal)distance, robotVelocity * efficiencyFactor) + (numberOfTags * inspectionTime)); // min
+            EstimatedDuration = TimeSpan.FromMinutes(estimate);
+        }
```
MortFred, line 127: Make a test for this function?

I think perhaps all the 'magic numbers' used in this function can either be moved to a config file, or should in the future come from the robot type.
That way we can make a test for the function without having to keep ensuring that the test and this function has the same config values.

MortFred (changes requested):

backend/api/Database/Models/Mission.cs
```diff
@@ -103,6 +104,27 @@ public static MissionStatus MissionStatusFromString(string status)
                   )
             };
         }
+
+        public void CalculateEstimatedDuration()
+        {
+            decimal robotVelocity = (decimal)1.5 * 1000 / 60; // km/t => m/min
+            decimal efficiencyFactor = (decimal)0.20;
+            decimal inspectionTime = 2; // min/tag
```
GodVenn, line 112: I think doubles are plenty good enough here, no need to introduce another data type?

backend/api/Database/Models/Mission.cs
```diff
@@ -103,6 +104,27 @@ public static MissionStatus MissionStatusFromString(string status)
...
+            int assumedXYMetersFromFirst = 20;
+
+            double distance = 0;
+            int numberOfTags = 0;
+            var prevPosition = new Position(PlannedTasks.First().Pose.Position.X + assumedXYMetersFromFirst, PlannedTasks.First().Pose.Position.Y + assumedXYMetersFromFirst, PlannedTasks.First().Pose.Position.Z);
+            foreach (var plannedTask in PlannedTasks)
+            {
+                numberOfTags += plannedTask.Inspections.Count;
+                var currentPosition = plannedTask.Pose.Position;
+                distance += Math.Abs(currentPosition.X - prevPosition.X) + Math.Abs(currentPosition.Y - prevPosition.Y);
```
GodVenn, line 122: Should we use vector distance instead of euclidean distance here?

backend/api/Database/Models/Mission.cs
GodVenn: Not sure it makes sense to test this function, since its an estimate, we can't really test the functionality other than the exact results it gives us now?

backend/api/Database/Models/Mission.cs
Eddasol: I got overflow when dividing using "/" but changing to the function decimal.Devide solved the problem

backend/api/Database/Models/Mission.cs
Eddasol: It's rarely possible to drive diagonal so I think vector distance will be more representative for the driving pattern

GodVenn: Good point

backend/api/Database/Models/Mission.cs
Eddasol: Leaves the constants as is, since this is a temporary solution for the constants and they will be replaced with robot specific constants at a later time.

GodVenn (approved):
LGTM

frontend/src/components/Pages/MissionPage/MissionHeader/MissionHeader.tsx
Eddasol: Agree

MortFred (approved):
Good work!

PR #585, opened 2023-02-15
Add workflow verifying migrations if needed

GodVenn:
With this, you cannot merge if migrations are added and you haven't updated database


GodVenn:
This wont work because we need the workflow to always run, even if there is nothing related to databases in the PR (This is because we want it to be required)

GodVenn:
> This wont work because we need the workflow to always run, even if there is nothing related to databases in the PR (This is because we want it to be required)

Fixed by adding another workflow

Christdej (approved):

PR #586, opened 2023-02-15
Fix frontend base route

tsundvoll:
Makes the base route configurable. For running locally, set the variable `REACT_APP_FRONTEND_BASE_ROUTE` to `""` for convenience. For deploying the frontend it can be configured as needed.


Eddasol (approved):
LGTM

PR #588, opened 2023-02-16
Test that predefined poses are converted equally to and from the echo database

eivindsjovold:


backend/api.test/Database/Models.cs
```diff
@@ -10,12 +10,63 @@ public class TestPose
...
+        }
+
+        private EchoPose ConvertPredefinedPoseToEchoPose(Position position, Orientation orientation, string frame)
+        {
+            var enuPosition = new EchoVector(position.X, position.Y, position.Z);
+            var axisAngle = ConvertOrientation(orientation);
+            return new EchoPose(enuPosition, axisAngle, "robot");
+        }
+        private static AxisAngle ConvertOrientation(Orientation orientation)
+        // This is the method used to convert predefined poses to the Angle-Axis representation used by
```
GodVenn, line 39: used by who?

backend/api.test/Database/Models.cs
```diff
@@ -10,12 +10,63 @@ public class TestPose
...
+        {
+            public EchoVector Position;
+            public AxisAngle Orientation;
+            public string Frame;
+
+            public EchoPose(EchoVector position, AxisAngle orientation, string frame)
+            {
+                Position = position;
+                Orientation = orientation;
+                Frame = frame;
```
GodVenn, line 68: Doesn't this already exist?

eivindsjovold: Not to my knowledge, but it should probably

GodVenn: Checked, it doesn't, it's different for our other purposes so this is all good

GodVenn (approved):
LGTM

PR #626, opened 2023-02-25
Always download map from blob

oysand:


backend/api/Controllers/MissionController.cs
```diff
@@ -91,9 +92,13 @@ public async Task<ActionResult<byte[]>> GetMap([FromRoute] string id)
     {
         try
         {
-            string filePath = await _mapService.FetchMapImage(id);
-            var returnFile = PhysicalFile(filePath, "image/png");
-            return returnFile;
+            Image mapImage = await _mapService.FetchMapImage(id);
```
anetteu, line 95: Use `var` instead of explicit type

```suggestion
            var mapImage = await _mapService.FetchMapImage(id);
```

backend/api/Controllers/MissionController.cs
```diff
@@ -91,9 +92,13 @@ public async Task<ActionResult<byte[]>> GetMap([FromRoute] string id)
...
+
+            using (var ms = new MemoryStream())
+            {
+                mapImage.Save(ms, mapImage.RawFormat);
+                return File(ms.ToArray(), "image/png");
```
anetteu, line 100: Avoid abbreviations in var names:

```suggestion
            using (var memoryStream = new MemoryStream())
            {
                mapImage.Save(memoryStream, mapImage.RawFormat);
                return File(memoryStream.ToArray(), "image/png");
```

anetteu (approved):

PR #632, opened 2023-02-27
Use dynamic installation code rather than config

### LINKED ISSUE #627: Missions from JS does not schedule
**Describe the bug**
Missions from JS does not start

**To Reproduce**
When scheduling echo mission 387 in flotilla, the mission does not start.
This was tested both locally and in the aurora dev environment.

**Expected behavior**
Missions from JS should be able to be scheduled.

**Screenshots**

aestene:
It is no longer required to configure installations code in the backend. This is provided through the API from the frontend as the use selects which installation they are on. 

Closes #627

### DIFF
--- a/backend/api/Controllers/MissionController.cs
+++ b/backend/api/Controllers/MissionController.cs
@@ -164,7 +164,7 @@ public class MissionController : ControllerBase
             .Select(
                 t =>
                 {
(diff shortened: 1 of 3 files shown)


aestene:
image

Christdej (approved):
LGTM

PR #639, opened 2023-03-02
Remove option to choose mission start time

Eddasol:


backend/api/Controllers/Models/ScheduledMissionQuery.cs
```diff
@@ -4,7 +4,6 @@ public struct ScheduledMissionQuery
     {
         public string RobotId { get; set; }
         public int EchoMissionId { get; set; }
-        public DateTimeOffset StartTime { get; set; }
```
GodVenn, line 7: Why are we removing this option? 
If we do, we should rename from 'Schedule' to 'Queue'

GodVenn: Even if we want to remove the option for frontend users, we don't have to remove it as an option in the backend

Eddasol: I reintroduced startTime but set it as optional

GodVenn:
In order to maintain the functionality in the backend, we need to keep the start time as required and use it in the scheduler.
To remove the functionality from the frontend we can just always set start_time=now in the api call.

backend/api/Database/Models/Mission.cs
```diff
@@ -53,10 +53,9 @@ public MissionStatus MissionStatus
         [Required]
         public MissionMap Map { get; set; }
 
-        [Required]
-        public DateTimeOffset StartTime { get; set; }
+        public DateTimeOffset? StartTime { get; set; }
```
GodVenn, line 56:
```suggestion
        [Required]
        public DateTimeOffset StartTime { get; set; }
```

frontend/src/api/ApiCaller.tsx
```diff
@@ -135,14 +135,13 @@ export class BackendAPICaller {
...
     }
-    async postMission(echoMissionId: number, robotId: string, startTime: Date, assetCode: string | null) {
+    async postMission(echoMissionId: number, robotId: string, assetCode: string | null) {
         const path: string = 'missions'
         const robots: Robot[] = await this.getRobots()
         const desiredRobot = filterRobots(robots, robotId)
         const body = {
             robotId: desiredRobot[0].id,
             echoMissionId: echoMissionId,
-            startTime: startTime,
```
GodVenn, line 145: startTime: Date.Now()

GodVenn (changes requested):

backend/api/Database/Context/InitDb.cs
```diff
@@ -87,7 +87,7 @@ private static List<Mission> GetMissions()
             EchoMissionId = 95,
             IsarMissionId = "1",
             MissionStatus = MissionStatus.Pending,
-            StartTime = DateTimeOffset.UtcNow.AddHours(3),
+            StartTime = DateTimeOffset.UtcNow,
```
aestene, line 90: Having these three missions available like this might be annoying. Every time we start Flotilla locally it will immediately try to schedule a mission. 

Are they used for testing in any way? We can also have missions that are not `Pending` but marked as successful etc. to show up in the other tables. For `Pending` maybe having none is better but check if there are tests relying on them.

backend/api/EventHandlers/MissionScheduler.cs
```diff
@@ -41,7 +41,6 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
                     }
                     if (
                         freshMission.Robot.Status is not RobotStatus.Available
-                        || freshMission.StartTime > DateTimeOffset.UtcNow
```
aestene, line 44: Have you tested if the order of missions is maintained?

So if you schedule multiple missions (three or more) and we schedule them one at a time again and again and again in the same order. Do we know that the missions will be executed in the correct order? Now the UpcomingMissions object is collected from the database but I'm not sure if the order is guaranteed to be the same.

aestene (changes requested):
Have a look at my two comments.

backend/api/EventHandlers/MissionScheduler.cs
GodVenn: I think we should keep this line to keep scheduling functionality in backend

backend/api/Database/Models/Mission.cs
GodVenn: Maybe we should rename this to `ScheduledStartTime`, and add another field `StartTime` that represents the actual time the mission started.

backend/api/EventHandlers/MissionScheduler.cs
Eddasol: The order of the missions is not maintained as is either

frontend/src/components/Pages/MissionPage/MissionHeader/MissionHeader.tsx
```diff
@@ -60,32 +60,31 @@ function HeaderText(text: string) {
...
-    let usedTime = 0
+    var usedTime
 
     let dateTime = mission.estimatedDuration.split('.')
     const days = dateTime.length === 1 ? 0 : dateTime[0].split(':')[0]
     const time = dateTime.length === 1 ? dateTime[0].split(':') : dateTime[1].split(':')
     const estimatedDuration = +time[1] + 60 * (+time[0] + +days * 24)
 
-    if (mission.missionStatus === MissionStatus.Ongoing || mission.missionStatus === MissionStatus.Paused) {
+    if (mission.endTime) {
```
GodVenn, line 73: Have you tested that "if (mission.endTime)" behaves as expected without endTime?
I sometimes experienced that endTime is not null but instead some random time in the frontend

frontend/src/components/Pages/MissionPage/MissionHeader/MissionHeader.tsx
```diff
@@ -60,32 +60,31 @@ function HeaderText(text: string) {
...
+        startTime = mission.startTime
+            ? format(new Date(mission.startTime), 'HH:mm')
+            : format(new Date(mission.endTime), 'HH:mm')
+        usedTime = mission.startTime ? differenceInMinutes(new Date(mission.endTime), new Date(mission.startTime)) : 0
+        remainingTime = 'N/A'
+    } else if (mission.startTime) {
```
GodVenn, line 79: Same question as for endTime

backend/api/Database/Context/InitDb.cs
Eddasol: There are tests relying on them. Suggest that modification of these placeholder missions get made into a separate issue

backend/api/Database/Context/InitDb.cs
```diff
@@ -87,7 +87,7 @@ private static List<Mission> GetMissions()
             EchoMissionId = 95,
             IsarMissionId = "1",
             MissionStatus = MissionStatus.Pending,
-            DesiredStartTime = DateTimeOffset.UtcNow.AddHours(3),
+            DesiredStartTime = DateTimeOffset.UtcNow,
```
GodVenn, line 90: Like @aestene mentioned, maybe we want to keep these ass scheduled for a couple of hours after startup?
We used to have them start right away and it was sometimes hard to test new things if they started and crashed right away

frontend/src/components/Pages/MissionPage/MissionHeader/MissionHeader.tsx
Eddasol: It didn't use to work but with the change in Mission.cs allowing endTime to be either DateTimeOffset or null this works

GodVenn: Nice

backend/api/Database/Context/InitDb.cs
Eddasol: The missions have been changed to successful such that they don't start

backend/api/Database/Context/InitDb.cs
Eddasol: Status of these missions has been changed to successful to prevent them from starting

GodVenn (approved):
LGTM, maybe @aestene wants to approve as well

aestene (approved):
LGTM

PR #644, opened 2023-03-03
Handle KeyNotFoundException for map metadata

aestene:


backend/api/Services/MapService.cs
```diff
@@ -173,14 +173,14 @@ private int[] ExtractImageSize(BlobItem map)
                 int y = int.Parse(map.Metadata["imageHeight"], CultureInfo.CurrentCulture);
                 return new int[] { x, y };
             }
-            catch (FormatException e)
+            catch (Exception e) when (e is FormatException || e is KeyNotFoundException)
```
GodVenn, line 176: We should probably catch all errors here to log the warning and reraise
```suggestion
            catch (Exception e)
```

backend/api/Services/MapService.cs
```diff
@@ -154,14 +154,14 @@ private Boundary ExtractMapMetadata(BlobItem map)
                     double.Parse(map.Metadata["upperRightY"], CultureInfo.CurrentCulture) / 1000;
                 return new Boundary(lowerLeftX, lowerLeftY, upperRightX, upperRightY);
             }
-            catch (FormatException e)
+            catch (Exception e) when (e is FormatException || e is KeyNotFoundException)
```
GodVenn, line 157: We should probably catch all errors here to log the warning and reraise
```suggestion
            catch (Exception e)
```

backend/api/Services/MapService.cs
```diff
@@ -77,7 +77,7 @@ await foreach (var blobPage in resultSegment)
                             boundaries.Add(blobItem.Name, ExtractMapMetadata(blobItem));
                             imageSizes.Add(blobItem.Name, ExtractImageSize(blobItem));
                         }
-                        catch (FormatException)
+                        catch (Exception e) when (e is FormatException || e is KeyNotFoundException)
```
GodVenn, line 80: Why do we just ignore and continue in the case of these exceptions? Should we not even log a warning?

GodVenn (changes requested):

PR #653, opened 2023-03-08
Only fetch echo missions when scheduling missions

Eddasol:


frontend/src/components/Contexts/AssetContext.tsx
```diff
@@ -10,16 +10,18 @@ interface Props {
...
-    asset: 'test',
+    asset: '',
     switchAsset: (newAsset: string) => {},
 }
 
 export const AssetContext = createContext<IAssetContext>(defaultAsset)
 
 export const AssetProvider: FC<Props> = ({ children }) => {
-    const [asset, setAsset] = useState(defaultAsset.asset)
+    const prevAsset = sessionStorage.getItem('assetString')
```
GodVenn, line 20: We could consider using localStorage instead of session here, that way we store the last Asset for the entire browser, even if its closed. I imagine this is more common for users of flotilla, as one use/browser/computer will most likely only be used for one plant

frontend/src/components/Pages/FrontPage/MissionOverview/MissionQueueView.tsx
```diff
@@ -52,8 +53,19 @@ export function MissionQueueView({ refreshInterval }: RefreshProps) {
     const [robotOptions, setRobotOptions] = useState<Map<string, Robot>>(new Map<string, Robot>())
     const [scheduleButtonDisabled, setScheduleButtonDisabled] = useState<boolean>(true)
     const [frontPageScheduleButtonDisabled, setFrontPageScheduleButtonDisabled] = useState<boolean>(true)
+    const [loadingEchoMissions, setLoadingEchoMissions] = useState<boolean>(false)
```
GodVenn, line 56:
```suggestion
    const [isLoadingEchoMissions, setIsLoadingEchoMissions] = useState<boolean>(false)
```

frontend/src/components/Pages/FrontPage/MissionOverview/MissionQueueView.tsx
```diff
@@ -52,8 +53,19 @@ export function MissionQueueView({ refreshInterval }: RefreshProps) {
...
+    const { asset } = useAssetContext()
     const echoURL = 'https://echo.equinor.com/mp?instCode='
-    const savedAsset = sessionStorage.getItem('assetString')
+
+    const onFrontPageScheduleButtenPress = () => {
```
GodVenn, line 60:
```suggestion
    const onFrontPageScheduleButtonPress = () => {
```
Typo in button

frontend/src/components/Contexts/AssetContext.tsx
Eddasol: Agree

PR #656, opened 2023-03-09
Provide user with loading info

Eddasol:


Eddasol:
image
Added text in the loading window. This is the design in this PR

GodVenn (approved):
LGTM, disabled the check for this PR, will  look into why it fails if it happens again. Seems like a weird bug

PR #664, opened 2023-03-10
Common Task and Inspection

### LINKED ISSUE #356: Differ more clearly between ISAR Task and Echo Tag / Task
There is a bit of an ambiguity with the IsarTask concept and the EchoTag / (EchoTask) concept. Consider making this clearer.

```
export interface Mission {
    ... 
    tasks: IsarTask[]
    plannedTasks: EchoTag[]
}

``` 

This would probably require a change in the backend api as well.

None

### LINKED ISSUE #406: Add link to EchoMission in tasktable in front end
The tasktable for ongoing missions has an empty field for link to EchoMission. This should be populated.

image

### LINKED ISSUE #530: Task should be arranged from the first to the last task of the mission
At the moment, the tasks in the tasks list of an ongoing mission are in random order. Should be arranged from the first to the last, with task number 1 as the first task and following. See screenshot of the task list from the ongoing mission page, where task number 3 is the first task of the mission. No clear identication wether task 1 or task 2 is the next task of the mission. 

MicrosoftTeams-image (23)

No change in functionality, but easier to follow ongoing missions in flotilla. 

No change.

### LINKED ISSUE #635: Sort tasks from first to last utilizing the order index in IsarTask
Sort tasks from first to last utilizing the order index in mission's IsarTask

The tasks in each mission will be shown chronologically in task table of a mission

N/A

### LINKED ISSUE #638: Add start and end time to mission, task and step objects
The mission, task and step database objects should record when they where started (by ISAR) and when they were completed (by ISAR).

Suggested solution is to set start time and end time values when the mission, step and task is created in the database to null. Then the events for mission updates, task updates and step updates may be used to set start time and end time for each individual object.

When the event `OnStepUpdate` occurs the `Step.Status` should be checked. If it is `StepStatus.InProgress` then start time may be set to `Now`. If the status is a finished status, for example `StepStatus.Succesful` the end time may be set for that step.

The same may be applied to mission and task for the functions `OnMissionUpdate` and `OnTaskUpdate` respectively. 

This will enable us to compute duration of each step, task and mission and may be used in the future to estimate the duration of missions for the robot. 

No change.

### LINKED ISSUE #661: Merge ISAR Task and Planned Task in Backend
Just one common task containing the information we currently have in two data structures `PlannedTask` and `IsarTask`.
This implies also merging the sub data structures `PlannedInspection` and `IsarStep`.

This will make our data structure overall more intuitive.

N/A

GodVenn:
Closes #661 
Closes #638 
Closes #635 
Closes #356 
Closes #530 
Closes #406


GodVenn:
Will add migrations when PR approved

backend/api/Services/MapService.cs
```diff
@@ -136,7 +121,7 @@ private async Task<Image> DownloadMapImageFromBlobStorage(Mission currentMission
             );
             var blobClient = blobContainerClient.GetBlobClient(currentMission.Map.MapName);
 
-            using var stream = await blobClient.OpenReadAsync();
+            await using var stream = await blobClient.OpenReadAsync();
```
aestene, line 124: I'm not familiar with using await in front here. Is this correct?

GodVenn: Resharper suggested it

GodVenn: https://stackoverflow.com/questions/58791938/c-sharp-8-understanding-await-using-syntax
"Similar as **using (...)** uses `IDisposable` to clean up resources, **await using (...)** uses `[IAsyncDisposable]`(https://learn.microsoft.com/en-us/dotnet/api/system.iasyncdisposable?view=dotnet-plat-ext-3.0). This allows to perform also time-consuming tasks (e.g involving I/O) on cleanup without blocking."

backend/api/Controllers/MissionController.cs
```diff
@@ -160,32 +166,33 @@ public async Task<ActionResult<byte[]>> GetMap([FromRoute] string id)
             return StatusCode(StatusCodes.Status502BadGateway, message);
         }
 
-        var plannedTasks = echoMission.Tags
+        var tasks = echoMission.Tags
```
oysand, line 169: If tasks is a list of MissionTasks, should we rename it to mission tasks?

backend/api/Controllers/RobotController.cs
```diff
@@ -369,9 +365,7 @@ public async Task<ActionResult<IList<VideoStream>>> GetVideoStreams([FromRoute]
             return StatusCode(StatusCodes.Status500InternalServerError, message);
         }
 
-        mission.IsarMissionId = isarServiceStartMissionResponse.IsarMissionId;
-        mission.StartTime = isarServiceStartMissionResponse.StartTime;
-        mission.Tasks = isarServiceStartMissionResponse.Tasks;
+        mission.UpdateWithIsarInfo(isarMission);
```
oysand, line 368: Would it be a better name with UpdateIsarInfo?

backend/api/Database/Models/MissionTask.cs
```diff
@@ -0,0 +1,136 @@
...
+using Api.Controllers.Models;
+using Api.Services.Models;
+using Microsoft.EntityFrameworkCore;
+
+#nullable disable
+
+namespace Api.Database.Models
+{
+    [Owned]
+    public class MissionTask
```
oysand, line 12: Could this be named only Task?

backend/api/Database/Models/MissionTask.cs
```diff
@@ -0,0 +1,136 @@
...
+                        step.IsarStepId.Equals(
+                            inspection.IsarStepId.ToString(),
+                            StringComparison.Ordinal
+                        )
+                );
+                inspection.UpdateWithIsarInfo(correspondingStep);
+            }
+        }
+
+        public void UpdateTaskStatus(IsarTaskStatus isarStatus)
```
oysand, line 100: Could this be just UpdateStatus at it could only be the Status for the Task?

backend/api/Database/Models/MissionTask.cs
```diff
@@ -0,0 +1,136 @@
...
+        {
+            return Inspections.FirstOrDefault(
+                inspection => inspection.IsarStepId.Equals(isarStepId, StringComparison.Ordinal)
+            );
+        }
+
+#nullable disable
+    }
+
+    public enum TaskStatus
```
oysand, line 126: Is this used outside the task or could it make sense to call it Status?

backend/api/Services/IsarService.cs
```diff
@@ -71,22 +64,14 @@ public IsarService(ILogger<IsarService> logger, IDownstreamWebApi downstreamWebA
...
-            return new IsarMissionDefinition(tasks: tasks.ToList());
-        }
-
-        public async Task<IsarServiceStartMissionResponse> StartMission(
-            Robot robot,
-            IsarMissionDefinition missionDefinition
-        )
+        public async Task<IsarMission> StartMission(Robot robot, Mission mission)
         {
+            var missionDefinitionObject = new IsarMissionDefinition(mission);
```
oysand, line 69: Would it be better to just do this directly at line 74?

backend/api/Database/Models/MissionTask.cs
oysand: Nvm, just saw the IsarTask

backend/api/Database/Models/MissionTask.cs
```diff
@@ -0,0 +1,136 @@
...
+
+        [Required]
+        public Pose RobotPose { get; set; }
+
+        public int EchoPoseId { get; set; }
+
+        private TaskStatus _taskStatus;
+
+        [Required]
+        public TaskStatus TaskStatus
```
oysand, line 42: Status of type TaskStatus

backend/api/Controllers/MissionController.cs
```diff
@@ -160,32 +166,33 @@ public async Task<ActionResult<byte[]>> GetMap([FromRoute] string id)
...
         var scheduledMission = new Mission
         {
             Name = echoMission.Name,
             Robot = robot,
             EchoMissionId = scheduledMissionQuery.EchoMissionId,
             MissionStatus = MissionStatus.Pending,
             DesiredStartTime = scheduledMissionQuery.DesiredStartTime,
-            PlannedTasks = plannedTasks,
-            Tasks = new List<IsarTask>(),
+            Tasks = tasks,
```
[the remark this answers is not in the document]
GodVenn: Fixed

backend/api/Database/Models/MissionTask.cs
GodVenn: Fixed

backend/api/Database/Models/MissionTask.cs
GodVenn: Fixed

backend/api/Services/IsarService.cs
GodVenn: Fixed

backend/api/Database/Models/MissionTask.cs
GodVenn: Agreed to keep enum name and change variable name, since the enum is available from outside the class

frontend/src/models/Pose.ts
```diff
@@ -1,19 +1,10 @@
...
-    z: number
-    w: number
-}
+import { Orientation } from './Orientation'
+import { Position } from './Position'
 
 export interface Pose {
     position: Position
     orientation: Orientation
+    frame?: string
```
oysand, line 7: Do we need the frame or could we just remove it? Could it be anything other than asset?

backend/api.test/EventHandlers/TestMissionScheduler.cs
```diff
@@ -122,7 +122,7 @@ MissionStatus postStatus
             // Assert start conditions
             var preMission = await _missionService.ReadById(ScheduledMission.Id);
             Assert.NotNull(preMission);
-            Assert.True(preMission!.MissionStatus == preStatus);
+            Assert.True(preMission!.Status == preStatus);
```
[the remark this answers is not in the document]
GodVenn: It means that i promise the mission is not null (because it has been asserted at line 124)

frontend/src/models/Pose.ts
oysand: I'd argue that we should remove it both in the frontend and backend unless we want Flotilla to use it actively. We can argue that all Poses in Flotilla is by default asset as we cannot handle other than asset

GodVenn: True

oysand (approved):
LGTM, but remember to fix the tests and run one last time before merging

GodVenn:
Need to add Migrations

PR #671, opened 2023-03-17
Add IsarId to local database

GodVenn:
The ISAR Id is required so we need to add it in the local test database


phhuseboe (approved):
LGTM

PR #695, opened 2023-03-21
Improvement - Show status reasons

### LINKED ISSUE #401: Display error information when the backend fails to schedule a mission
Currently there is no user feedback in the frontend whenever the backend fails to schedule a mission to ISAR. This should be displayed to the user.

Improved user friendliness.

Not affected.

GodVenn:
Closes #401 

Shows status reasons in frontend.
We still need to send status reasons through MQTT to cover other cases than failing to start.

Example of status showing in mission page:

image

### DIFF
--- a/backend/api/EventHandlers/MissionScheduler.cs
+++ b/backend/api/EventHandlers/MissionScheduler.cs
@@ -47,16 +47,21 @@ namespace Api.EventHandlers
                         continue;
                     }
 
-                    bool startedSuccessfull = await StartMission(queuedMission);
-                    if (!startedSuccessfull)
+                    try
                     {
-                        var newStatus = MissionStatus.Failed;
+                        await StartMission(queuedMission);
+                    }
+                    catch (MissionException e)
+                    {
+                        const MissionStatus NewStatus = MissionStatus.Failed;
                         _logger.LogWarning(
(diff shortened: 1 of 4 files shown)


backend/api/EventHandlers/MissionScheduler.cs
```diff
@@ -47,45 +47,50 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
+                    throw new MissionException(errorMessage);
                 }
-                _logger.LogInformation("Started mission '{id}'", scheduledMission.Id);
+                _logger.LogInformation("Started mission '{id}'", queuedMission.Id);
             }
             catch (MissionException e)
             {
-                _logger.LogError(e, "Failed to start mission '{id}'", scheduledMission.Id);
-                return false;
+                _logger.LogError(e, "Failed to start mission '{id}'", queuedMission.Id);
```
MortFred, line 91: We log this excption twice I think, once here and once again when it is caught on line 57?

GodVenn: True, will remove one

GodVenn: Fixed

tsundvoll (approved):
LGTM :+1:

PR #717, opened 2023-03-29
Feature ordering missions api

GodVenn:
This is important for pagination to make sense, you can't order on the client side if the response is paginated 
You can order the 10 items to be ordered by end time for example, but that page might still not hold the most recent end times
You need to be able to order on server side for this, and that's what this pull request does

### DIFF
--- a/backend/api/Database/Context/FlotillaDbContext.cs
+++ b/backend/api/Database/Context/FlotillaDbContext.cs
@@ -1,5 +1,7 @@
 ﻿using Api.Database.Models;
 using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
+using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
 
 namespace Api.Database.Context;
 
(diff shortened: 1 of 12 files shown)


GodVenn:
> 🔔 Changes in database folder detected 🔔 Do these changes require **adding new migrations**? 🤔 In that case follow [these steps](https://github.com/equinor/flotilla/tree/main/backend#Database-model-and-EF-Core). If you are uncertain, ask a database admin on the team 😄

Not needed

backend/api/Services/MissionService.cs
```diff
@@ -346,5 +354,59 @@ MissionQueryStringParameters parameters
...
+                if (objectProperty == null)
+                    throw new InvalidDataException(
+                        $"Mission has no property '{propertyFromQueryName}' for ordering"
+                    );
+                string sortingOrder = param.EndsWith(" desc", StringComparison.OrdinalIgnoreCase)
+                  ? "descending"
+                  : "ascending";
+#pragma warning disable CA1305
+                orderQueryBuilder.Append($"{objectProperty.Name} {sortingOrder}, ");
+#pragma warning restore CA1305
```
tsundvoll, line 399: Why is this needed?

backend/api/Services/MissionService.cs
```diff
@@ -346,5 +354,59 @@ MissionQueryStringParameters parameters
...
+            }
+
+            string orderQuery = orderQueryBuilder.ToString().TrimEnd(',', ' ');
+            if (string.IsNullOrWhiteSpace(orderQuery))
+            {
+                missions = missions.OrderBy(x => x.Name);
```
tsundvoll, line 405: Can we choose a more descriptive name than `x`?

backend/api/Database/Context/FlotillaDbContext.cs
```diff
@@ -12,25 +14,88 @@ public class FlotillaDbContext : DbContext
...
+                            pose =>
+                            {
+                                pose.OwnsOne(p => p.Position);
+                                pose.OwnsOne(p => p.Orientation);
+                            }
+                        );
+                    }
+                );
+            }
+        );
```
tsundvoll, line 52: Can we have more descriptive variable names than `m`, `t`, `r` and `p`?

backend/api/Services/MissionService.cs
GodVenn: Yes!

tsundvoll (approved):
Fix comments above, then it looks good to me :+1:

PR #720, opened 2023-03-31
Order mission queue by time in backend

### LINKED ISSUE #660: Order mission queue
Currently the missions get picked from the mission queue in a seemingly random order. It would be better if the mission that got added to the queue first get executed first. 

Bonus: Making sure that the order is presented correctly in frontend as well and allowing the user to modify the order of the missions trough drag-and-drop 

GodVenn:
Closes #660


phhuseboe (approved):
LGTM

Issue #721, opened 2023-03-31
Add robot model class to database


UsamaEquinorAFK:
**Describe the new feature you would like to see**

Include a robot model class in database (instead of an enum). Robot model class defines robot model specific characteristics, add pressure limit and battery limit  as new characteristics.  

**Describe the solution you'd like**
NA

**How will this feature affect the current Threat Model?**
NA

UsamaEquinorAFK:
#727 Added the variables to the Robot class instead of robot model as that is an enum and cant be initialized as a static class.

GodVenn:
After discussion the class will still be added and enum included in class

PR #727, opened 2023-04-04
Add pressure and battery threshold to robot class in DB

UsamaEquinorAFK:


GodVenn (changes requested):
We should make a new class "RobotModel" as described in #721 which holds these parameters along with the current "RobotModel" enum.

UsamaEquinorAFK:
Handled by #762

PR #728, opened 2023-04-04
Temporary fix to ordering of tasks

tsundvoll:
Increment instead of using taskOrder


tsundvoll:
Closing this, as the issue was something else

PR #734, opened 2023-04-12
Add DB changes to release notes

GodVenn:
Useful to know when creating releases that we need to apply migrations to databases in staging and prod if we promote from dev.


aestene (approved):
LGTM

PR #737, opened 2023-04-14
Localization mission

Eddasol:
Dependent on https://github.com/equinor/isar/pull/456

Further improvement recommendations that aren't implemented here:
* Ignore localization tasks in MQTT to prevent warning
* Fetch only assetDecks for current asset in localization page
* Add asset to the localization mission (and decide if the localization missions should be handled/hidden when they show up in ongoing and mission history)
* Get feedback about success/failure of localization to user
* Present localization pose in map
* Only make localization available for robots that support localization 
* Disable localization button if robot is busy

### DIFF
--- a/backend/api/Controllers/RobotController.cs
+++ b/backend/api/Controllers/RobotController.cs
@@ -528,6 +528,89 @@ public class RobotController : ControllerBase
         return NoContent();
     }
 
+    /// <summary>
+    /// Start a localization mission with localization in the pose 'localizationPose' for the robot with id 'robotId'
+    /// </summary>
+    /// <remarks>
+    /// <para> This query starts a localization for a given robot </para>
+    /// </remarks>
+    [HttpPost]
+    [Route("{robotId}/start-localization")]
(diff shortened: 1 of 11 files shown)


backend/api/Controllers/RobotController.cs
```diff
@@ -528,6 +528,92 @@ public async Task<ActionResult> ResumeMission([FromRoute] string robotId)
...
+        {
+            isarMission = await _isarService.StartLocalizationMission(robot, localizationPose);
+        }
+        catch (HttpRequestException e)
+        {
+            string message = $"Could not reach ISAR at {robot.IsarUri}";
+            _logger.LogError(e, "{message}", message);
+            robot.Enabled = false;
+            robot.Status = RobotStatus.Offline;
+            await _robotService.Update(robot);
```
GodVenn, line 590: Replace with call to "OnIsarUnavailable"

GodVenn (approved):
LGTM

PR #738, opened 2023-04-14
Fix token updates in api caller

### LINKED ISSUE #50: Handle refresh tokens for authentication
**Describe the bug**
If the user authenticates in the OpenAPI documentation the authentication token will not be automatically refreshed after 1 hours (token timeout). This should be handled. 

**To Reproduce**
- Authenticate in OpenAPI
- Wait 1 hour
- MAke a request

**Expected behavior**
Instead of failing authentication a refresh token is used to authenticate again and from the users POV there is no change.

**Screenshots**
image

### LINKED ISSUE #678: Token does not refresh when expiring in frontend
**Describe the bug**
When the access token is expired it does not refresh automatically which forces the user to log in again manually. 

**To Reproduce**
1. Open a browser with the Flotilla frontend
2. Wait for the token to expire (this might take 60-90 minutes)
3. Observe behavior in network log and on the site. It may also be seen on the backend log that requests using an expired token is happening.

**Expected behavior**
When the token expires the refresh token should be used to refresh the access token automatically.

**Screenshots**

GodVenn:
Closes #678 
Refresh tokens are handled by the MSAL framework

Our problem was that these were not propagated to the API caller as this was only instantiated at the start of a render.
Now, the API caller is static and being updated using contexts for asset code and access token

### DIFF
--- a/frontend/src/api/ApiCaller.tsx
+++ b/frontend/src/api/ApiCaller.tsx
@@ -9,6 +9,7 @@ import { PaginatedResponse, PaginationHeader, PaginationHeaderName } from 'model
 import { Pose } from 'models/Pose'
 import { AssetDeck } from 'models/AssetDeck'
 import { timeout } from 'utils/timeout'
+import { tokenReverificationInterval } from 'components/Contexts/AuthProvider'
 
 /** Implements the request sent to the backend api. */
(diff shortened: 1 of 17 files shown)


tsundvoll (approved):
LGTM :+1:

PR #741, opened 2023-04-17
Fix env variables in workflows

GodVenn:


GodVenn:
Cannot use env variables for input to reusable workflows:
https://github.com/orgs/community/discussions/26529

PR #752, opened 2023-04-19
Update README with deployments in Aurora

GodVenn:
Add links to aurora deployments and badges for the deployment pipelines.
Note: The badges might currently be bugged (https://github.com/actions/starter-workflows/issues/1525)


tsundvoll (approved):
👍

PR #762, opened 2023-04-21
Add robot model class

### LINKED ISSUE #721: Add robot model class to database

GodVenn:
Closes #721

### DIFF
--- a/backend/api.test/EventHandlers/TestMissionScheduler.cs
+++ b/backend/api.test/EventHandlers/TestMissionScheduler.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
@@ -7,44 +8,25 @@ using Api.Database.Context;
 using Api.Database.Models;
 using Api.EventHandlers;
 using Api.Services;
+using Api.Services.Models;
 using Api.Test.Mocks;
+using Api.Utilities;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Moq;
 using Xunit;
 
+#pragma warning disable CS1998
 namespace Api.Test.EventHandlers
 {
     [Collection("Database collection")]
     public class TestMissionScheduler : IDisposable
     {
-        private static RobotModel TurtlebotModel =>
-            new()
-            {
-                Id = "TestModel",
-                Type = RobotType.Turtlebot,
-                BatteryWarningThreshold = 20f,
-                LowerPressureWarningThreshold = 20f,
-                UpperPressureWarningThreshold = 80f
-            };
-        private static Robot Robot =>
-            new()
-            {
-                Id = "IamTestRobot",
-                Status = RobotStatus.Available,
-                Host = "localhost",
-                Model = TurtlebotModel,
-                Name = "TestosteroneTesty",
-                SerialNumber = "12354",
-                Enabled = true
-            };
         private static Mission ScheduledMission =>
             new()
             {
-                Id = "testMission",
                 Name = "testMission",
                 EchoMissionId = 2,
-                Robot = Robot,
                 Status = MissionStatus.Pending,
                 DesiredStartTime = DateTimeOffset.Now,
                 AssetCode = "TestAsset",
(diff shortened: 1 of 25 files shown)


backend/api/Controllers/Models/CreateRobotModelQuery.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        /// <summary>
+        /// The type of robot model
+        /// </summary>
+        public RobotType RobotType { get; set; }
+
+        /// <summary>
+        /// Lower battery warning threshold in percentage
+        /// </summary>
+        public float BatteryWarningThreshold { get; set; }
```
Christdej, line 15: Should this be optional aswell?

frontend/src/components/Contexts/AuthProvider.tsx
```diff
@@ -28,7 +28,5 @@ export const AuthProvider = (props: Props) => {
         return () => clearInterval(id)
     }, [])
 
-    VerifyToken()
```
Christdej, line 31: Why is this removed?

backend/api.test/EventHandlers/TestMissionScheduler.cs
```diff
@@ -103,20 +100,30 @@ public void Dispose()
...
 
             var cts = new CancellationTokenSource();
 
             // Add Scheduled mission
-            await _missionService.Create(ScheduledMission);
+            await _missionService.Create(mission);
+
+            _robotControllerMock.RobotServiceMock
+                .Setup(service => service.ReadById(mission.Robot.Id))
+                .Returns(async () => mission.Robot);
```
Christdej, line 119: Alot of methods missing await, might wanna ignore warning or add await

backend/api/Controllers/RobotModelController.cs
```diff
@@ -0,0 +1,237 @@
...
+    [ProducesResponseType(typeof(RobotModel), StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    public async Task<ActionResult<Robot>> UpdateRobotModelById(
+        [FromRoute] string id,
+        [FromBody] UpdateRobotModelQuery robotModelQuery
+    )
```
Christdej, line 149: Would it be better if this was a patch endpoint?

backend/api/Database/Models/RobotModel.cs
```diff
@@ -10,4 +22,52 @@ public enum RobotModel
...
+        [Required]
+        //Save robot type as string in database
+        [Column(TypeName = "nvarchar(56)")]
+        public RobotType Type { get; set; }
+
+        /// <summary>
+        /// Lower battery warning threshold in percentage
+        /// </summary>
+        [Required]
+        public float BatteryWarningThreshold { get; set; }
```
Christdej, line 44: Do we want this to be required?

Christdej (changes requested):

frontend/src/components/Contexts/AuthProvider.tsx
GodVenn: It creates an error, shouldn't call callback functions in component definition

backend/api/Controllers/RobotModelController.cs
GodVenn: A patch would just updating one field, but here we are updating the entire model

backend/api.test/EventHandlers/TestMissionScheduler.cs
GodVenn: Fixed

backend/api/Database/Models/RobotModel.cs
GodVenn: I think you might have a point, it does not need to be required

backend/api/Controllers/Models/CreateRobotModelQuery.cs
GodVenn: Yes, changing to optional

Christdej (approved):

PR #763, opened 2023-04-21
Add support for scheduling custom json missions

### LINKED ISSUE #760: Fetch mission from other sources than Echo
In some cases the thing we want to look at in a mission is not related to a tag and hence, cannot be fetched from Echo, so we need functionality to create missions manually and load them into Flotilla.

GodVenn:
Closes #760

### DIFF
--- a/backend/api/Controllers/MissionController.cs
+++ b/backend/api/Controllers/MissionController.cs
@@ -158,6 +158,7 @@ public class MissionController : ControllerBase
     [ProducesResponseType(StatusCodes.Status400BadRequest)]
     [ProducesResponseType(StatusCodes.Status401Unauthorized)]
     [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
     [ProducesResponseType(StatusCodes.Status500InternalServerError)]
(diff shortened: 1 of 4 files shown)


backend/api/Controllers/MissionController.cs
```diff
@@ -234,6 +234,51 @@ public async Task<ActionResult<byte[]>> GetMap([FromRoute] string id)
...
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    public async Task<ActionResult<Mission>> Create(
+        [FromBody] CustomMissionQuery customMissionQuery
+    )
+    {
+        var robot = await _robotService.ReadById(customMissionQuery.RobotId);
+        if (robot is null)
+            return NotFound($"Could not find robot with id {customMissionQuery.RobotId}");
```
aestene, line 256: Should a 404 response be included as `ProducesResponseType`?

GodVenn: Yes, good catch!

aestene (approved):

PR #779, opened 2023-04-27
Add task description to mission page

Eddasol:


GodVenn:
Could you add the mission description under the mission title in this same PR?

PR #776, opened 2023-04-27
Safe zones for robot

Christdej:

PR #786, opened 2023-05-04
Add stop all button

mrica-equinor:


Eddasol:
Waiting for #776

aestene:

backend/api/Controllers/RobotController.cs
```diff
@@ -776,27 +784,6 @@ public async Task<ActionResult> ResumeMission([FromRoute] string robotId)
             return NotFound("No safe positions found");
         }
 
-        try
-        {
-            await _isarService.StopMission(robot);
```
andchiind, line 781: What happens if we use this endpoint before the last mission is done? Will sending a new mission to isar without cancelling the last simply override the last one?

mrica-equinor: Basically it couldn't stop the mission because the mission was already stopped on the frontend so it would return the error in line 789. I decided to handle this on the frontend also because you want to send all robots to a safe zone even if they don't have any missions.

mrica-equinor: I think if you send a new mission to isar before the last one is finished it adds it to the queue in a Pending state

andchiind: I think it would be better to do most handling in the backend since we cannot always assume that requests are coming from our frontend. However if this solution never breaks the functionality when just using the backend then this solution is fine. Otherwise it would be better to fix the backend so that it doesn't crash if there is no mission ongoing. I think with this current change, sending the robot to a safezone using the backend will just queue it up which might take a while.

mrica-equinor: I can have the ongoing mission being stop on the backend like before, and also change it so the the pending missions are deleted in the backend as well. 

It shouldn't queue the drive to safe position mission because the other have been deleted/stopped (and you await for this before posting the drive to safe mission). But I will look into having the backend handling this instead and then you can see what you think

mrica-equinor: I have made the changes, isar is getting a conflict so will look into it. But looks good and I do agree that handling it in the backend is much better and also easier I think. Let me know :)

aestene:

PR #787, opened 2023-05-05
Button to re-run missions

GodVenn:
This feature enables us to easily re-run historic missions.

When clicking the icon, the popup menu appears on the side to choose full mission or partial rerun

image

image

The current implementation redirects the user to the front page.

This was done because the idea is that the front page will show the user the new mission in a context, as it may not be started right away but rather added to the queue, in which case it makes sense to see it in the queue.
The front page is also where the user will usually be after scheduling a mission.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -26,7 +26,7 @@
                 "react-dom": "^18.2.0",
                 "react-error-boundary": "^3.1.4",
                 "react-modal": "^3.15.1",
-                "react-router-dom": "^6.3.0",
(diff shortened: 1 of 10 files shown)


GodVenn:
Please test locally.

When i was testing, something weird happened with the autocomplete boxes, could be due to some updated packages?
Wanted to see if this was just for me, or just locally, or if it happens in Aurora as well.

aestene (approved):

aestene (approved):

PR #790, opened 2023-05-08
Use int for estimated duration

GodVenn:
The old type "TimeSpan" is not widely supported.
Int is simpler and leaves less room for misinterpretation, no need for string parsing on the client side of our API.


Christdej (approved):
LGTM

PR #797, opened 2023-05-11
Associate robot with plant

### LINKED ISSUE #743: A robot should be associated with a plant
The robot model should be associated with the name of the asset it is located on.

- A field should be added to the robot model which gives the asset which the robot is located on.

This will enable us to display only robots on the chosen asset.

Not affected

OleDrange:
closes #743 
Connected to https://github.com/equinor/isar/pull/463
Waiting for #776 to merge before add database migrations

### DIFF
--- a/backend/api/Database/Models/Robot.cs
+++ b/backend/api/Database/Models/Robot.cs
@@ -26,6 +26,8 @@ namespace Api.Database.Models
         [MaxLength(200)]
         public string SerialNumber { get; set; }
 
+        public string CurrentAsset { get; set; }
+
         public float BatteryLevel { get; set; }
 
         public float? PressureLevel { get; set; }
@@ -69,6 +71,7 @@ namespace Api.Database.Models
             IsarId = "defaultIsarId";
             Name = "defaultId";
             SerialNumber = "defaultSerialNumber";
+            CurrentAsset = "defaultAsset";
             Status = RobotStatus.Offline;
             Enabled = false;
             Host = "localhost";
@@ -93,6 +96,7 @@ namespace Api.Database.Models
             IsarId = createQuery.IsarId;
             Name = createQuery.Name;
             SerialNumber = createQuery.SerialNumber;
+            CurrentAsset = createQuery.CurrentAsset;
             VideoStreams = videoStreams;
             Host = createQuery.Host;
             Port = createQuery.Port;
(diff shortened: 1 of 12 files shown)


frontend/src/components/Pages/FrontPage/RobotCards/RobotStatusView.tsx
```diff
@@ -40,10 +44,26 @@ export function RobotStatusSection({ refreshInterval }: RefreshProps) {
...
+
+    if (ActiveAsset() === '') {
+        robotDisplay = robots.map(function (robot) {
+            return <RobotStatusCard key={robot.id} robot={robot} />
+        })
+    } else {
+        robotDisplay = filteredRobots.map(function (robot) {
+            return <RobotStatusCard key={robot.id} robot={robot} />
+        })
+    }
```
GodVenn, line 65: Have you tested that the robots change when you change the asset?

GodVenn (commented):
Should also add an update of robot on receiving IsarRobotInfo if the old robotAsset is different than the new one (like we currently are doing for the robot.Port and robot.Host (Line 178 - 188 in `MqttEventHandler.cs`)

backend/api/Database/Models/Robot.cs
```diff
@@ -26,6 +26,8 @@ public class Robot
         [MaxLength(200)]
         public string SerialNumber { get; set; }
 
+        public string RobotAsset { get; set; }
```
GodVenn, line 29:
```suggestion
        public string CurrentAsset { get; set; }
```
Don't include the name of the parent object as part of the child parameter

frontend/src/components/Pages/FrontPage/RobotCards/RobotStatusView.tsx
OleDrange: Yes

backend/api/Database/Models/Robot.cs
OleDrange: done

OleDrange:
> Should also add an update of robot on receiving IsarRobotInfo if the old robotAsset is different than the new one (like we currently are doing for the robot.Port and robot.Host (Line 178 - 188 in `MqttEventHandler.cs`)

done

frontend/src/models/Robot.ts
```diff
@@ -14,6 +14,7 @@ export interface Robot {
     name?: string
     model: RobotModel
     serialNumber?: string
+    currentAsset?: string
```
GodVenn, line 17:
```suggestion
    currentAsset: string
```
If its not nullable in the backend it shouldn't be nullable here.

GodVenn (commented):
I think the database update will fail due to adding a required column,but we can look at it together to fix that

frontend/src/models/Robot.ts
OleDrange: done

GodVenn (approved):
LGTM

PR #800, opened 2023-05-12
Update to Pascal case for react component

UsamaEquinorAFK:


GodVenn (commented):
Isn't this a function, and not a react component?

UsamaEquinorAFK:
> Isn't this a function, and not a react component?

It is or should be a react component in my understanding since we are using a hook inside.

OleDrange (approved):
lgtm

PR #802, opened 2023-05-12
Fix all warnings in frontend

Christdej:
Removing unused code, add callbacks and other minor changes.

### DIFF
--- a/broker/README.md
+++ b/broker/README.md
@@ -23,6 +23,8 @@ From the flotilla root directory, run the following command:
 docker compose up --build broker
 ```
 
+If address is already in use, you might have to kill the mosquitto process `sudo pkill mosquitto`
+
 ## Authentication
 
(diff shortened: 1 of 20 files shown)


GodVenn (commented):
Are you sure this works as expected? I have bad experiences with adding parameters in the useEffect parameter array

Christdej:
> Are you sure this works as expected? I have bad experiences with adding parameters in the useEffect parameter array

Yes, I have tested quite a bit locally atleast. And I agreed that there can be problems adding parameters in the useEffect parameter array, if you dont do it correctly. That is why I have added useCallback on most of the parameters in the parameters array, to try to ensure correct use, such that we dont create a new instance of an object every re-render of the object. I think there is only a few places I have added parameters to the array without adding callbacks or other measures to ensure it is not bugged

aestene (approved):

PR #810, opened 2023-05-24
Localization map

Eddasol:


Christdej (approved):
LGTM

Could consider adding a test of some sort

PR #812, opened 2023-05-25
Disallow changing of asset in mission view

### LINKED ISSUE #497: Remove option to change asset when in mission view
Option to change asset while being in the mission view should be removed. It doesn't make sense to change asset while viewing a mission that's related to a specific asset.

Avoid mismatch between asset and mission

N/A

Image

andchiind:
Closes #497


Eddasol:
Try to clean up the commits such that we don't have merge commits and formatting commits

Eddasol (approved):
LGTM :)

PR #815, opened 2023-05-26
Add DeckID to database

UsamaEquinorAFK:
Fetches Deck id from front end and adds to database, migration steps will follow after review.

### DIFF
--- a/backend/api/Controllers/Models/CreateRobotQuery.cs
+++ b/backend/api/Controllers/Models/CreateRobotQuery.cs
@@ -23,5 +23,7 @@ namespace Api.Controllers.Models
         public bool Enabled { get; set; }
 
         public RobotStatus Status { get; set; }
+
+        public AssetDeck CurrentAssetDeck { get; set; }
     }
 }
--- a/backend/api/Controllers/Models/CreateRobotQuery.cs
+++ b/backend/api/Controllers/Models/CreateRobotQuery.cs
@@ -12,8 +12,6 @@ namespace Api.Controllers.Models
 
         public string SerialNumber { get; set; }
 
-        public string CurrentAsset { get; set; }
-
         public IList<CreateVideoStreamQuery> VideoStreams { get; set; }
 
         public string Host { get; set; }
--- a/backend/api/Controllers/Models/ScheduleLocalizationMissionQuery.cs
+++ b/backend/api/Controllers/Models/ScheduleLocalizationMissionQuery.cs
@@ -1,11 +1,8 @@
-﻿using Api.Database.Models;
-
-namespace Api.Controllers.Models
+﻿namespace Api.Controllers.Models
 {
     public struct ScheduleLocalizationMissionQuery
     {
         public string RobotId { get; set; }
-        public Pose LocalizationPose { get; set; }
         public string DeckId { get; set; }
     }
 }
--- a/backend/api/Controllers/Models/ScheduleLocalizationMissionQuery.cs
+++ b/backend/api/Controllers/Models/ScheduleLocalizationMissionQuery.cs
@@ -1,0 +1,11 @@
+﻿using Api.Database.Models;
+
+namespace Api.Controllers.Models
+{
+    public struct ScheduleLocalizationMissionQuery
+    {
+        public string RobotId { get; set; }
+        public Pose LocalizationPose { get; set; }
+        public string DeckId { get; set; }
+    }
+}
--- a/backend/api/Controllers/RobotController.cs
+++ b/backend/api/Controllers/RobotController.cs
@@ -674,6 +674,7 @@ public class RobotController : ControllerBase
 
         var deck = await _assetDeckService.ReadById(scheduleLocalizationMissionQuery.DeckId);
 
+
         if (deck == null)
         {
             _logger.LogWarning("Could not find deck with id={id}", scheduleLocalizationMissionQuery.DeckId);
@@ -694,7 +695,7 @@ public class RobotController : ControllerBase
         IsarMission isarMission;
         try
         {
-            isarMission = await _isarService.StartLocalizationMission(robot, scheduleLocalizationMissionQuery.LocalizationPose);
+            isarMission = await _isarService.StartLocalizationMission(robot, deck.DefaultLocalizationPose);
         }
         catch (HttpRequestException e)
         {
--- a/backend/api/Controllers/RobotController.cs
+++ b/backend/api/Controllers/RobotController.cs
@@ -643,7 +643,7 @@ public class RobotController : ControllerBase
     /// </remarks>
     [HttpPost]
     [Authorize(Roles = Role.User)]
-    [Route("{robotId}/start-localization")]
+    [Route("start-localization")]
     [ProducesResponseType(typeof(Mission), StatusCodes.Status200OK)]
     [ProducesResponseType(StatusCodes.Status400BadRequest)]
     [ProducesResponseType(StatusCodes.Status401Unauthorized)]
@@ -651,14 +651,14 @@ public class RobotController : ControllerBase
     [ProducesResponseType(StatusCodes.Status404NotFound)]
     [ProducesResponseType(StatusCodes.Status500InternalServerError)]
     public async Task<ActionResult<Mission>> StartLocalizationMission(
-        [FromRoute] string robotId,
-        [FromBody] Pose localizationPose
+        [FromBody] ScheduleLocalizationMissionQuery scheduleLocalizationMissionQuery
     )
     {
-        var robot = await _robotService.ReadById(robotId);
+        var robot = await _robotService.ReadById(scheduleLocalizationMissionQuery.RobotId);
+
         if (robot == null)
         {
-            _logger.LogWarning("Could not find robot with id={id}", robotId);
+            _logger.LogWarning("Could not find robot with id={id}", scheduleLocalizationMissionQuery.RobotId);
             return NotFound("Robot not found");
         }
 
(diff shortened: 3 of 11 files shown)


backend/api/Database/Models/Robot.cs
```diff
@@ -28,6 +28,8 @@ public class Robot
 
         public string CurrentAsset { get; set; }
 
+        public string CurrentAssetDeckId { get; set; }
+
```
Christdej, line 32: Is it perhaps better to use
```suggestion
        public AssetDeck CurrentAssetDeckId { get; set; }
```

Christdej (changes requested):

backend/api/Database/Models/Robot.cs
UsamaEquinorAFK: Do we want to remove CurrentAsset as well and replace it with CurrentAssetDeckId ?

Eddasol: We chouldn't change currentAsset but "string CurrentAssetDeckId" should be switched out with "AssetDeck CurrentAssetDeck"

Christdej: Ideally we replace both `string CurrentAsset` and `string CurrentAssetDeckId` with `AssetDeck CurrentAssetDeck`

If you get some issues with this, I can probably help =)

Christdej (changes requested):

backend/api/Controllers/Models/ScheduleLocalizationMissionQuery.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Api.Database.Models;
+
+namespace Api.Controllers.Models
+{
+    public struct ScheduleLocalizationMissionQuery
+    {
+        public string RobotId { get; set; }
+        public Pose LocalizationPose { get; set; }
+        public string DeckId { get; set; }
```
Christdej, line 9: At the moment an AssetDeck has one and only one DefaultLocalizationPose, why do we need both a localizationPose and a DeckId if we fetch the AssetDeck from the database? Then we can just read the Pose from the AssetDeck

backend/api/Controllers/RobotController.cs
```diff
@@ -625,7 +633,7 @@ public async Task<ActionResult> ResumeMission([FromRoute] string robotId)
         IsarMission isarMission;
         try
         {
-            isarMission = await _isarService.StartLocalizationMission(robot, localizationPose);
+            isarMission = await _isarService.StartLocalizationMission(robot, scheduleLocalizationMissionQuery.LocalizationPose);
```
Christdej, line 636: See comment above, get the LocalizationPose from the deck you got from assetDeckService rather than scheduleLocalizationMissionQuery.

Christdej (changes requested):
Also keep in mind that we are doing major changes to AssetDeck soon, then this will be done in another way

backend/api/Controllers/RobotController.cs
UsamaEquinorAFK: I will look at these, currently I am trying troubleshoot a problem:  ---> Microsoft.Data.Sqlite.SqliteException (0x80004005): SQLite Error 19: 'NOT NULL constraint failed: AssetDecks.AssetCode'. For some reason I am initializing the AssetCode as null somewhere and since it is required so it is not nullable.

backend/api/Controllers/Models/ScheduleLocalizationMissionQuery.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Api.Controllers.Models
+{
+    public struct ScheduleLocalizationMissionQuery
+    {
+        public string RobotId { get; set; }
+        public string DeckId { get; set; }
+    }
+}
```
Christdej, line 8: Can remove this file and then in the RobotController:
public async Task<ActionResult<Mission>> StartLocalizationMission(
        [FromRoute] string robotId, [FromRoute] string deckId
    )

Christdej (changes requested):

backend/api/Controllers/Models/CreateRobotQuery.cs
```diff
@@ -23,5 +23,7 @@ public struct CreateRobotQuery
         public bool Enabled { get; set; }
 
         public RobotStatus Status { get; set; }
+
+        public AssetDeck CurrentAssetDeck { get; set; }
```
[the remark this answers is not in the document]
UsamaEquinorAFK: One of them is for the asset and the other for the assetdeck ? Forexample look at flotilla\backend\api\Database\Models\Robot.cs you will find the place where they are both used at line 101 and 102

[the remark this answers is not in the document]
Christdej: `CurrentAsset` stores the same information as `AssetDeck.AssetCode` so I think we can remove it from both this file, and where it is used in `Robot.cs`. And elsewhere in the code, we update to use `Robot.CurrentAssetDeck.AssetCode` instead of `Robot.CurrentAsset`

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -20,6 +20,8 @@ public class MqttEventHandler : EventHandlerBase
         private readonly ILogger<MqttEventHandler> _logger;
         private readonly IServiceScopeFactory _scopeFactory;
 
+        public object CurrentAssetDeck { get; private set; }
```
Christdej, line 23: Is not this variable unused?

Christdej (changes requested):

backend/api/EventHandlers/MqttEventHandler.cs
UsamaEquinorAFK: Yes seems unused.

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -236,13 +237,13 @@ private static void UpdateHostIfChanged(string host, ref Robot robot, ref List<s
         private static void UpdateCurrentAssetIfChanged(string newCurrentAsset, ref Robot robot,
             ref List<string> updatedFields)
         {
-            if (newCurrentAsset.Equals(robot.CurrentAsset, StringComparison.Ordinal))
+            if (newCurrentAsset.Equals(robot.CurrentAssetDeck.AssetCode, StringComparison.Ordinal))
```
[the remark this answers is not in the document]
UsamaEquinorAFK: I still get warning of dereferencing to a "possibly null object", but the condition we added for this should fix it I think.

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -236,13 +235,13 @@ private static void UpdateHostIfChanged(string host, ref Robot robot, ref List<s
         private static void UpdateCurrentAssetIfChanged(string newCurrentAsset, ref Robot robot,
             ref List<string> updatedFields)
         {
-            if (newCurrentAsset.Equals(robot.CurrentAsset, StringComparison.Ordinal))
+            if ((robot.CurrentAssetDeck.AssetCode != null) && newCurrentAsset.Equals(robot.CurrentAssetDeck.AssetCode, StringComparison.Ordinal))
```
Christdej, line 238: You are still getting Dereference of a possible null reference because you are still trying to access robot.CurrentAssetDeck that might be null.

First you need to check that the `robot.CurrentAssetDeck` is not null

Christdej (changes requested):

Christdej (approved):
LGTM

frontend/src/api/ApiCaller.tsx
```diff
@@ -210,11 +210,12 @@ export class BackendAPICaller {
...
-    static async postLocalizationMission(localizationPose: Pose, robotId: string) {
-        const path: string = 'robots/' + robotId + '/start-localization'
+    static async postLocalizationMission(localizationPose: Pose, robotId: string, deck: string) {
+        const path: string = 'robots/start-localization'
         const body = {
-            position: localizationPose.position,
-            orientation: localizationPose.orientation,
+            robotId: robotId,
+            localizationPose: localizationPose,
+            deckId: deck,
```
eivindsjovold, line 218: why is deckId named deck? is not deckId a better name?

frontend/src/components/Pages/RobotPage/LocalizationSection.tsx
```diff
@@ -8,6 +8,39 @@ interface RobotProps {
 }
 
 export function LocalizationSection({ robot }: RobotProps) {
+    const [selectedAssetDeck, setSelectedAssetDeck] = useState<AssetDeck>()
+    const [assetDecks, setAssetDecks] = useState<AssetDeck[]>()
+
+    useEffect(() => {
+        BackendAPICaller.getAssetDecks().then((response: AssetDeck[]) => {
```
eivindsjovold: should we declare the api caller in the function instead of calling static?

frontend/src/components/Pages/RobotPage/LocalizationSection.tsx
```diff
@@ -8,6 +8,39 @@ interface RobotProps {
...
+            setAssetDecks(response)
+        })
+    }, [])
```
eivindsjovold: this will only run once. Is that intentional?

Eddasol: This was with an assumption that assetDeck is something that rarely/never change creating a limited need to update

frontend/src/api/ApiCaller.tsx
UsamaEquinorAFK: deck is a function argument, but I see your point.

frontend/src/components/Pages/RobotPage/LocalizationSection.tsx
UsamaEquinorAFK: I think this is outdated.

PR #817, opened 2023-05-30
Allows user to change mission order in queue

### LINKED ISSUE #766: Allow user to change order of mission queue
It should be possible for the user to change the order of missions.

Initially order should be the order they are scheduled in but with the option to change the order (e.g. by drag and drop)

andchiind:
Closes #766

### DIFF
--- a/frontend/src/components/Pages/FrontPage/MissionOverview/MissionQueueCard.tsx
+++ b/frontend/src/components/Pages/FrontPage/MissionOverview/MissionQueueCard.tsx
@@ -1,4 +1,4 @@
-import { Button, Card, Checkbox, Dialog, Icon, Typography } from '@equinor/eds-core-react'
+import { Button, Card, Checkbox, Chip, Dialog, Icon, Typography } from '@equinor/eds-core-react'
 import { config } from 'config'
 import { tokens } from '@equinor/eds-tokens'
 import { Mission } from 'models/Mission'
@@ -10,15 +10,23 @@ import { Icons } from 'utils/icons'
 
 interface MissionQueueCardProps {
     mission: Mission
+    order: number
     onDeleteMission: (mission: Mission) => void
+    onReorderMission: (mission: Mission, offset: number) => void
 }
 
 interface MissionDisplayProps {
     mission: Mission
 }
 
+const IconStyle = styled.div`
+    display: flex;
+    align-items: center;
+    flex-direction: row-reverse;
+`
+
 const StyledMissionCard = styled(Card)`
-    width: 900px;
+    width: 1080px;
     display: flex;
 `
 const HorizontalContent = styled.div`
@@ -28,7 +36,7 @@ const HorizontalContent = styled.div`
 `
 const HorizontalNonButtonContent = styled.div`
     display: grid;
-    grid-template-columns: 50px 400px auto 90px 180px;
+    grid-template-columns: 50px 20px 30px 30px 400px auto 90px 180px;
     align-items: center;
 `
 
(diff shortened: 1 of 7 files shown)


Eddasol:
Tested it, and it doesn't work if I schedule several missions at a time

frontend/src/components/Pages/FrontPage/MissionOverview/MissionQueueCard.tsx
```diff
@@ -53,15 +61,30 @@ export function MissionQueueCard({ mission, onDeleteMission }: MissionQueueCardP
     return (
         <StyledMissionCard key={mission.id} variant="default" style={{ boxShadow: tokens.elevation.raised }}>
             <HorizontalContent>
-                <HorizontalNonButtonContent onClick={routeChange}>
+                <HorizontalNonButtonContent>
                     <Checkbox />
-                    <Button variant="ghost" fullWidth>
+                    <Chip variant="active">
+                        <Typography variant="caption" color="#6F6F6F">
```
Eddasol, line 67: Suggest changing to this in order to match figma sketch:
<Typography variant="body_short_bold">

frontend/src/components/Pages/FrontPage/MissionOverview/MissionQueueCard.tsx
```diff
@@ -53,15 +62,30 @@ export function MissionQueueCard({ mission, onDeleteMission }: MissionQueueCardP
     return (
         <StyledMissionCard key={mission.id} variant="default" style={{ boxShadow: tokens.elevation.raised }}>
             <HorizontalContent>
-                <HorizontalNonButtonContent onClick={routeChange}>
+                <HorizontalNonButtonContent>
```
Eddasol, line 65: Can move the buttons outside of this style to avoid the need for moving route

andchiind: I am worried that this would overcomplicate the styling, since it currently uses a grid display

Eddasol: image

Moving some of the "grid-template-columns" arguments between the styles as this gives the same result as the current solution

andchiind: I have tried to be more consistent with the current approach by moving the buttons to the HorizontalNonButtonContent now, so all the buttons are near each other at the end of the row.

andchiind:
> Tested it, and it doesn't work if I schedule several missions at a time

After commit [b223b33](https://github.com/equinor/flotilla/pull/817/commits/b223b33b693c6b257e23530e395a1a2a4504aad4) I have been unable to recreate the bug we discussed. Are you OK with considering the current version as stable?

aestene:

andchiind:
This PR should hopefully become viable again once we remove the main mission scheduler loop

andchiind:
> If I remember correctly the main mission scheduler loop should now be removed. Is this viable again now?

I think the issue last time was partially related to the fact that the order in which we changed values in the database was not consistent, and since the queue is basically just a reference to the database, there may still be issues. But I can try it again, and in the worst case I can lock the queue when performing the reordering.

andchiind:
One potential solution would be to have an increment value which assigns an ordering to each mission in the database. Since we only care about the ordering of those in the queue and not the long since finished missions, we could then be safe simply using a large integers which wraps to zero when it is maximised, although for a 64bit value this would anyways not be a worry. In general the way in which we order missions in the queue is not very obvious as we simply rely on the inbuilt ordering.

PR #828, opened 2023-06-02
Calculate pose with quaternion clockwise rotation

### LINKED ISSUE #730: Quaternions in flotilla backend should be calculated with clockwise rotation, not counterclockwise
**Describe the bug**
The convertion from Echo's Axis-angle representation to quaternion makes the calculated pose rotate counterclockwise instead of clockwise, making the robot move to wrong pose

**To Reproduce**
1. Run mission with taurob or turtlebot
2. Observe that it rotates to wrong pose by doing counterclockwise rotation, not clockwise

**Expected behavior**
The resulting pose should be set following clockwise rotation

**Proposed solution**
Negate the angle used in convertion from Axis-Angle to quaternions

**Screenshots**
N/A

mrica-equinor:
Closes #730

### DIFF
--- a/backend/api.test/Database/Models.cs
+++ b/backend/api.test/Database/Models.cs
@@ -68,7 +68,7 @@ namespace Api.Test.Services
             if (orientation.Z >= 0)
                 angle = 2 * MathF.Acos(qw);
 
-            return new AxisAngle(new EchoVector(0, 0, 1), angle);
+            return new AxisAngle(new EchoVector(0, 0, 1), -angle);
         }
 
         public class AxisAngle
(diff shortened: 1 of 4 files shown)


Christdej (approved):
LGTM, does not require database migrations as only functions have changed, not properties stored in the database =)

mrica-equinor:
Not even for the ones on the backend/api/Services/Models/EchoPoseResponse.cs file?

Christdej:
> Not even for the ones on the backend/api/Services/Models/EchoPoseResponse.cs file?

Only for files in the backend/api/Database folder migrations are needed, as these are the only files stored in the database

PR #829, opened 2023-06-02
Add test for assetDeck map

Eddasol:


Christdej (approved):
LGTM, does not require database migration

PR #831, opened 2023-06-02
Make EchoMissionId nullable on Mission

### LINKED ISSUE #794: Make EchoMissionId nullable on Mission
As we are making custom mission the EchoMissionId in Mission should be nullable to clearify that this mission was not from Echo.

mrica-equinor:
Closes #794


Eddasol (approved):
LGTM :)

PR #840, opened 2023-06-07
Add mission frequency & decks to model and API

### LINKED ISSUE #836: Specify/store asset deck and mission frequency when creating custom mission
The asset deck and mission frequency should optionally be provided when creating custom missions.

This asset deck and mission frequency should be stored in the mission model, but does not need to be utilised by ISAR. If a frequency is specified but no deck then an error should be returned.

andchiind:
Closes #836

### DIFF
--- a/backend/api/Controllers/MissionController.cs
+++ b/backend/api/Controllers/MissionController.cs
@@ -254,6 +254,10 @@ public class MissionController : ControllerBase
 
         var missionTasks = customMissionQuery.Tasks.Select(task => new MissionTask(task)).ToList();
 
+        // TODO: verify customMissionQuery.AssetDeck against customMissionQuery.AssetCode
+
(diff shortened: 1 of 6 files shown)


backend/api/Controllers/MissionController.cs
```diff
@@ -254,7 +254,12 @@ public async Task<ActionResult<byte[]>> GetMap([FromRoute] string assetCode, str
 
         var missionTasks = customMissionQuery.Tasks.Select(task => new MissionTask(task)).ToList();
 
-        var scheduledMission = new Mission
+        // TODO: verify customMissionQuery.AssetDeck against customMissionQuery.AssetCode
```
andchiind, line 257: The assetDeck database is not currently populated

andchiind:
In the future we may want to add this information to records in another table which stores each unique mission as opposed to attaching it to records for each mission run.

PR #841, opened 2023-06-08
Add hover icons on ongoing mission card

### LINKED ISSUE #827: Missing hover icon when clicking mission on frontpage
**Describe the bug**
When hovering over the currently running mission on the frontpage there is no icon switch for clicking the button.

**To Reproduce**
1. Schedule mission in frontend
2. Hover mouse over currently running mission and observe

**Expected behavior**
Similarly to upcoming missions there should be a change in icon when hovering over a clickable object.

**Screenshots**
image

mrica-equinor:
Closes #827 

If anyone has an idea of how to remove the border on the mission title card, that would be great but I don't think the [eds-storybook-react](https://eds-storybook-react.azurewebsites.net/?path=/docs/surfaces-card--introduction) allows it. It could be changed into a button but I prefer how it looks with the card since it's bigger.


Christdej (approved):
LGTM, not sure how to remove the border no, but I think we merge this in anyways and think about the border in the future ;)

PR #842, opened 2023-06-08
Set boundaries to min and max values automatically

aestene:
If the coordinates were selected in the wrong order for a map the system would not work.

Also, if they were in the wrong order tha transformation would not be ideal. Thus, X1 should always be the smaller value compared to X2 etc.


aestene:
> bell Changes in database folder detected bell Do these changes require **adding new migrations**? thinking In that case follow [these steps](https://github.com/equinor/flotilla/tree/main/backend#Database-model-and-EF-Core). If you are uncertain, ask a database admin on the team smile

No need to run migrations based on this change.

PR #846, opened 2023-06-11
Add service to manage blob files

aestene:


backend/api/Services/BlobService.cs
```diff
@@ -0,0 +1,74 @@
...
+
+            AsyncPageable<BlobItem>? blobs;
+            try
+            {
+                blobs = blobContainerClient.GetBlobsAsync(BlobTraits.Metadata);
+            }
+            catch (RequestFailedException e)
+            {
+                _logger.LogWarning("Failed to fetch blob items because: {ErrorMessage}", e.Message);
+                return null;
```
OleDrange, line 54: Return error message instead of null?

backend/api/Services/BlobService.cs
```diff
@@ -0,0 +1,74 @@
...
+            var blobClient = blobContainerClient.GetBlobClient(blobName);
+
+            await using var stream = await blobClient.OpenReadAsync();
+
+            byte[] result = new byte[stream.Length];
+
+            _ = await stream.ReadAsync(result);
+
+            return result;
+        }
```
Christdej, line 39:
```suggestion
        public async Task<byte[]> DownloadBlob(string blobName, string containerName, string accountName)
        {
            var blobContainerClient = GetBlobContainerClient(containerName, accountName);
            var blobClient = blobContainerClient.GetBlobClient(blobName);
        
            using MemoryStream memoryStream = new MemoryStream();
            await blobClient.DownloadToAsync(memoryStream);
        
            return memoryStream.ToArray();
        }
```
Maybe this is a slightly better approach?

Christdej (changes requested):

backend/api/Services/BlobService.cs
```diff
@@ -0,0 +1,71 @@
...
+            }
+            catch (RequestFailedException e)
+            {
+                string errorMessage = $"Failed to fetch blob items because: {e.Message}";
+                _logger.LogError(e, "{ErrorMessage}", errorMessage);
+                throw;
+            }
+
+            return blobs;
+        }
```
Christdej, line 55:
```suggestion
        public AsyncPageable<BlobItem> FetchAllBlobs(string containerName, string accountName)
        {
            try
            {
                var blobContainerClient = GetBlobContainerClient(containerName, accountName);
                return blobContainerClient.GetBlobsAsync(BlobTraits.Metadata);
            }
            catch (RequestFailedException e)
            {
                string errorMessage = $"Failed to fetch blob items because: {e.Message}";
                _logger.LogError(e, "{ErrorMessage}", errorMessage);
                throw;
            }
        }
```

Christdej (changes requested):

Christdej (approved):

PR #855, opened 2023-06-15
Change language selection

OleDrange:


Eddasol:
> Looks like this: image
> 
> Edit: Flags not allowed, changed to look like equinor.com language selector image

A few suggestions:
* The language that is selected should be highlighted
* Should get background color when hovering (like the buttons to the left)
* Line between should be same color and preferably a bit shorter

OleDrange:
> 

Implemented the highlight when hovering, and looks like this now:
image

Christdej (approved):

PR #858, opened 2023-06-16
Rename missionMap to mapMetadata

Eddasol:
missionMap is used both for missions and for assetDecks. Should therefor be renamed to something less mission specific


FuzzyReality (approved):
LGTM

PR #859, opened 2023-06-16
Update EDS URL

mrica-equinor:


frontend/public/index.html
```diff
@@ -7,8 +7,8 @@
         <meta name="theme-color" content="#000000" />
         <meta name="description" content="Web site created using create-react-app" />
         <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
-        <link rel="stylesheet" href="https://eds-static.equinor.com/font/equinor-font.css" />
-        <link rel="stylesheet" href="https://eds-static.equinor.com/font/equinor-font.css" />
+        <link rel="stylesheet" href="https://cdn.eds.equinor.com/font/equinor-font.css" />
```
andchiind, line 10: Hmm, not sure why there were 2 links before. It looks good to me as is, but unless someone knows why there were 2 previously, we can probably delete one now.

andchiind: There will be no harm from having 2 instead of 1, but it can be good to have 1 so that people in the future won't also have this same conversation about why there were 2.

PR #864, opened 2023-06-19
Mission run refactor

andchiind:


backend/api.test/EndpointTest.cs
```diff
@@ -190,7 +190,8 @@ public async Task StartMissionTest()
             var query = new ScheduledMissionQuery
             {
                 RobotId = robotId,
-                AssetCode = "test",
+                AssetCode = "JSV",
```
aestene, line 193:
```suggestion
                AssetCode = "test",
```

Unless the `AssetCode` is required to be `JSV` I would put it as an anonymous name.

aestene: Or even better, use the arrange setup which is used in the "AreaTest" function.

flotilla.code-workspace
```diff
@@ -0,0 +1,28 @@
...
+			{
+				"name": "Launch flotilla",
+				"configurations": [
+					"Launch backend",
+                    "Launch frontend"
+				]
+			}
+		]
+	}
+}
```
aestene, line 28: .gitignore this file

backend/api/Database/Models/Area.cs
```diff
@@ -4,19 +4,26 @@
...
         [Required]
-        [MaxLength(200)]
-        public string DeckName { get; set; }
+        public Installation Installation { get; set; }
+
+        [Required]
+        public Asset Asset { get; set; }
+
+        [Required]
+        public string Name { get; set; }
```
aestene, line 23:
```suggestion
        [Required]
        [MaxLength(50)]
        public string Name { get; set; }
```

backend/api/Database/Models/MissionDefinition.cs
```diff
@@ -0,0 +1,32 @@
...
+
+#pragma warning disable CS8618
+namespace Api.Database.Models
+{
+    public class MissionDefinition
+    {
+        [Key]
+        [Required]
+        [MaxLength(200)]
+        public string Id { get; set; }
```
aestene, line 11: Why is this one set by us and not database generated like the others? 

I think it should be generated by the database to ensure uniqueness. Searching for `MissionDefinitions` may be done through an API call which can have options for name, area etc.

backend/api/Database/Models/MissionDefinition.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        [MaxLength(1000)]
+        public string? Comment { get; set; }
+
+        public TimeSpan? InspectionFrequency { get; set; }
+
+        public MissionRun? LastRun { get; set; }
+
+        [Required]
+        public string AssetCode { get; set; }
```
aestene, line 28:
```suggestion
        [Required]
        public string AssetCode { get; set; }
        
        public TimeSpan? InspectionFrequency { get; set; }

        public MissionRun? LastRun { get; set; }
```

Group the fields that are `Required` first and afterwards the optional fields.

backend/api/Database/Models/MissionRun.cs
```diff
@@ -5,14 +5,14 @@
...
-    public class Mission
+    public class MissionRun
     {
         [Key]
         [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
         public string Id { get; set; }
 
         [MaxLength(200)]
-        public int? EchoMissionId { get; set; }
+        public string? MissionId { get; set; }
```
aestene, line 15:
```suggestion
        public MissionDefinition MissionDefinition { get; set; }
```

I don't think this should be optional, please correct me if I'm wrong. Also if we are pointing to the ID it makes more sense to point straight to the model.

aestene: In addition, I think this field will be breaking for all old missions in the database.

aestene: You could set the `MissionDefinition` as optional for now such that this does not break compatability with old missions and then create an issue which is to add a `MissionDefinition` for all old missions such that this one may be made required.

backend/api/Database/Models/MissionRun.cs
```diff
@@ -31,9 +31,10 @@ public class Mission
         public string? Comment { get; set; }
 
         [Required]
-        [MaxLength(200)]
```
aestene, line 34: Should this one have been removed? I don't see the reason for this change.

backend/api/Database/Models/MissionRun.cs
```diff
@@ -31,9 +31,10 @@ public class Mission
...
         public string AssetCode { get; set; }
 
+        public Area? Area { get; set; }
+
```
aestene, line 37: Would also group the `Required` fields in this class.

backend/api/Database/Models/RobotModel.cs
```diff
@@ -79,7 +79,7 @@ public void Update(UpdateRobotModelQuery updateQuery)
         /// Updates the <see cref="AverageDurationPerTag"/> based on the data in the <paramref name="recentMissionsForModelType"/> provided
         /// </summary>
         /// <param name="recentMissionsForModelType"></param>
-        public void UpdateAverageDurationPerTag(List<Mission> recentMissionsForModelType)
+        public void UpdateAverageDurationPerTag(List<MissionRun> recentMissionsForModelType)
```
aestene, line 82:
```suggestion
        public void UpdateAverageDurationPerTag(List<MissionRun> recentMissionRunsForModelType)
```

backend/api/Database/Models/RobotModel.cs
```diff
@@ -79,7 +79,7 @@ public void Update(UpdateRobotModelQuery updateQuery)
...
         {
             if (recentMissionsForModelType.Any(mission => mission.Robot.Model.Type != Type))
```
aestene, line 84:
```suggestion
            if (recentMissionsForModelType.Any(missionRun => missionRun.Robot.Model.Type != Type))
```

aestene: Go through the rest of the file and check this.

I would also to a search throughout the codebase for things named `mission`. It should be very clear now that we use either `missionRun` or `missionDefinition` as variable names. If we start using `mission` about both we will confuse ourselves.

backend/api/Database/Models/Source.cs
```diff
@@ -0,0 +1,24 @@
...
+    public class Source
+    {
+        [Key]
+        [Required]
+        [MaxLength(200)]
+        public string Id { get; set; }
+
+        [Required]
+        public string URL { get; set; }
+
```
aestene, line 15:
```suggestion
        [Required]
        public string sourceId { get; set; }

```

My thoughts have been that this URL field should point to either the echo mission or the custom mission. 

For echo mission that would be an ID: int (can be string)
For custom mission it would be a json file in blob storage: URL

If we call it `SourceId` we can have it as the int (as a string) for echo and the path or url to the json in blob storage for a custom mission.

backend/api/EventHandlers/IsarConnectionEventHandler.cs
```diff
@@ -18,8 +18,8 @@ public class IsarConnectionEventHandler : EventHandlerBase
         private IRobotService RobotService =>
             _scopeFactory.CreateScope().ServiceProvider.GetRequiredService<IRobotService>();
 
-        private IMissionService MissionService =>
-            _scopeFactory.CreateScope().ServiceProvider.GetRequiredService<IMissionService>();
+        private IMissionRunService MissionService =>
```
aestene, line 21:
```suggestion
        private IMissionRunService MissionRunService =>
```

aestene: Change this everywhere where applicable.

backend/api/EventHandlers/MissionScheduler.cs
```diff
@@ -13,10 +13,10 @@ public class MissionScheduler : BackgroundService
         private readonly int _timeDelay;
         private readonly IServiceScopeFactory _scopeFactory;
 
-        private IList<Mission> MissionQueue =>
+        private IList<MissionRun> MissionQueue =>
```
aestene, line 16:
```suggestion
        private IList<MissionRun> MissionRunQueue =>
```

backend/api/EventHandlers/MissionScheduler.cs
```diff
@@ -13,10 +13,10 @@ public class MissionScheduler : BackgroundService
...
             MissionService
```
aestene, line 17:
```suggestion
            MissionRunService
```

I would expect this line to break the code currently as you have renamed the `MissionService` to `MissionRunService`

backend/api/EventHandlers/MissionScheduler.cs
```diff
@@ -25,8 +25,8 @@ public class MissionScheduler : BackgroundService
                 )
                 .Result;
 
-        private IMissionService MissionService =>
-            _scopeFactory.CreateScope().ServiceProvider.GetRequiredService<IMissionService>();
+        private IMissionRunService MissionService =>
```
aestene, line 28:
```suggestion
        private IMissionRunService MissionRunService =>
```

backend/api/EventHandlers/MissionScheduler.cs
```diff
@@ -80,7 +80,7 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
             }
         }
 
-        private async Task StartMission(Mission queuedMission)
+        private async Task StartMission(MissionRun queuedMission)
```
aestene, line 83:
```suggestion
        private async Task StartMissionRun(MissionRun queuedMissionRun)
```

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -249,15 +249,15 @@ private static void UpdateHostIfChanged(string host, ref Robot robot, ref List<s
         private async void OnMissionUpdate(object? sender, MqttReceivedArgs mqttArgs)
         {
             var provider = GetServiceProvider();
-            var missionService = provider.GetRequiredService<IMissionService>();
+            var missionService = provider.GetRequiredService<IMissionRunService>();
```
aestene, line 252:
```suggestion
            var missionRunService = provider.GetRequiredService<IMissionRunService>();
```

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -327,7 +327,7 @@ private async void OnMissionUpdate(object? sender, MqttReceivedArgs mqttArgs)
                     .AddDays(-timeRangeInDays)
                     .ToUnixTimeSeconds();
                 var missionsForEstimation = await missionService.ReadAll(
```
aestene, line 329:
```suggestion
                var missionRunsForEstimation = await missionService.ReadAll(
```

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -350,7 +350,7 @@ private async void OnMissionUpdate(object? sender, MqttReceivedArgs mqttArgs)
         private async void OnTaskUpdate(object? sender, MqttReceivedArgs mqttArgs)
         {
             var provider = GetServiceProvider();
-            var missionService = provider.GetRequiredService<IMissionService>();
+            var missionService = provider.GetRequiredService<IMissionRunService>();
```
aestene, line 353:
```suggestion
            var missionRunService = provider.GetRequiredService<IMissionRunService>();
```

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -388,7 +388,7 @@ private async void OnTaskUpdate(object? sender, MqttReceivedArgs mqttArgs)
         private async void OnStepUpdate(object? sender, MqttReceivedArgs mqttArgs)
         {
             var provider = GetServiceProvider();
-            var missionService = provider.GetRequiredService<IMissionService>();
+            var missionService = provider.GetRequiredService<IMissionRunService>();
```
aestene, line 391:
```suggestion
            var missionRunService = provider.GetRequiredService<IMissionRunService>();
```

backend/api/appsettings.Development.json
```diff
@@ -35,6 +35,10 @@
     "MaxRetryAttempts": 5,
     "ShouldFailOnMaxRetries": false
   },
+  "Blob": {
+    "CustomMissionContainerName": "custommission",
+    "ConnectionString": ""
```
aestene, line 40:
```suggestion
    "CustomMissionContainerName": "custommission",
    "AccountName": ""

```

backend/api/Services/AreaService.cs
```diff
@@ -0,0 +1,247 @@
...
+
+        public async Task<IEnumerable<Area>> ReadAll()
+        {
+            return await GetAreas().ToListAsync();
+        }
+
+        private IQueryable<Area> GetAreas()
+        {
+            return _context.Areas.Include(a => a.SafePositions)
+                .Include(a => a.Deck).ThenInclude(d => d.Installation).ThenInclude(i => i.Asset);
```
aestene, line 57:
```suggestion
            return _context.Areas.Include(a => a.SafePositions)
                .Include(a => a.Deck).Include(d => d.Installation).Include(i => i.Asset);
```
Since the `Area` model now has all these included in a flat structure it should not need to request in a nested manner through `Deck`.

backend/api/Services/AreaService.cs
```diff
@@ -0,0 +1,247 @@
...
+        }
+
+        public async Task<Area?> ReadById(string id)
+        {
+            return await GetAreas()
+                .FirstOrDefaultAsync(a => a.Id.Equals(id));
+        }
+
+        public async Task<Area?> ReadByAssetAndName(Asset? asset, string name)
```
aestene, line 66:
```suggestion
        public async Task<Area?> ReadByAssetAndName(Asset? asset, string areaName)
```

backend/api/Services/AreaService.cs
```diff
@@ -0,0 +1,247 @@
...
+        public async Task<IEnumerable<Area>> ReadByAsset(string assetName)
+        {
+            var asset = await ReadAssetByName(assetName);
+            if (asset == null)
+                return new List<Area>();
+            return await _context.Areas.Where(a =>
+                a.Asset.Equals(asset)).Include(a => a.SafePositions).ToListAsync();
+        }
+
+        public async Task<Area?> ReadByAssetAndName(string assetName, string name)
```
aestene, line 85:
```suggestion
        public async Task<Area?> ReadByAssetAndName(string assetName, string areaName)
```

backend/api/Services/AreaService.cs
```diff
@@ -0,0 +1,247 @@
...
+        {
+            // TODO: can we assume that this combination will be unique? Are area names specific enough?
+            var asset = await ReadAssetByName(assetName);
+            if (asset == null)
+                return null;
+            return await _context.Areas.Where(a =>
+                a.Asset.Equals(asset) &&
+                a.Name.ToLower().Equals(name.ToLower())
+            ).Include(a => a.SafePositions).FirstOrDefaultAsync();
+        }
```
aestene, line 95: Group the overloaded functions together.

backend/api/Services/AreaService.cs
```diff
@@ -0,0 +1,247 @@
...
+            var asset = await ReadAssetByName(assetName);
+            if (asset == null)
+                return new List<Area>();
+            return await _context.Areas.Where(a =>
+                a.Asset.Equals(asset)).Include(a => a.SafePositions).ToListAsync();
+        }
+
+        public async Task<Area?> ReadByAssetAndName(string assetName, string name)
+        {
+            // TODO: can we assume that this combination will be unique? Are area names specific enough?
```
aestene, line 87:
```suggestion
```
I think we can assume that this is correct.

backend/api/Services/AreaService.cs
```diff
@@ -0,0 +1,247 @@
+﻿using Api.Controllers.Models;
+using Api.Database.Context;
+using Api.Database.Models;
+using Microsoft.EntityFrameworkCore;
+
+namespace Api.Services
+{
+    public interface IAreaService
+    {
```
aestene, line 9: There are several public functions in this class that have not been included in the interface. Consider if they should be included or made private.

backend/api/Services/AreaService.cs
```diff
@@ -0,0 +1,247 @@
...
+        public async Task<Asset?> ReadAssetByName(string asset)
+        {
+            return await _context.Assets.Where(a =>
+                a.ShortName.ToLower().Equals(asset.ToLower())
+            ).FirstOrDefaultAsync();
+        }
+
+        public async Task<Area> Create(CreateAreaQuery newAreaQuery, List<Pose> safePositions)
+        {
+            var sp = new List<SafePosition>();
```
aestene, line 140:
```suggestion
            var safePositions = new List<SafePosition>();
```

backend/api/Services/AreaService.cs
```diff
@@ -0,0 +1,247 @@
...
+            foreach (var p in safePositions)
```
aestene, line 141:
```suggestion
            foreach (var pose in safePositions)
```

backend/api/Services/AreaService.cs
```diff
@@ -0,0 +1,247 @@
...
+            if (asset == null)
+            {
+                asset = new Asset
+                {
+                    Name = "", // TODO:
+                    ShortName = newAreaQuery.AssetCode
+                };
+                await _context.Assets.AddAsync(asset);
+                await _context.SaveChangesAsync();
+            }
```
aestene, line 156: If the asset does not exist I would not create a new asset but rather fail and return a message to the user saying that the asset does not exist.

aestene: See comment at top of file on splitting into separate services.

backend/api/Services/AreaService.cs
```diff
@@ -0,0 +1,247 @@
...
+            {
+                installation = new Installation
+                {
+                    Asset = asset,
+                    Name = "", // TODO:
+                    ShortName = newAreaQuery.InstallationName
+                };
+                await _context.Installations.AddAsync(installation);
+                await _context.SaveChangesAsync();
+            }
```
aestene, line 169: Should not create a new installation but return a message to the user that the installation does not exist.

aestene: See comment at top of file on splitting into separate services.

backend/api/Services/AreaService.cs
```diff
@@ -0,0 +1,247 @@
...
+            {
+                deck = new Deck
+                {
+                    Installation = installation,
+                    Asset = asset,
+                    Name = newAreaQuery.DeckName
+                };
+                await _context.Decks.AddAsync(deck);
+                await _context.SaveChangesAsync();
+            }
```
aestene, line 182: Should not create a new deck but rather return a message to the user that the deck does not exist.

aestene: See comment at top of file on splitting into separate services.

backend/api/Services/AreaService.cs
```diff
@@ -0,0 +1,247 @@
...
+
+            var existingArea = await ReadAreaByAssetAndInstallationAndDeckAndName(
+                asset, installation, deck, newAreaQuery.AreaName);
+            if (existingArea != null)
+            {
+                // TODO: maybe just append safe positions, or return an error
+            }
```
aestene, line 189: If the area already exists do not change it at all, merely return an error. 

Updating safe positions should be a seperate API endpoint.

backend/api/Services/AreaService.cs
```diff
@@ -0,0 +1,247 @@
...
+            return newArea;
+        }
+
+        public async Task<Area> Create(CreateAreaQuery newArea)
+        {
+            var area = await Create(newArea, new List<Pose>());
+            return area;
+        }
+
+        public async Task<Area?> AddSafePosition(string asset, string name, SafePosition safePosition)
```
aestene, line 212:
```suggestion
        public async Task<Area?> AddSafePosition(string assetName, string areaName, SafePosition safePosition)
```

backend/api/Services/AreaService.cs
```diff
@@ -0,0 +1,247 @@
...
+        {
+            var area = await ReadByAssetAndName(asset, name);
+            if (area is null)
+            {
+                return null;
+            }
+            area.SafePositions.Add(safePosition);
```
aestene, line 219:
```suggestion
            }
            
            area.SafePositions.Add(safePosition);
```

aestene: Use some whitespace for readability.

backend/api/Services/AreaService.cs
```diff
@@ -0,0 +1,247 @@
+﻿using Api.Controllers.Models;
+using Api.Database.Context;
+using Api.Database.Models;
+using Microsoft.EntityFrameworkCore;
+
+namespace Api.Services
+{
+    public interface IAreaService
```
aestene, line 8: There are several functions in this class which have not been included in the interface. There are also some functions that relate to reading other entities than area, for example asset.

I suggest creating individual services for `AssetService`, `DeckService` and `InstallationService`. These files will be short but have the functionality to `ReadByName` and `ReadById` plus `Create` and `Delete`. These also need endpoints for create and delete. 

Then you can use these services in this file to reduce the complexity here. Also, the creation of these models will be handled outside the `AreaController` but then in separate controllers. More files, more lines but easier to read in every file.

backend/api/Controllers/Models/ShortMissionDefinition.cs
```diff
@@ -1,7 +1,7 @@
 ﻿# nullable disable
 namespace Api.Controllers.Models
 {
-    public class MissionDefinition
+    public class ShortMissionDefinition
```
aestene, line 4:
```suggestion
    public class ShortEchoMission
```

backend/api/Services/SourceService.cs
```diff
@@ -0,0 +1,96 @@
...
+using System.Text.Json;
+using Api.Database.Models;
+using Api.Options;
+using Azure.Storage.Blobs;
+using Microsoft.Extensions.Options;
+
+namespace Api.Services
+{
+
+    public interface ISourceService
```
aestene, line 12: Suggestion is to remove all functionality related to managing blobs from this file and use what is implemented in #865 instead to have it more modular.

Additionally, `SourceService` is currently only interacting with `CustomMission`. I would rename the entire service to `CustomMissionService`.

aestene: Will not review this file further as I expect it will change a lot because of this.

backend/api/Services/Models/IsarMissionDefinition.cs
```diff
@@ -25,7 +25,7 @@ public IsarMissionDefinition(List<IsarTaskDefinition> tasks)
             Tasks = tasks;
         }
 
-        public IsarMissionDefinition(Mission mission)
+        public IsarMissionDefinition(MissionRun mission)
```
aestene, line 28:
```suggestion
        public IsarMissionDefinition(MissionRun missionRun)
```

backend/api/Services/Models/IsarMissionDefinition.cs
```diff
@@ -47,7 +47,7 @@ public struct IsarTaskDefinition
         [JsonPropertyName("inspections")]
         public List<IsarInspectionDefinition> Inspections { get; set; }
 
-        public IsarTaskDefinition(MissionTask missionTask, Mission mission)
+        public IsarTaskDefinition(MissionTask missionTask, MissionRun mission)
```
aestene, line 50:
```suggestion
        public IsarTaskDefinition(MissionTask missionTask, MissionRun missionRun)
```

backend/api/Services/Models/IsarMissionDefinition.cs
```diff
@@ -81,7 +81,7 @@ public struct IsarInspectionDefinition
         [JsonPropertyName("metadata")]
         public Dictionary<string, string?>? Metadata { get; set; }
 
-        public IsarInspectionDefinition(Inspection inspection, MissionTask task, Mission mission)
+        public IsarInspectionDefinition(Inspection inspection, MissionTask task, MissionRun mission)
```
aestene, line 84:
```suggestion
        public IsarInspectionDefinition(Inspection inspection, MissionTask task, MissionRun missionRun)
```

backend/api/Services/EchoService.cs
```diff
@@ -88,6 +89,33 @@ public async Task<EchoMission> GetMissionById(int missionId)
             return mission;
         }
 
+        public async Task<EchoMission> GetMissionByPath(string relativePath)
```
aestene, line 92: Given the suggestion that the `Source` should contain the `id` for the echo mission and not the path I think this method might no longer be required.

backend/api/Services/EchoService.cs
```diff
@@ -170,17 +198,17 @@ private List<EchoTag> ProcessPlanItems(List<PlanItem> planItems, string installa
...
+            var availableMissions = new List<ShortMissionDefinition>();
 
             foreach (var echoMission in echoMissions)
             {
                 if (echoMission.PlanItems is null)
                     continue;
                 try
                 {
-                    var mission = new MissionDefinition()
+                    var mission = new ShortMissionDefinition()
```
aestene, line 211:
```suggestion
                    var condensedEchoMissionDefinition = new ShortMissionDefinition()
```

backend/api/Services/IsarService.cs
```diff
@@ -8,7 +8,7 @@ namespace Api.Services
 {
     public interface IIsarService
     {
-        public abstract Task<IsarMission> StartMission(Robot robot, Mission mission);
+        public abstract Task<IsarMission> StartMission(Robot robot, MissionRun mission);
```
aestene, line 11:
```suggestion
        public abstract Task<IsarMission> StartMission(Robot robot, MissionRun missionRun);
```

backend/api/Services/IsarService.cs
```diff
@@ -68,7 +68,7 @@ public IsarService(ILogger<IsarService> logger, IDownstreamWebApi downstreamWebA
             return response;
         }
 
-        public async Task<IsarMission> StartMission(Robot robot, Mission mission)
+        public async Task<IsarMission> StartMission(Robot robot, MissionRun mission)
```
aestene, line 71:
```suggestion
        public async Task<IsarMission> StartMission(Robot robot, MissionRun missionRun)
```

backend/api/Services/MapService.cs
```diff
@@ -9,7 +9,7 @@ public interface IMapService
     {
         public Task<byte[]> FetchMapImage(string mapName, string assetCode);
         public Task<MissionMap?> ChooseMapFromPositions(IList<Position> positions, string assetCode);
-        public Task AssignMapToMission(Mission mission);
+        public Task AssignMapToMission(MissionRun mission);
```
aestene, line 12:
```suggestion
        public Task AssignMapToMission(MissionRun missionRun);
```

backend/api/Services/MapService.cs
```diff
@@ -70,7 +70,7 @@ await foreach (var blob in blobs)
             return map;
         }
 
-        public async Task AssignMapToMission(Mission mission)
+        public async Task AssignMapToMission(MissionRun mission)
```
aestene, line 73:
```suggestion
        public async Task AssignMapToMission(MissionRun missionRun)
```

backend/api/Services/MissionDefinitionService.cs
```diff
@@ -0,0 +1,224 @@
...
+
+        public async Task<MissionDefinition> Create(MissionDefinition mission)
+        {
+            await _context.MissionDefinitions.AddAsync(mission);
+            await _context.SaveChangesAsync();
+
+            return mission;
+        }
+
+        private IQueryable<MissionDefinition> GetMissionsWithSubModels()
```
aestene, line 58:
```suggestion
        private IQueryable<MissionDefinition> GetMissionDefinitionsWithSubModels()
```

backend/api/Services/MissionDefinitionService.cs
```diff
@@ -0,0 +1,224 @@
...
+        {
+            return _context.MissionDefinitions
+                .Include(mission => mission.Area)
```
aestene, line 61:
```suggestion
                .Include(missionDefinition => missionDefinition.Area)
```

backend/api/Services/MissionDefinitionService.cs
```diff
@@ -0,0 +1,224 @@
...
+                .ThenInclude(robot => robot.Deck)
```
aestene, line 62:
```suggestion
                .ThenInclude(area => area.Deck)
```

backend/api/Services/MissionDefinitionService.cs
```diff
@@ -0,0 +1,224 @@
...
+                .ThenInclude(robot => robot.Installation)
+                .ThenInclude(robot => robot.Asset)
+                .Include(mission => mission.Source)
+                .Include(mission => mission.LastRun)
+                .ThenInclude(planTask => planTask == null ? null : planTask.StartTime);
+        }
```
aestene, line 68:
```suggestion
                .Include(missionDefinition => missionDefinition.Area)
                .ThenInclude(area => area.Deck)
                .ThenInclude(area => area.Installation)
                .ThenInclude(area => area.Asset)
                .Include(missionDefinition => missionDefinition.Source)
                .Include(missionDefinition => missionDefinition.LastRun)
                .ThenInclude(planTask => planTask == null ? null : planTask.StartTime);
        }
```

backend/api/Services/MissionDefinitionService.cs
```diff
@@ -0,0 +1,224 @@
...
+        private IQueryable<MissionDefinition> GetMissionsWithSubModels()
+        {
+            return _context.MissionDefinitions
+                .Include(mission => mission.Area)
+                .ThenInclude(robot => robot.Deck)
+                .ThenInclude(robot => robot.Installation)
+                .ThenInclude(robot => robot.Asset)
+                .Include(mission => mission.Source)
+                .Include(mission => mission.LastRun)
+                .ThenInclude(planTask => planTask == null ? null : planTask.StartTime);
```
aestene, line 67: Not certain what this line is supposed to do in this setting? A `LastRun` is a `MissionRun` which does not have something called `PlanTask`

backend/api/Services/MissionDefinitionService.cs
```diff
@@ -0,0 +1,224 @@
...
+            if (mission is null)
+            {
+                return null;
+            }
+
+            _context.MissionDefinitions.Remove(mission);
+            await _context.SaveChangesAsync();
+
+            return mission;
+        }
```
aestene, line 114: Should we have a delete functionality at all? Or should we instead have a field which may say that the `MissionDefinition` is deprecated or inactive? Thus we don't break anything for old `MissionRun` instances but we don't show `MissionDefinnition`objects that should not be shown?

backend/api/Services/MissionDefinitionService.cs
```diff
@@ -0,0 +1,224 @@
...
+
+        /// <summary>
+        /// Filters by <see cref="MissionDefinitionQueryStringParameters.AssetCode"/> and <see cref="MissionDefinitionQueryStringParameters.Status"/>
+        ///
+        /// <para>Uses LINQ Expression trees (see <seealso href="https://docs.microsoft.com/en-us/dotnet/csharp/expression-trees"/>)</para>
+        /// </summary>
+        /// <param name="parameters"> The variable containing the filter params </param>
+        private static Expression<Func<MissionDefinition, bool>> ConstructFilter(
+            MissionDefinitionQueryStringParameters parameters
+        )
```
aestene, line 135: Can this function be refactored? See very similar code in `MissionRunService`

backend/api/Services/MissionDefinitionService.cs
```diff
@@ -0,0 +1,224 @@
...
+                    Expression.Invoke(missionTypeFilter, mission)
+                    )
+                );
+
+            // Constructing the resulting lambda expression by combining parameter and body
+            return Expression.Lambda<Func<MissionDefinition, bool>>(body, mission);
+        }
+
+        private static void ApplySort(ref IQueryable<MissionDefinition> missions, string orderByQueryString)
+        {
```
aestene, line 169: Can this function be refactored? See very similar code in `MissionRunService`

backend/api/Services/MissionRunService.cs
```diff
@@ -12,17 +12,17 @@
 
 namespace Api.Services
 {
-    public interface IMissionService
+    public interface IMissionRunService
     {
-        public abstract Task<Mission> Create(Mission mission);
+        public abstract Task<MissionRun> Create(MissionRun mission);
```
aestene, line 17:
```suggestion
        public abstract Task<MissionRun> Create(MissionRun missionRun);
```

backend/api/Services/MissionRunService.cs
```diff
@@ -12,17 +12,17 @@
...
 
-        public abstract Task<PagedList<Mission>> ReadAll(MissionQueryStringParameters parameters);
+        public abstract Task<PagedList<MissionRun>> ReadAll(MissionRunQueryStringParameters parameters);
 
-        public abstract Task<Mission?> ReadById(string id);
+        public abstract Task<MissionRun?> ReadById(string id);
 
-        public abstract Task<Mission> Update(Mission mission);
+        public abstract Task<MissionRun> Update(MissionRun mission);
```
aestene, line 23:
```suggestion
        public abstract Task<MissionRun> Update(MissionRun missionRun);
```

backend/api/Services/MissionRunService.cs
```diff
@@ -12,17 +12,17 @@
...
 
-        public abstract Task<Mission?> UpdateMissionStatusByIsarMissionId(
+        public abstract Task<MissionRun?> UpdateMissionStatusByIsarMissionId(
```
aestene, line 25:
```suggestion
        public abstract Task<MissionRun?> UpdateMissionRunStatusByIsarMissionId(
```

backend/api/Services/MissionRunService.cs
```diff
@@ -58,20 +58,24 @@ IsarStepStatus stepStatus
...
 
-        public MissionService(FlotillaDbContext context, ILogger<MissionService> logger)
+        public MissionRunService(FlotillaDbContext context, ILogger<MissionRunService> logger)
         {
             _context = context;
             _logger = logger;
         }
 
-        private IQueryable<Mission> GetMissionsWithSubModels()
+        private IQueryable<MissionRun> GetMissionsWithSubModels()
```
aestene, line 72:
```suggestion
        private IQueryable<MissionRun> GetMissionRunsWithSubModels()
```

backend/api/Services/MissionRunService.cs
```diff
@@ -58,20 +58,24 @@ IsarStepStatus stepStatus
...
         {
-            return _context.Missions
+            return _context.MissionRuns
+                .Include(mission => mission.Area)
+                .ThenInclude(a => a.Deck)
+                .ThenInclude(d => d.Installation)
+                .ThenInclude(i => i.Asset)
```
aestene, line 78:
```suggestion
            return _context.MissionRuns
                .Include(mission => mission.Area)
                .ThenInclude(area => area.Deck)
                .ThenInclude(area => area.Installation)
                .ThenInclude(installation => area.Asset)
```

backend/api/Services/MissionRunService.cs
```diff
@@ -82,15 +86,15 @@ private IQueryable<Mission> GetMissionsWithSubModels()
                 .ThenInclude(task => task.Inspections);
         }
 
-        public async Task<Mission> Create(Mission mission)
+        public async Task<MissionRun> Create(MissionRun mission)
```
aestene, line 89:
```suggestion
        public async Task<MissionRun> Create(MissionRun missionRun)
```

backend/api/Services/MissionRunService.cs
```diff
@@ -103,27 +107,27 @@ public async Task<PagedList<Mission>> ReadAll(MissionQueryStringParameters param
...
 
-        public async Task<Mission?> ReadById(string id)
+        public async Task<MissionRun?> ReadById(string id)
         {
             return await GetMissionsWithSubModels()
                 .FirstOrDefaultAsync(mission => mission.Id.Equals(id));
         }
 
-        public async Task<Mission> Update(Mission mission)
+        public async Task<MissionRun> Update(MissionRun mission)
```
aestene, line 123:
```suggestion
        public async Task<MissionRun> Update(MissionRun missionRun)
```

backend/api/Services/MissionRunService.cs
```diff
@@ -103,27 +107,27 @@ public async Task<PagedList<Mission>> ReadAll(MissionQueryStringParameters param
...
         {
             var entry = _context.Update(mission);
             await _context.SaveChangesAsync();
             return entry.Entity;
         }
 
-        public async Task<Mission?> Delete(string id)
+        public async Task<MissionRun?> Delete(string id)
         {
             var mission = await GetMissionsWithSubModels()
```
aestene, line 132:
```suggestion
            var missionRun = await GetMissionsWithSubModels()
```

backend/api/Services/MissionRunService.cs
```diff
@@ -270,7 +274,7 @@ IsarStepStatus stepStatus
 
         #endregion ISAR Specific methods
 
-        private static void SearchByName(ref IQueryable<Mission> missions, string? name)
+        private static void SearchByName(ref IQueryable<MissionRun> missions, string? name)
```
aestene, line 277:
```suggestion
        private static void SearchByName(ref IQueryable<MissionRun> missionRuns, string? name)
```

backend/api/Services/MissionRunService.cs
```diff
@@ -281,7 +285,7 @@ private static void SearchByName(ref IQueryable<Mission> missions, string? name)
             );
         }
 
-        private static void SearchByRobotName(ref IQueryable<Mission> missions, string? robotName)
+        private static void SearchByRobotName(ref IQueryable<MissionRun> missions, string? robotName)
```
aestene, line 288:
```suggestion
        private static void SearchByRobotName(ref IQueryable<MissionRun> missionRuns, string? robotName)
```

backend/api/Services/MissionRunService.cs
```diff
@@ -291,7 +295,7 @@ private static void SearchByRobotName(ref IQueryable<Mission> missions, string?
             );
         }
 
-        private static void SearchByTag(ref IQueryable<Mission> missions, string? tag)
+        private static void SearchByTag(ref IQueryable<MissionRun> missions, string? tag)
```
aestene, line 298:
```suggestion
        private static void SearchByTag(ref IQueryable<MissionRun> missionRuns, string? tag)
```

backend/api.test/Services/MissionService.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
```
aestene, line 2: The file name here does not reflect the name change of the original file. File should be named `MissionRunService.cs`

backend/api.test/Services/MissionService.cs
```diff
@@ -15,14 +16,14 @@ namespace Api.Test.Services
...
+        private readonly ILogger<MissionRunService> _logger;
+        private readonly MissionRunService _missionService;
 
         public MissionServiceTest(DatabaseFixture fixture)
         {
             _context = fixture.NewContext;
-            _logger = new Mock<ILogger<MissionService>>().Object;
-            _missionService = new MissionService(_context, _logger);
+            _logger = new Mock<ILogger<MissionRunService>>().Object;
+            _missionService = new MissionRunService(_context, _logger);
```
aestene, line 26:
```suggestion
            _missionRunService = new MissionRunService(_context, _logger);
```

backend/api.test/Services/MissionService.cs
```diff
@@ -43,22 +44,49 @@ public async Task Create()
...
+                Name = "test test"
+            };
+            var testInstallation = new Installation
+            {
+                ShortName = "test",
+                Name = "test test",
+                Asset = testAsset
+            };
+
+            MissionRun mission =
```
aestene, line 62:
```suggestion
            MissionRun missionRun =
```

backend/api/Controllers/AreaController.cs
```diff
@@ -0,0 +1,248 @@
...
+        [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+        public async Task<ActionResult<Area>> Create([FromBody] CreateAreaQuery area)
+        {
+            _logger.LogInformation("Creating new area");
+            try
+            {
+                var existingArea = await _areaService.ReadByAssetAndName(area.AssetCode, area.AreaName);
+                if (existingArea != null)
+                {
+                    _logger.LogInformation("An area for given name and asset already exists");
```
aestene, line 107:
```suggestion
                    _logger.LogWarning("An area for given name and asset already exists");
```

backend/api/Controllers/AreaController.cs
```diff
@@ -0,0 +1,248 @@
...
+                    return BadRequest($"Area already exists");
```
aestene, line 108: I would use `Conflict` instead of `BadRequest` here.

backend/api/Controllers/AreaController.cs
```diff
@@ -0,0 +1,248 @@
...
+            catch (Exception e)
+            {
+                _logger.LogError(e, "Error while creating new area");
+                throw;
+            }
+        }
+
+
+        /// <summary>
+        /// Add a safe position to a area
```
aestene, line 131:
```suggestion
        /// Add safe position to an area
```

backend/api/Controllers/AreaController.cs
```diff
@@ -0,0 +1,248 @@
...
+        [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+        public async Task<ActionResult<Area>> AddSafePosition(
+            [FromRoute] string asset,
+            [FromRoute] string installationName,
+            [FromRoute] string deckName,
+            [FromRoute] string areaName,
+            [FromBody] Pose safePosition
+        )
+        {
+            _logger.LogInformation("Adding new safe position");
```
aestene, line 152:
```suggestion
            _logger.LogInformation("Adding new safe position to {Asset}, {Installtion}, {Deck}, {Area}", asset, installationName, deckName, areaName);
```

backend/api/Controllers/AreaController.cs
```diff
@@ -0,0 +1,248 @@
...
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+        public async Task<ActionResult<Area>> AddSafePosition(
+            [FromRoute] string asset,
+            [FromRoute] string installationName,
+            [FromRoute] string deckName,
+            [FromRoute] string areaName,
+            [FromBody] Pose safePosition
+        )
```
aestene, line 150:
```suggestion
        public async Task<ActionResult<Area>> AddSafePosition(
            [FromRoute] string assetName,
            [FromRoute] string installationName,
            [FromRoute] string deckName,
            [FromRoute] string areaName,
            [FromBody] Pose safePosition
        )
```

backend/api/Controllers/AreaController.cs
```diff
@@ -0,0 +1,248 @@
...
+                            AssetCode = asset,
+                            AreaName = areaName,
+                            InstallationName = installationName,
+                            DeckName = deckName,
+                            DefaultLocalizationPose = new Pose()
+                        },
+                        new List<Pose> { tempPose }
+                    );
+                    return CreatedAtAction(nameof(GetAreaById), new { id = area.Id }, area);
+                }
```
aestene, line 178: I would return a NotFound(area) instead of creating a new area. Creating a new area should be done through the dedicated endpoint and not as a side effect where one might not expect it. For example, if someone accidentely wrote the area name incorrectly we would suddenly get a new area in the database.

backend/api/Controllers/AreaController.cs
```diff
@@ -0,0 +1,248 @@
+﻿using Api.Controllers.Models;
+using Api.Database.Models;
+using Api.Services;
+using Api.Utilities;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Api.Controllers
```
aestene, line 8: Group the admin endpoints in order and then the role Any afterwards.

backend/api/Controllers/AreaController.cs
```diff
@@ -0,0 +1,248 @@
...
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(StatusCodes.Status404NotFound)]
+        [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+        public async Task<ActionResult<MissionMap>> GetMapMetadata([FromRoute] string id)
+        {
+            var area = await _areaService.ReadById(id);
+            if (area is null)
+            {
+                _logger.LogError("Area not found for area ID {areaId}", id);
+                return NotFound("Could not find this area");
```
aestene, line 223:
```suggestion
                string errorMessage = $"Area not found for area with ID {id}"
                _logger.LogError("{ErrorMessage}", errorMessage);
                return NotFound(errorMessage);
```

backend/api/Controllers/AreaController.cs
```diff
@@ -0,0 +1,248 @@
...
+                area.DefaultLocalizationPose.Position
+            };
+            try
+            {
+                map = await _mapService.ChooseMapFromPositions(positions, area.Deck.Installation.Asset.ShortName);
+            }
+            catch (ArgumentOutOfRangeException)
+            {
+                _logger.LogWarning("Unable to find a map for area '{areaId}'", area.Id);
+                return NotFound("Could not find map suited for the positions in this area");
```
aestene, line 238:
```suggestion
                string errorMessage = $"Unable to find map for area with ID {id}";
                _logger.LogWarning("{ErrorMessage}", errorMessage);
                return NotFound(errorMessage);
```

backend/api/Controllers/MissionController.cs
```diff
@@ -12,45 +12,54 @@ namespace Api.Controllers;
...
         _stidService = stidService;
         _logger = logger;
     }
 
     /// <summary>
-    /// List all missions in the Flotilla database
+    /// List all mission runs in the Flotilla database
     /// </summary>
     /// <remarks>
     /// <para> This query gets all missions </para>
```
aestene, line 52:
```suggestion
    /// <para> This query gets all mission runs </para>
```

backend/api/Controllers/MissionController.cs
```diff
@@ -66,10 +75,57 @@ IStidService stidService
             return BadRequest("Max EndTime cannot be less than min EndTime");
         }
 
-        PagedList<Mission> missions;
+        PagedList<MissionRun> missions;
```
aestene, line 78:
```suggestion
        PagedList<MissionRun> missionRuns;
```

backend/api/Controllers/MissionController.cs
```diff
@@ -66,10 +75,57 @@ IStidService stidService
...
+        );
+
+        return Ok(missions);
+    }
+
+    /// <summary>
+    /// List all mission definitions in the Flotilla database
+    /// </summary>
+    /// <remarks>
+    /// <para> This query gets all missions </para>
```
aestene, line 111:
```suggestion
    /// <para> This query gets all mission definitions </para>
```

backend/api/Controllers/MissionController.cs
```diff
@@ -66,10 +75,57 @@ IStidService stidService
...
+    [ProducesResponseType(typeof(IList<MissionDefinition>), StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    public async Task<ActionResult<IList<MissionDefinition>>> GetMissionDefinitions(
+        [FromQuery] MissionDefinitionQueryStringParameters parameters
+    )
+    {
+        // TODO: define new parameters using primarily area and source type
```
aestene, line 124:
```suggestion
```
Add this as a separated issue if it will not be implemented in this PR

aestene: This PR is big enough as it is so please do a new issue and a new PR later.

backend/api/Controllers/MissionController.cs
```diff
@@ -96,19 +152,38 @@ IStidService stidService
...
+    [Authorize(Roles = Role.Any)]
+    [Route("definitions/{id}")]
+    [ProducesResponseType(typeof(MissionRun), StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    public async Task<ActionResult<MissionDefinition>> GetMissionDefinitionById([FromRoute] string id)
+    {
+        var mission = await _missionDefinitionService.ReadById(id);
```
aestene, line 186:
```suggestion
        var missionDefinition = await _missionDefinitionService.ReadById(id);
```

backend/api/Controllers/MissionController.cs
```diff
@@ -141,20 +216,114 @@ public async Task<ActionResult<byte[]>> GetMap([FromRoute] string assetCode, str
     }
 
     /// <summary>
-    /// Schedule a new mission
+    /// Reschedule an existing mission
```
aestene, line 219:
```suggestion
    /// Schedule an existing mission definition
```
Not sure I agree with naming this a reschedule. The mission definitions may have changed since the last time as well so better to call it schedule.

backend/api/Controllers/MissionController.cs
```diff
@@ -141,20 +216,114 @@ public async Task<ActionResult<byte[]>> GetMap([FromRoute] string assetCode, str
...
+    [ProducesResponseType(typeof(MissionRun), StatusCodes.Status201Created)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    public async Task<ActionResult<MissionRun>> Rechedule(
+        [FromRoute] string missionId,
+        [FromBody] RescheduleMissionQuery rescheduledMissionQuery
+    )
```
aestene, line 235: Would include the missionId in the query object and thus in the body, not as part of the query string.

backend/api/Controllers/MissionController.cs
```diff
@@ -141,20 +216,114 @@ public async Task<ActionResult<byte[]>> GetMap([FromRoute] string assetCode, str
...
+            return NotFound($"Could not find robot with id {rescheduledMissionQuery.RobotId}");
+
+        MissionDefinition? missionDefinition;
+        try
+        {
+            missionDefinition = await _missionDefinitionService.ReadById(missionId);
+            if (missionDefinition == null)
+                return NotFound("Mission definition not found");
+        }
+        catch (HttpRequestException e)
```
aestene, line 248: Why are we catching and `HttpRequestException` here? The underlying functionality here is only calling the database for a `MissionDefinition`

backend/api/Controllers/MissionController.cs
```diff
@@ -141,20 +216,114 @@ public async Task<ActionResult<byte[]>> GetMap([FromRoute] string assetCode, str
...
+                            return new MissionTask(t, tagPosition);
+                        }
+                    )
+                    .ToList();
+                break;
+            case MissionSourceType.Custom:
+                missionTasks = _sourceService.GetMissionTasksFromURL(missionDefinition.Source.URL);
+                break;
+            default:
+                return BadRequest("Invalid mission source type provided");
```
aestene, line 283: Do an internal server error here. This bad request cannot happen as the MissionSourceType is collected from the mission definition itself and not from the request.

backend/api/Controllers/MissionController.cs
```diff
@@ -141,20 +216,114 @@ public async Task<ActionResult<byte[]>> GetMap([FromRoute] string assetCode, str
...
+        }
+
+        if (missionTasks == null)
+            return NotFound("No mission tasks were found for the requested mission");
+
+        var scheduledMission = new MissionRun
```
aestene, line 289:
```suggestion
        var missionRun = new MissionRun
```

backend/api/Controllers/MissionController.cs
```diff
@@ -141,20 +216,114 @@ public async Task<ActionResult<byte[]>> GetMap([FromRoute] string assetCode, str
...
+            Area = missionDefinition.Area,
+            Map = new MissionMap()
+        };
+
+        await _mapService.AssignMapToMission(scheduledMission);
+
+        if (scheduledMission.Tasks.Any())
+            scheduledMission.CalculateEstimatedDuration();
+
+        var newMission = await _missionRunService.Create(scheduledMission);
```
aestene, line 307:
```suggestion
        var newMissionRun = await _missionRunService.Create(scheduledMission);
```

backend/api/Controllers/MissionController.cs
```diff
@@ -141,20 +216,114 @@ public async Task<ActionResult<byte[]>> GetMap([FromRoute] string assetCode, str
...
-    [ProducesResponseType(typeof(Mission), StatusCodes.Status201Created)]
+    [ProducesResponseType(typeof(MissionRun), StatusCodes.Status201Created)]
     [ProducesResponseType(StatusCodes.Status400BadRequest)]
     [ProducesResponseType(StatusCodes.Status401Unauthorized)]
     [ProducesResponseType(StatusCodes.Status403Forbidden)]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
     [ProducesResponseType(StatusCodes.Status500InternalServerError)]
-    public async Task<ActionResult<Mission>> Create(
+    public async Task<ActionResult<MissionRun>> Create(
         [FromBody] ScheduledMissionQuery scheduledMissionQuery
```
aestene, line 327: Consider deleting this endpoint as it would no longer be used.

backend/api/Controllers/MissionController.cs
```diff
@@ -272,27 +498,48 @@ public async Task<ActionResult<byte[]>> GetMap([FromRoute] string assetCode, str
...
+    [ProducesResponseType(typeof(MissionRun), StatusCodes.Status200OK)]
     [ProducesResponseType(StatusCodes.Status401Unauthorized)]
     [ProducesResponseType(StatusCodes.Status403Forbidden)]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
     [ProducesResponseType(StatusCodes.Status500InternalServerError)]
-    public async Task<ActionResult<Mission>> DeleteMission([FromRoute] string id)
+    public async Task<ActionResult<MissionRun>> DeleteMissionRun([FromRoute] string id)
     {
-        var mission = await _missionService.Delete(id);
+        var mission = await _missionRunService.Delete(id);
```
aestene, line 540:
```suggestion
        var missionRun = await _missionRunService.Delete(id);
```

backend/api/Controllers/Models/RescheduleMissionQuery.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Api.Database.Models;
+
+namespace Api.Controllers.Models
+{
+    public class RescheduleMissionQuery
+    {
+        public MissionSourceType MissionType { get; set; }
```
aestene, line 7: `MissionSourceType` likely not needed as this is given in the `MissionDefinition`

backend/api/Controllers/Models/RescheduleMissionQuery.cs
```diff
@@ -0,0 +1,12 @@
...
+        public string RobotId { get; set; }
+        public DateTimeOffset DesiredStartTime { get; set; }
+        public TimeSpan? InspectionFrequency { get; set; }
```
aestene, line 10: Should these two be a part of this? The frequency should not change every time a mission is scheduled.

We are also not using a desired start time other than now I think so might not be necessary as an input yet.

backend/api/Controllers/MissionController.cs
```diff
@@ -141,20 +216,114 @@ public async Task<ActionResult<byte[]>> GetMap([FromRoute] string assetCode, str
...
+    [HttpPost("{missionId}/reschedule")]
+    [Authorize(Roles = Role.User)]
+    [ProducesResponseType(typeof(MissionRun), StatusCodes.Status201Created)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    public async Task<ActionResult<MissionRun>> Rechedule(
+        [FromRoute] string missionId,
```
aestene, line 233:
```suggestion
        [FromRoute] string missionDefinitionId,
```

backend/api/Controllers/RobotController.cs
```diff
@@ -16,25 +16,25 @@ public class RobotController : ControllerBase
...
     private readonly IRobotModelService _robotModelService;
-    private readonly IAssetDeckService _assetDeckService;
+    private readonly IAreaService _areaService;
 
     public RobotController(
         ILogger<RobotController> logger,
         IRobotService robotService,
         IIsarService isarService,
-        IMissionService missionService,
+        IMissionRunService missionService,
```
aestene, line 27:
```suggestion
        IMissionRunService missionRunService,
```

backend/api/Controllers/RobotController.cs
```diff
@@ -16,25 +16,25 @@ public class RobotController : ControllerBase
...
         IRobotModelService robotModelService,
-        IAssetDeckService assetDeckService
+        IAreaService areaService
     )
     {
         _logger = logger;
         _robotService = robotService;
         _isarService = isarService;
         _missionService = missionService;
```
aestene, line 35:
```suggestion
        _missionRunService = missionRunService;
```

backend/api/Controllers/RobotController.cs
```diff
@@ -726,18 +726,18 @@ public async Task<ActionResult> ResumeMission([FromRoute] string robotId)
...
     [Authorize(Roles = Role.User)]
-    [ProducesResponseType(typeof(Mission), StatusCodes.Status200OK)]
+    [ProducesResponseType(typeof(MissionRun), StatusCodes.Status200OK)]
     [ProducesResponseType(StatusCodes.Status400BadRequest)]
     [ProducesResponseType(StatusCodes.Status401Unauthorized)]
     [ProducesResponseType(StatusCodes.Status403Forbidden)]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
     [ProducesResponseType(StatusCodes.Status500InternalServerError)]
-    public async Task<ActionResult<Mission>> SafePosition(
+    public async Task<ActionResult<MissionRun>> SafePosition(
```
aestene, line 737:
```suggestion
    public async Task<ActionResult<MissionRun>> SendRobotToSafePosition(
```

backend/api/Database/Context/FlotillaDbContext.cs
```diff
@@ -9,8 +9,12 @@ public class FlotillaDbContext : DbContext
 {
     public DbSet<Robot> Robots => Set<Robot>();
     public DbSet<RobotModel> RobotModels => Set<RobotModel>();
-    public DbSet<Mission> Missions => Set<Mission>();
-    public DbSet<AssetDeck> AssetDecks => Set<AssetDeck>();
+    public DbSet<MissionRun> MissionRuns => Set<MissionRun>();
+    public DbSet<MissionDefinition> MissionDefinitions => Set<MissionDefinition>();
```
aestene, line 13: Is there missing some setup in the modelBuilder below for mission definition?

aestene (changes requested):
Look at comments

backend/api/Database/Models/Asset.cs
```diff
@@ -0,0 +1,21 @@
...
+        [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
+        public string Id { get; set; }
+
+        [Required]
+        [MaxLength(200)]
+        public string Name { get; set; }
+
+        [Required]
+        [MaxLength(10)]
+        public string ShortName { get; set; }
```
Afonso-2403, line 19:
```suggestion
        public string AssetCode { get; set; }
```
So that we keep consistency with other places in the code and with Equinor naming system.

flotilla.code-workspace
andchiind: oops, thanks

backend/api/Database/Models/MissionDefinition.cs
andchiind: The only reason I did it this way was because I used the ID when setting the blob URL for the source JSON, and since this URL was inserted into the missionDefinition, it made sense to generate the ID first. I have now changed it so that the source URL is random instead.

backend/api/Controllers/MissionController.cs
andchiind: https://github.com/equinor/flotilla/issues/878

andchiind:
The main issue at the moment is related to storing the LastRun in the MissionDefinition as well as the relevant MissionDefinition in MissionRun. This kind of relationship is straightforward for one-to-one mappings, but in this case even though each MissionDefinition needs to have a unique LastRun, many MissionRuns can have the same MissionDefinition. This kind of mapping is seemingly not straightforward to enable in Entity Framework since if the mapping is not addressed the framework complains that the mapping is not explicit ("The dependent side could not be determined for the one-to-one relationship between 'MissionDefinition.LastRun' and 'MissionRun.MissionDefinition'. To identify the dependent side of the relationship, configure the foreign key property. If these navigations should not be part of the same relationship, configure them independently via separate method chains in 'OnModelCreating'. See http://go.microsoft.com/fwlink/?LinkId=724062 for more details."). If we on the other hand do specify a one-to-one mapping then the framework requires the foreign key to be unique, which we cannot guarantee. Even if the MissionRun ID is used as the foreign key then the framework automatically assigns duplicate IDs to MissionRuns which map to the same MissionDefinition.

For now I will just store the MissionId in the MissionRun instead of linking to the MissionDefinition explicitly.

backend/api/Controllers/Models/RescheduleMissionQuery.cs
andchiind: Since we are already using DesiredStartTime in the main branch I will keep it in even if it does not have an effect, so that we don't need to worry about readding it later, but I will remove InspectionFrequency.

backend/api/Controllers/MissionController.cs
andchiind: The simplest option would be to check if there exists a mission definition that uses the echo mission which the user is trying to schedule. In that case we can then reuse that mission definition. The problem arises when they want to use a different inspection frequency, or if they want to change it in other ways. An alternative which deals with this is to keep the MissionDefinitions separate from the echo missions, and expect users to first check MissionDefinitions for the echo mission they want before getting a new one from the echo API.

backend/api/Database/Models/MissionRun.cs
andchiind: I also think it would be nice to include MissionDefinition here. See this comment for why I have not yet implemented it: https://github.com/equinor/flotilla/pull/864#issuecomment-1621340083

backend/api/Services/MissionDefinitionService.cs
andchiind: Since the MissionDefinitionService filter is so much simpler than the MissionRun one, I think for now it is easier and more readable to keep them separate, but that is just my subjective opinion.

backend/api/Services/SourceService.cs
andchiind: I will implement this change last, as it requires rebasing the branch from main

andchiind:
Marked as draft due to this migration error: 
image
It seems like a common error: https://github.com/dotnet/efcore/issues/27504 and https://github.com/dotnet/efcore/issues/26405 both refer to a variety of scenarios where it happens.

Feel free to still review the PR in the meantime though :)

backend/api/Services/AreaService.cs
```diff
@@ -0,0 +1,215 @@
...
+                return null;
+
+            return await _context.Areas.Where(a =>
+                a.Name.ToLower().Equals(areaName.ToLower()) &&
+                a.Installation.Id.Equals(installation.Id)
+            ).Include(a => a.SafePositions).Include(a => a.Installation)
+                .Include(a => a.Plant).Include(a => a.Deck).FirstOrDefaultAsync();
+        }
+
+        public async Task<Area?> ReadByInstallationAndName(string installationCode, string areaName)
```
mrica-equinor, line 86: Both this and the previous have the same name ReadByInstallationAndName, I think this one should be ReadByInstallationCodeAndName, no?

backend/api/Controllers/AreaController.cs
```diff
@@ -0,0 +1,302 @@
...
+        [Authorize(Roles = Role.Admin)]
+        [Route("{installationCode}/{plantCode}/{deckName}/{areaName}/safe-position")]
+        [ProducesResponseType(typeof(AreaResponse), StatusCodes.Status201Created)]
+        [ProducesResponseType(StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+        public async Task<ActionResult<AreaResponse>> AddSafePosition(
+            [FromRoute] string installationCode,
+            [FromRoute] string plantCode,
```
mrica-equinor, line 102: You don't really use plantCode and deckName except for the logger and the exception, do we really need them as inputs, then?

backend/api/Controllers/PlantController.cs
```diff
@@ -0,0 +1,155 @@
...
+            {
+                var existingInstallation = await _installationService.ReadByName(plant.InstallationCode);
+                if (existingInstallation == null)
+                {
+                    return NotFound($"Installation with installation code {plant.InstallationCode} not found");
+                }
+                var existingPlant = await _plantService.ReadByInstallationAndName(existingInstallation, plant.PlantCode);
+                if (existingPlant != null)
+                {
+                    _logger.LogInformation("An plant for given name and plant already exists");
```
mrica-equinor, line 114: A plant instead of an plant (I know it's just a detail ahah)

backend/api/Services/AreaService.cs
```diff
@@ -0,0 +1,215 @@
...
+                return new List<Area>();
+
+            return await _context.Areas.Where(a =>
+                a.Installation.Id.Equals(installation.Id)).Include(a => a.SafePositions).Include(a => a.Installation)
+                .Include(a => a.Plant).Include(a => a.Deck).ToListAsync();
+        }
+
+        public async Task<Area?> ReadByInstallationAndPlantAndDeckAndName(Installation? installation, Plant? plant, Deck? deck, string areaName)
+        {
+            if (installation == null || plant == null || deck == null)
```
mrica-equinor, line 112: If they can't be null, why make them optional inputs in the first place?

backend/api/Services/DeckService.cs
```diff
@@ -0,0 +1,144 @@
...
+            _installationService = installationService;
+            _plantService = plantService;
+        }
+
+        public async Task<IEnumerable<Deck>> ReadAll()
+        {
+            return await GetDecks().ToListAsync();
+        }
+
+        private IQueryable<Deck> GetDecks()
```
mrica-equinor, line 57: Does this crash if there are no decks? Since we don't have the "?" after Deck

andchiind: Since it's returning the query instead of the object itself I don't think it should be null. But I could be wrong.

mrica-equinor: You should know more than me for sure ahah thanks!

backend/api/Services/DeckService.cs
```diff
@@ -0,0 +1,144 @@
...
+                };
+                await _context.Decks.AddAsync(deck);
+                await _context.SaveChangesAsync();
+            }
+            return deck!;
+        }
+
+        public async Task<Deck> Update(Deck deck)
+        {
+            var entry = _context.Update(deck);
```
mrica-equinor, line 124: We don't make any checks when updating the deck or the area, are we assuming context handles it if the deck passed does not exist? Or does the code never call this unless it exists already?

andchiind: The update method is never used, but I suppose for now we can assume the praxis would be to try catch this method. Alternatively we could try catch in here, but we would still end up returning a new exception. The documentation indicates that the object will be added if it does not exist, but I don't always trust the documentation to mention all the possible exceptions https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.dbcontext.update?view=efcore-7.0

backend/api/Controllers/MissionController.cs
```diff
@@ -96,19 +152,38 @@ IStidService stidService
...
+    [ProducesResponseType(typeof(MissionRun), StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    public async Task<ActionResult<MissionDefinition>> GetMissionDefinitionById([FromRoute] string id)
+    {
+        var mission = await _missionDefinitionService.ReadById(id);
         if (mission == null)
             return NotFound($"Could not find mission with id {id}");
```
OleDrange: Should be missionDefinition:
`return NotFound($"Could not find missionDefinition with id {id}");`

andchiind:
Merged branch in another PR

PR #873, opened 2023-07-03
Fix when to show move arm buttons

OleDrange:


frontend/src/components/Pages/RobotPage/RobotArmMovement.tsx
```diff
@@ -1,23 +1,42 @@
...
     robot: Robot
     armPosition: string
+    buttonActivated?: boolean
 }
 const feedbackTimer = 10000 // Clear feedback after 10 seconds
 
-export function MoveRobotArm({ robot, armPosition }: RobotProps) {
+export function MoveRobotArm({ robot, armPosition, buttonActivated }: RobotProps) {
     const [feedback, setFeedback] = useState('')
+    const [Usable, setUsable] = useState(!!buttonActivated)
```
Eddasol, line 16: Lower letter in usable

frontend/src/components/Pages/RobotPage/RobotArmMovement.tsx
```diff
@@ -1,23 +1,42 @@
...
+
+    useEffect(() => {
+        if (typeof buttonActivated !== 'undefined') {
+            setUsable(buttonActivated)
+        }
+    }, [buttonActivated])
+
+    const handleNotUsable = () => {
```
Eddasol, line 24: suggest renaming to "armButtonStyle"

Eddasol:
The user should get some feedback when the button is disabled. E.g. a popover or popup dialog if clicking the disabled button

frontend/src/components/Pages/RobotPage/RobotArmMovement.tsx
```diff
@@ -1,23 +1,42 @@
...
-import { useState } from 'react'
+import { useState, useEffect } from 'react'
 import { BackendAPICaller } from 'api/ApiCaller'
 import { Robot } from 'models/Robot'
 import { TranslateText } from 'components/Contexts/LanguageContext'
 
 interface RobotProps {
     robot: Robot
     armPosition: string
+    buttonActivated?: boolean
```
Eddasol, line 10: Do we need this to be optional? If we don't parse a value for buttonActivated the component will always be disabled and not usable anyways

frontend/src/components/Pages/RobotPage/RobotPage.tsx
```diff
@@ -1,4 +1,4 @@
-import { Typography } from '@equinor/eds-core-react'
+import { Button, Typography } from '@equinor/eds-core-react'
```
Eddasol, line 1: Button is not used

frontend/src/components/Pages/RobotPage/RobotArmMovement.tsx
Eddasol, line 10: buttonActivated can be interpreted as button has been clicked/activated. Consider different name.

Possibly name isRobotAvailable?

frontend/src/components/Pages/RobotPage/RobotArmMovement.tsx
```diff
@@ -1,23 +1,42 @@
...
 }
 const feedbackTimer = 10000 // Clear feedback after 10 seconds
 
-export function MoveRobotArm({ robot, armPosition }: RobotProps) {
+export function MoveRobotArm({ robot, armPosition, buttonActivated }: RobotProps) {
     const [feedback, setFeedback] = useState('')
+    const [Usable, setUsable] = useState(!!buttonActivated)
+
+    useEffect(() => {
+        if (typeof buttonActivated !== 'undefined') {
```
Eddasol, line 19:
```suggestion
        if (buttonActivated) {
```

OleDrange:
> The user should get some feedback when the button is disabled. E.g. a popover or popup dialog if clicking the disabled button

I feel we don't need that. The button goes gray and the mouse pointer shows it is not clickable. Personally, I would be annoyed by a popup for that. Maybe we should get a third opinion? Or some other graphical change on the button itself?

Eddasol:
> I feel we don't need that. The button goes gray and the mouse pointer shows it is not clickable. Personally, I would be annoyed by a popup for that. Maybe we should get a third opinion? Or some other graphical change on the button itself?

Can be popup only for click, and not for hover. This way it's only visible for the confused users. 

Also, most will understand that it's disabled but maybe not why and how to fix it

frontend/src/components/Pages/RobotPage/RobotArmMovement.tsx
```diff
@@ -35,7 +55,9 @@ export function MoveRobotArm({ robot, armPosition }: RobotProps) {
     }
     return (
         <>
-            <Button onClick={onClickMoveArm}>{TranslateText('Set robot arm to ') + '"' + armPosition + '"'}</Button>
+            <Button style={moveArmButtonStyle()} disabled={!usable} onClick={onClickMoveArm}>
```
Eddasol, line 58:
```suggestion
            <Button disabled={!usable} onClick={onClickMoveArm}>
```
No need with custom disabled style. Happens automatically when disabled

Eddasol:
The button doesn't update to disabled if the robot becomes busy while at robotPage

Can test this by starting a localization mission from the robotPage

OleDrange:
> The button doesn't update to disabled if the robot becomes busy while at robotPage
> 
> Can test this by starting a localization mission from the robotPage

Fixed

OleDrange:
> > I feel we don't need that. The button goes gray and the mouse pointer shows it is not clickable. Personally, I would be annoyed by a popup for that. Maybe we should get a third opinion? Or some other graphical change on the button itself?
> 
> Can be popup only for click, and not for hover. This way it's only visible for the confused users.
> 
> Also, most will understand that it's disabled but maybe not why and how to fix it

Added a popover

frontend/src/components/Pages/RobotPage/RobotArmMovement.tsx
Eddasol: If we don't use the built in disable I would use this style: 
"
style = {{
    backgroundColor: tokens.colors.interactive.disabled__fill.hex,
    color: tokens.colors.interactive.disabled__text.hex,
    borderColor: tokens.colors.interactive.disabled__border.hex,
    cursor: "not-allowed",
}}"

to make sure that the style is the same as other disabled buttons

OleDrange: done

andchiind (approved):
LGTM

Issue #875, opened 2023-07-03
Add Filter and better navigation to Mission History


OleDrange:
**Describe the improvement you would like to see**
Add Filter and better navigation to Mission History. Now it looks like the image below. Should be able to navigate quickly to page 30, or any other page. Now this takes 20 clicks at least. 
Also, it would be good to go back to the same place after pressing back in a mission. Now this will return you to page 1 even if you clicked the mission on page 30. 
![image](https://github.com/equinor/flotilla/assets/42737657/d13439a2-fd79-4b23-9af3-77509ac3a848)

**How will this change existing functionality?**
Possible to find missions easier in the mission history. 

**How will this improvement affect the current Threat Model?**

aestene:
This issue has automatically been marked as stale as there has been no activity for 60 days.

anetteu:
Why is this issue archived but not closed? I think this is implemented? Restoring issue and moving to new to force an evalution during planning.

oysand:
This should be closed

PR #884, opened 2023-07-10
Filtering possibility in mission history

Eddasol:


Eddasol:
Question: Right now you stay on the same page when adding/changing filtering. Is this desired or would it be better to always get sent back to page 1 on filter change?

Eddasol:
Displaying of active filters is not included as we're waiting for design

mrica-equinor:
> Question: Right now you stay on the same page when adding/changing filtering. Is this desired or would it be better to always get sent back to page 1 on filter change?

I think for now we can leave it as is and then later get feedback from the operators

frontend/src/components/Pages/FrontPage/MissionOverview/OngoingMissionView.tsx
```diff
@@ -29,45 +29,30 @@ const ButtonStyle = styled.div`
...
-    useEffect(() => {
-        const missions: Mission[] = ongoingMissions.concat(pausedMissions)
-        setMissionsToDisplay(missions)
-    }, [ongoingMissions, pausedMissions])
+    }, [refreshInterval, updateOngoingAndPausedMissions])
 
-    var missionDisplay = missionsToDisplay.map(function (mission, index) {
+    var missionDisplay = missions.map(function (mission, index) {
         return <OngoingMissionCard key={index} mission={mission} />
     })
```
andchiind, line 57:
```suggestion
    var missionDisplay = missions.map((mission, index) =>
        <OngoingMissionCard key={index} mission={mission} />
    )
```

andchiind: Lambda functions look nice and help promote a functional style in my opinion🙂

frontend/src/components/Pages/MissionHistoryPage/FilterSection.tsx
```diff
@@ -0,0 +1,203 @@
...
+    const onRobotNameChange = (changes: ChangeEvent<HTMLInputElement>) => {
+        switchRobotName(changes.target.value)
+    }
+
+    const onTagIdChange = (changes: ChangeEvent<HTMLInputElement>) => {
+        switchTagId(changes.target.value)
+    }
+
+    const onMinStartTimeChange = (changes: ChangeEvent<HTMLInputElement>) => {
+        setFormatedMinStartTime(changes.target.value)
```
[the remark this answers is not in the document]
andchiind: The same applies for other "Formated" instances

frontend/src/components/Pages/MissionHistoryPage/FilterSection.tsx
```diff
@@ -0,0 +1,203 @@
...
+        switchInspectionTypes,
+        switchMinStartTime,
+        switchMaxStartTime,
+        switchMinEndTime,
+        switchMaxEndTime,
+        resetFilters,
+    } = useMissionFilterContext()
+
+    useEffect(() => {
+        if (formatedMinStartTime) switchMinStartTime(new Date(formatedMinStartTime).getTime() / 1000)
```
andchiind, line 59: Couldn't this check be done inside the filter context? Then we could avoid the extra state variables here. I might be misunderstanding

frontend/src/components/Pages/MissionHistoryPage/MissionHistoryView.tsx
```diff
@@ -23,41 +27,75 @@ const StyledLoading = styled.div`
...
+            if (page > paginatedMissions.pagination.TotalPages && paginatedMissions.pagination.TotalPages > 0) {
+                switchPage(paginatedMissions.pagination.TotalPages)
+                setIsResettingPage(true)
             }
-        )
-    }, [completedStatuses, currentPage, pageSize])
+            setIsLoading(false)
+        })
+    }, [
+        page,
```
andchiind, line 73: If you do end up switching to having a single state object in the context then you would only need to check for an update to that object, since here we don't care which specific value is updated

frontend/src/components/Pages/MissionHistoryPage/FilterSection.tsx
Eddasol: I changed the type to be optional number so the check is no longer necessary

frontend/src/components/Pages/MissionHistoryPage/FilterSection.tsx
```diff
@@ -0,0 +1,207 @@
...
+                        onChange={onRobotNameChange}
+                    />
+                    <Search
+                        value={tagId ?? ''}
+                        placeholder={TranslateText('Search for a tag')}
+                        onChange={onTagIdChange}
+                    />
+                    <TextField
+                        id="datetime"
+                        value={formattedMinStartTime}
```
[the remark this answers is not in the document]
Eddasol: I tried that first but without success. Need a specific format and since this uses built in date time pickers the format will vary with chosen browser. Didn't find a different alternative but if you have any ideas I would love to try

frontend/src/language/en.json
```diff
@@ -92,5 +92,16 @@
...
     "Stop button pressed confirmation text": "Are you sure you want to stop the mission?",
-    "Stop mission": "Stop mission"
+    "Stop mission": "Stop mission",
+    "Search for missions": "Search for missions",
+    "Search for a tag": "Search for a tag",
+    "Search for a robot name": "Search for a robot name",
+    "Filter": "Filter",
+    "Clear all filters": "Clear all filters",
+    "Mission status": "Mission status",
+    "Inspection type": "Inspeksjonstype",
```
mrica-equinor, line 102: `"Inspection type": "Inspection type",`

mrica-equinor (approved):
LGTM :)

mrica-equinor:
Closes #875

PR #894, opened 2023-07-19
Store error strings from isar

### LINKED ISSUE #871: Store error reasons for mission, task and step
After the implementation of https://github.com/equinor/isar/pull/457 ISAR will now include an error description with the MQTT messages that are sent when the status for a mission, task or step is updated. The MQTT message contains two fields
1. `ErrorDescription` (This contains a log message with some detailed information about what went wrong)
2. `ErrorReason` (This points to the type of exception which was raised in ISAR to cause this error)

Flotilla should
- Store the `ErrorDescription` and `ErrorReason` as fields on the respective models for mission, task and step when an MQTT message is collected (see here https://github.com/equinor/flotilla/blob/main/backend/api/EventHandlers/MqttEventHandler.cs#L350-L386 for how we currently update the status of a task)
- Show the user a valuable error message based on the `ErrorReason` 

It might be that the `ErrorDescription` should just be stored on the mission while the `ErrorReason` is used as an error code which translates into a default message for that type of error. For example
- The `ErrorReason` is "robot_communication_exception"
- The `ErrorDescription` is "Failed to schedule step {id} because: Unable to communicate with robot"
- The user facing description for `ErrorReason` "robot_communication_exception" could be " The step failed as the system was not able to communicate with the robot"

Note that the frontend part of this issue is covered by #757 

The end result of this issue should be that the information from ISAR messages are stored on the mission, task and step models such that #757 may implement the user facing error feedback.

- Should constitute a great improvement in error feedback to users which again will enable us to quicker identify issues which might also affect the threat model.

andchiind:
Closes #871

### DIFF
--- a/backend/api/Controllers/MissionController.cs
+++ b/backend/api/Controllers/MissionController.cs
@@ -54,12 +54,12 @@ public class MissionController : ControllerBase
     /// </remarks>
     [HttpGet("runs")]
     [Authorize(Roles = Role.Any)]
-    [ProducesResponseType(typeof(IList<MissionRun>), StatusCodes.Status200OK)]
+    [ProducesResponseType(typeof(IList<MissionRunDto>), StatusCodes.Status200OK)]
     [ProducesResponseType(StatusCodes.Status400BadRequest)]
     [ProducesResponseType(StatusCodes.Status401Unauthorized)]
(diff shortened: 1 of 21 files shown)


andchiind:
Marked as draft as I am uncertain of how to trigger these errors in isar.

aestene:

PR #895, opened 2023-07-19
Filter frontend assets by active robots

### LINKED ISSUE #495: Only show the assets that the user has access to (and that are set up with a robot)
Screenshot from 2023-01-09 12-21-22

This should be resolved by a backend call which only returns assets which has active robots. 

In the upper drop-down list, only show the assets that the user has access to (and that are set up with a robot)

andchiind:
Closes #495

### DIFF
--- a/backend/api/Controllers/EchoController.cs
+++ b/backend/api/Controllers/EchoController.cs
@@ -1,7 +1,9 @@
-﻿using System.Text.Json;
+﻿using System.Globalization;
+using System.Text.Json;
 using Api.Controllers.Models;
 using Api.Services;
 using Api.Services.Models;
+using Api.Utilities;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 
@@ -16,10 +18,13 @@ public class EchoController : ControllerBase
 
     private readonly IEchoService _echoService;
 
-    public EchoController(ILogger<EchoController> logger, IEchoService echoService)
+    private readonly IRobotService _robotService;
+
+    public EchoController(ILogger<EchoController> logger, IEchoService echoService, IRobotService robotService)
     {
         _logger = logger;
         _echoService = echoService;
+        _robotService = robotService;
     }
 
     /// <summary>
@@ -152,6 +157,51 @@ public class EchoController : ControllerBase
         try
         {
             var echoPlantInfos = await _echoService.GetEchoPlantInfos();
+            return Ok(echoPlantInfos);
+        }
+        catch (HttpRequestException e)
+        {
+            _logger.LogError(e, "Error getting plant info from Echo");
+            return new StatusCodeResult(StatusCodes.Status502BadGateway);
+        }
+        catch (JsonException e)
+        {
+            _logger.LogError(e, "Error deserializing plant info response from Echo");
+            return new StatusCodeResult(StatusCodes.Status500InternalServerError);
+        }
+    }
+
+    /// <summary>
+    /// Get all plants associated with an active robot.
+    /// </summary>
+    /// <remarks>
+    /// <para> Retrieves the plants that have an active robot </para>
+    /// </remarks>
+    [HttpGet]
+    [Authorize(Roles = Role.User)]
+    [Route("active-plants")]
+    [ProducesResponseType(typeof(IList<EchoPlantInfo>), StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    public async Task<ActionResult<IList<EchoPlantInfo>>> GetActivePlants()
+    {
+        var plants = await _robotService.ReadAllActivePlants();
+
+        if (plants == null)
+        {
+            _logger.LogWarning("Could not retrieve robot plants information");
+            throw new RobotInformationNotAvailableException("Could not retrieve robot plants information");
+        }
+
+        plants = plants.Select(p => p.ToLower(CultureInfo.CurrentCulture));
+
+        try
+        {
+            var echoPlantInfos = await _echoService.GetEchoPlantInfos();
+
+            echoPlantInfos = echoPlantInfos.Where(p => plants.Contains(p.PlantCode.ToLower(CultureInfo.CurrentCulture))).ToList();
             return Ok(echoPlantInfos);
         }
         catch (HttpRequestException e)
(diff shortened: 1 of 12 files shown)


andchiind:
I'm currently waiting on advice as to how to use the Button.Toggle component as it currently gives an error when used.

andchiind:
> I'm currently waiting on advice as to how to use the Button.Toggle component as it currently gives an error when used.

For now I think I'll just use a different component even though the Button.Toggle would be ideal in my opinion

frontend/src/components/Header/Header.tsx
```diff
@@ -74,26 +74,36 @@ function InstallationPicker(page: string) {
...
+                disabled={page === 'mission'}
+                initialSelectedOptions={[installationCode.toUpperCase()]}
+                placeholder={TranslateText('Select installation')}
+                onOptionsChange={({ selectedItems }) => {
+                    const mapKey = mappedOptions.get(selectedItems[0])
+                    if (mapKey !== undefined) switchInstallation(mapKey)
+                    else switchInstallation('')
+                }}
+            />
+            <Checkbox
```
mrica-equinor, line 101: The flotilla logo, autocomplete and checkbox are not aligned to the middle, can you just wrap them in something with an align item to the center? Seems to work good otherwise, it's just that

Eddasol:
Suggest hiding the checkbox into settings. Like a "settings" icon on the header which gives a popup dialog with the checkbox

backend/api/Controllers/RobotController.cs
```diff
@@ -277,6 +280,51 @@ public async Task<ActionResult<IList<VideoStream>>> GetVideoStreams([FromRoute]
...
+    /// </remarks>
+    [HttpGet]
+    [Authorize(Roles = Role.User)]
+    [Route("active-plants")]
+    [ProducesResponseType(typeof(IList<EchoPlantInfo>), StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    public async Task<ActionResult<IList<EchoPlantInfo>>> GetActivePlants()
```
oysand, line 297: Does this make sense to have in the robotService? Would it be better to have this call in the echoService and it can call the robot service to know what active robots there are in order to filter?

backend/api/Controllers/RobotController.cs
```diff
@@ -277,6 +280,51 @@ public async Task<ActionResult<IList<VideoStream>>> GetVideoStreams([FromRoute]
...
+    {
+        var plants = await _robotService.ReadAllActivePlants();
+
+        // TODO: call echo API here to get the remaining info so that we can rely solely on this endpoint
```
oysand, line 301: Todo needs to be addressed, though as the comment above states, should this be swapped around?

backend/api/Controllers/RobotController.cs
andchiind: Are you suggesting we put it in the echoController? I suppose we could also have it in the plantController

andchiind: I moved it to the echo controller

frontend/src/components/Header/Header.tsx
```diff
@@ -74,28 +75,49 @@ function InstallationPicker(page: string) {
...
+
     useEffect(() => {
-        BackendAPICaller.getEchoPlantInfo().then((response: EchoPlantInfo[]) => {
+        const plantPromise = showActivePlants ? BackendAPICaller.getActivePlants() : BackendAPICaller.getEchoPlantInfo()
+        plantPromise.then(async (response: EchoPlantInfo[]) => {
             const mapping = mapInstallationCodeToName(response)
             setAllPlantsMap(mapping)
         })
-    }, [])
+    }, [showActivePlants])
```
mrica-equinor, line 87: My understanding is the list of active plants is updated when the checkbox is ticked or unticked. I was considering updating it also when someone clicks in the autocomplete, with the onFocus parameter, what do you think?

andchiind: Sure I don't see why not

andchiind: It does make the code slightly uglier though since we need to add a dummy variable in the onFocus function which can be placed in the list of variables in the useEffect. Unless we want to reuse showActivePlants for that use, but I'm not sure if that's a clean way of doing it

frontend/src/components/Header/Header.tsx
```diff
@@ -74,28 +75,53 @@ function InstallationPicker(page: string) {
...
+                    if (mapKey !== undefined) switchInstallation(mapKey)
+                    else switchInstallation('')
+                }}
+                autoWidth={true}
+                onFocus={(e) => {
+                    e.preventDefault()
+                    setUpdateListOfActivePlants(!updateListOfActivePlants)
+                }}
+            />
+            <Icon name={Icons.Settings} size={16} aria-haspopup="dialog" onClick={() => setIsDialogOpen(true)} />
```
mrica-equinor, line 110: Icon size maybe change to 24 so that it's a bit bigger

mrica-equinor: Also maybe change the color to the other icons as well 

            <Icon name={Icons.Settings} size={24} aria-haspopup="dialog" color={tokens.colors.interactive.primary__resting.rgba} onClick={() => setIsDialogOpen(true)} />

mrica-equinor: Also, so that we keep the same format as the other buttons in the heather, i suggest changing the code to this:

<Button variant="ghost_icon" onClick={() => setIsDialogOpen(true)} >
                <Icon name={Icons.Settings} size={24} aria-haspopup="dialog" color={tokens.colors.interactive.primary__resting.rgba} />
            </Button>
            
 For consistency and because it's more clear when you hover and see it highlighted

backend/api/Controllers/EchoController.cs
```diff
@@ -165,4 +169,47 @@ public async Task<ActionResult<EchoPlantInfo>> GetEchoPlantInfos()
...
+        {
+            _logger.LogWarning("Could not retrieve robot plants information");
+            throw new RobotInformationNotAvailableException("Could not retrieve robot plants information");
+        }
+
+        try
+        {
+            var echoPlantInfos = await _echoService.GetEchoPlantInfos();
+
+            echoPlantInfos = echoPlantInfos.Where(p => plants.Contains(p.PlantCode)).ToList();
```
mrica-equinor, line 201: I noticed if you write kaa instead of KAA the list does not show the Kårstø installation, so maybe add a ToLower here

frontend/src/components/Header/Header.tsx
mrica-equinor: I would update all the icons in this file to 24 but could be because I'm also using aa small screen, you can decide on that

mrica-equinor (approved):
LGTM :)

PR #910, opened 2023-07-24
Mission run refactor, avoiding github actions issues

andchiind:
New PR created based on https://github.com/equinor/flotilla/pull/864 since the update database script does not recognise that some requested reviews have been withdrawn.


mrica-equinor (approved):
LGTM :)

andchiind:
Reference in case people wonder why there are so many migration attempts: 
image

PR #913, opened 2023-07-26
Use language context instead of unsafe translate

### LINKED ISSUE #904: Error due to the use of useContext in TranslateText
**Describe the bug**
When a call to TranslateText is made, the function calls useContext. This results in a React error. Instead we should use a translate function that's inside the context itself, so that we can safely use the context at the top of the React component which needs the translation. See #903 for an example, where the "translate" function is retrieved from the LanguageContext instead of using TranslateText. 

**To Reproduce**
Call TranslateText inside a conditional statement.

**Expected behavior**
The error should stop occuring as the order of Hooks in the components is constant. Ideally all calls to TranslateText should be replaced with a call to the context function.

andchiind:
Closes #904

### DIFF
--- a/frontend/src/components/Contexts/LanguageContext.tsx
+++ b/frontend/src/components/Contexts/LanguageContext.tsx
@@ -5,7 +5,7 @@ interface ILanguageContext {
     language: string
     textDictionary: { [text: string]: string }
     switchLanguage: (newLanguage: string) => void
-    translate: (str: string) => string
+    TranslateText: (str: string) => string
 }
 
 interface Props {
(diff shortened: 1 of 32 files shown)


frontend/src/components/Contexts/LanguageContext.tsx
```diff
@@ -57,7 +57,7 @@ export const LanguageProvider: FC<Props> = ({ children }) => {
 
 export const useLanguageContext = () => useContext(LanguageContext)
 
-export const TranslateText = (id: string): string => {
+export const TranslateTextUnsafe = (id: string): string => {
```
mrica-equinor, line 60: Don't call it unsafe, this is open source, call is deprecated or old or something

andchiind: I was thinking of calling it unsafe since it can cause frontend exceptions if used incorrectly. I don't think this poses a security risk since it is unrelated to the backend. I'm a little hesitant to calling it deprecated since we do need to use it in a few places where we want to translate outside of React components. It's not necessairly a direct replacement, but it can be good to have the name reflect that it is not the ideal function to use. But as long as this is clear I can change it to something else like "TranslateTextWithContext"

mrica-equinor: I think that's better, change the name and then you can merge

mrica-equinor (approved):

PR #916, opened 2023-07-27
Add endpoint for getting next mission run

### LINKED ISSUE #838: Endpoint for getting the next runs and the most recent run of a mission
An endpoint where the user gets a list of missions which are scheduled and those that have recently been run for a given mission, sorted by date.

The missions models returned should include the mission frequency, so that the UI can calculate the time until the next inspection.

andchiind:
Closes #838

### DIFF
--- a/backend/api.test/EndpointTest.cs
+++ b/backend/api.test/EndpointTest.cs
@@ -58,6 +58,8 @@ namespace Api.Test
                                 services.AddScoped<IIsarService, MockIsarService>();
                                 services.AddScoped<IEchoService, MockEchoService>();
                                 services.AddScoped<IMapService, MockMapService>();
+                                services.AddScoped<IBlobService, MockBlobService>();
+                                services.AddScoped<ICustomMissionService, MockCustomMissionService>();
                                 services.AddAuthorization(
                                     options =>
                                     {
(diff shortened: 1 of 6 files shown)


backend/api/Services/MissionRunService.cs
```diff
@@ -118,6 +120,14 @@ public async Task<PagedList<MissionRun>> ReadAll(MissionRunQueryStringParameters
                 .FirstOrDefaultAsync(missionRun => missionRun.Id.Equals(id));
         }
 
+        public async Task<MissionRun?> ReadNextScheduledRunByMissionId(string missionId)
+        {
+            return await GetMissionRunsWithSubModels()
+                .Where(m => m.MissionId == missionId && m.EndTime == null)
```
Afonso-2403, line 126:
```suggestion
                .Where(m => m.MissionId == missionId && !m.IsCompleted)
```
Would this be more legible?

backend/api/appsettings.Development.json
```diff
@@ -40,6 +40,6 @@
     "AccountName": "flotilladevsa"
   },
   "Database": {
-    "UseInMemoryDatabase": true
+    "UseInMemoryDatabase": false
```
Afonso-2403, line 43: As we discussed this should probably not be changed

backend/api/Services/MissionRunService.cs
andchiind: I agree that it would be more legible, but since IsCompleted is evaluated locally and does not exist in the database, we cannot put it in the LINQ statement. We could however check the state and make sure that it is cancelled, failed, successful or partially successful, but then it becomes less legible again in my opinion.

backend/api.test/EndpointTest.cs
```diff
@@ -179,27 +181,63 @@ private async Task PopulateAreaDb(string installationCode, string plantCode, str
...
             // Act
-            var installationResponse = await _client.PostAsync(installationUrl, installationContent);
-            Assert.NotNull(installationResponse);
-            var plantResponse = await _client.PostAsync(plantUrl, plantContent);
-            Assert.NotNull(plantResponse);
-            var deckResponse = await _client.PostAsync(deckUrl, deckContent);
-            Assert.NotNull(deckResponse);
-            var areaResponse = await _client.PostAsync(areaUrl, areaContent);
-            Assert.NotNull(areaResponse);
+            _ = await _client.PostAsync(installationUrl, installationContent);
```
Afonso-2403, line 184: Why were the `Assert.NotNull` removed?

Afonso-2403 (approved):
Lgtm

backend/api.test/EndpointTest.cs
andchiind: This was mainly done since this Post functionality is tested elsewhere so this populate function is treated as a utility function instead of a test. But a different PR does update this function which does include some more assertions.

PR #924, opened 2023-08-01
Create mission definition when using echo mission

### LINKED ISSUE #919: Running an echo mission should also create a mission definition
If a mission definition for the echo mission does not exist, creating a new mission run from echo should create one. If one does exist then it should be used instead of calling the echo API.

By having more mission definitions the user is able to set inspection plans and reschedule missions.

andchiind:
Close #919


andchiind:
Once this PR is merged we should have everything in place to implement the inspection page.

backend/api/Controllers/MissionController.cs
```diff
@@ -364,31 +365,49 @@ public async Task<ActionResult<byte[]>> GetMap([FromRoute] string installationCo
...
+            };
+        }
+        else
+        {
+            var missionDefinitions = await _missionDefinitionService.ReadMissionDefinitionsBySourceId(source.SourceId);
+            // In theory there could be more than one mission definition that is relevant, but we need to randomly choose one
+            missionDefinition = missionDefinitions.FindAll(m => area == null || (m.Area != null && m.Area.Id == area.Id))[0];
+        }
+
+        if (missionDefinition == null)
```
oysand, line 391: Should use ===

backend/api/Controllers/MissionController.cs
```diff
@@ -384,31 +385,49 @@ public async Task<ActionResult<byte[]>> GetMap([FromRoute] string installationCo
 
         if (area == null)
         {
-            // This is disabled for now as the Area database is not yet populated
+            // This is disabled for now as the Area database is not yet populated. We may want to auto populate it
```
Afonso-2403, line 388: This is outdated after #940 has been merged

backend/api/Controllers/MissionController.cs
```diff
@@ -384,31 +385,49 @@ public async Task<ActionResult<byte[]>> GetMap([FromRoute] string installationCo
...
-            InstallationCode = scheduledMissionQuery.InstallationCode,
-            Area = area
-        };
+            };
+        }
+        else
+        {
+            var missionDefinitions = await _missionDefinitionService.ReadMissionDefinitionsBySourceId(source.SourceId);
+            // In theory there could be more than one mission definition that is relevant, but we need to randomly choose one
+            missionDefinition = missionDefinitions.FindAll(m => area == null || (m.Area != null && m.Area.Id == area.Id))[0];
```
Afonso-2403, line 408: In what scenario would multiple missionDefinitions map to the same echo mission?
If that is the case, randomly choosing one sounds dangerous eheh

Afonso-2403 (commented):
Probably quite outdated with the bug fixes related to area that have been merged to main

Afonso-2403 (commented):
Probably quite outdated with the bug fixes related to area that have been merged to main

backend/api/Services/SourceService.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        private IQueryable<Source> GetSources()
+        {
+            return _context.Sources;
+        }
+
+        public async Task<Source?> ReadBySourceIdAndType(string id, MissionSourceType type)
+        {
+            return await GetSources().Where(s => s.Type == type)
+                .FirstOrDefaultAsync(source => source.SourceId != null && source.SourceId == id);
```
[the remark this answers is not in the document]
andchiind: No, I suppose not

backend/api/Controllers/MissionController.cs
andchiind: Currently we don't imagine any situation where this would happen, since each echo mission is tied to a specific area and so it doesn't make sense to have more than one definition, and the only difference they could have would be comment and inspection frequency. The only way for this to happen would be through manual database editing. As a result it shouldn't be dangerous to choose one at random, especially as this would update the lastRun for the given missionDefinition, preventing confusion where a mission has recently been run but we have not detected an inspection in the required inspection frequency time window. The only way we could prevent the need to assume more than one mission definition per source would be if a mission definition owns a source in the database but that is not currently the case. But even in this case we would still use the same code in the mission controller.

backend/api/Controllers/MissionController.cs
andchiind: Yes I have not yet rebased

backend/api/Controllers/MissionController.cs
andchiind: If we are worried about it though we could just create a new plain mission definition if there are currently none that don't have the optional values not set.

oysand: Currently I don't think there is anything tying an echo mission to a single area, other than we can choose to do so. With the ability to traverse between areas I think we need to assume that it is likely that an echo mission may span many areas.

andchiind: I suppose in this case, we would then only have an issue if more than one mission definition used the same source, whilst also mapping to different areas. I think we can likely safely have a mission definition own a source, since I don't see a scenario where we would need more than one mission definition for the same tasks. In that case we could also just create a new Source object. To simplify the database we can then make a mission definition own a source, but I think this may be outside the scope of this PR. If you disagree though then I can try add this.

backend/api/Services/MissionDefinitionService.cs
```diff
@@ -16,10 +16,14 @@ public interface IMissionDefinitionService
 
         public abstract Task<PagedList<MissionDefinition>> ReadAll(MissionDefinitionQueryStringParameters parameters);
 
+        public abstract Task<List<MissionDefinition>> ReadMissionDefinitionsBySourceId(string sourceId);
```
Afonso-2403, line 19: After the discussion above, shouldn't this be refactored to `ReadMissionDefinitionBySourceId`, returning just one missionDefinition?

backend/api/Controllers/MissionController.cs
```diff
@@ -384,25 +385,44 @@ public async Task<ActionResult<byte[]>> GetMap([FromRoute] string installationCo
...
-            InspectionFrequency = scheduledMissionQuery.InspectionFrequency,
-            InstallationCode = scheduledMissionQuery.InstallationCode,
-            Area = area
-        };
+            };
+        }
+        else
+        {
+            var missionDefinitions = await _missionDefinitionService.ReadMissionDefinitionsBySourceId(source.SourceId);
+            // In theory there could be more than one mission definition that is relevant, but we need to randomly choose one
```
Afonso-2403, line 402: Following the discussion on having a one-to-one mapping between missionDefinition and Source, this can be refactored

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -356,6 +357,9 @@ private async void OnMissionUpdate(object? sender, MqttReceivedArgs mqttArgs)
                     model.Type,
                     model.AverageDurationPerTag
                 );
+
+                if (flotillaMissionRun.MissionId != null)
```
Afonso-2403, line 361: This change is not yet in main

backend/api/Services/SortingService.cs
```diff
@@ -45,7 +45,7 @@ public class SortingService
 
                 if (objectProperty == null)
                     throw new InvalidDataException(
-                        $"Mission has no property '{propertyFromQueryName}' for ordering"
+                        $"{typeof(T).Name} has no property '{propertyFromQueryName}' for ordering"
```
Afonso-2403, line 48: This change is not yet in main

backend/api/Services/MissionDefinitionService.cs
```diff
@@ -96,13 +100,31 @@ public async Task<List<MissionDefinition>> ReadByAreaId(string areaId)
                 m => m.IsDeprecated == false && m.Area != null && m.Area.Id == areaId).ToListAsync();
         }
 
+        public async Task<List<MissionDefinition>> ReadMissionDefinitionsBySourceId(string sourceId)
```
Afonso-2403, line 103: This is already in main with the name ReadBySourceId

backend/api/Services/MissionDefinitionService.cs
Afonso-2403: This is already in main with the name `ReadBySourceId`

backend/api/Services/SourceService.cs
```diff
@@ -0,0 +1,57 @@
+﻿using Api.Database.Context;
```
Afonso-2403, line 1: This file is already in main, no extra functionality is brought in here

Afonso-2403 (changes requested):
Sorry for the confusing comments above, in short rebase from main and solve conflicts in the following manner:

- Accept `SourceService.cs` from main
- Accept `MissionDefinitionService.cs` changes from main
- Take in your changes from `MissionDefinitionQueryStringParameters.cs` (if you want to order by name)

After this, I believe the changes introduced in this PR are just `UpdateLastRun` in `MqttEventHandler.cs` and the general type in `SortingService.cs`

andchiind:
This PR has been made redundant by other PRs

PR #926, opened 2023-08-01
Store plant and installation in frontend

### LINKED ISSUE #876: Should use plantCode when fetching echo missions
Using installation code when fetching missions from echo gives a lot more missions than the once that are relevant. Will e.g. get missons from both Haugsneset and Johan Sverdrup when choosing either one of them. Same with Kårstø and Kalstø. See /api/plantInfo in https://echohubapi.equinor.com/.

Suggest getting echo to change parameter in endpoint "/api/robots/robot-plan" from installation code to plant code and implementing the same change on our side

andchiind:
Closes #876 on our end, but changes need to be implemented in echo. The PR does improve the frontend code regardless though as we now store both the current installation and plant.


andchiind:
Marked as draft since we might as well wait until we have the change in echo

PR #945, opened 2023-08-07
List only robots on selected asset in planner

### LINKED ISSUE #943: Robot dropdown contains all robots, not just robot for given installation
**Describe the bug**
When you have chosen a installation, you will only see robots from that installation on the front page.
When you then want to schedule a mission, and choose a robot that mission should be done by, you get a list of all robots in the database. This list should only show robots on the already chosen installation. If only one robot, it should probably be preselected.

Dropdown-list is marked with a red circle on the attached screenshot.

**Screenshots**
image

andchiind:
Closes #943


mrica-equinor (approved):
Nice cleanup, LGTM :)

PR #947, opened 2023-08-07
Move mission queue to react context

### LINKED ISSUE #942: Place the mission queue in the frontend in a React context
The mission queue and its related state should be placed in a React context.

Not only will this make the queue accessible from elsewhere in the program, but most importantly it would stop the queue from dissapearing and reappearing each time we exit and go back to the front page.

N/A

andchiind:
Closes #942


mrica-equinor (approved):
LGTM :)

PR #949, opened 2023-08-08
Fix robot direction in echo missions

### LINKED ISSUE #948: Direction robot should look is wrong
**Describe the bug**
We are using the information from Echo wrong at the moment. We are using `tiltDegClockwise` when we should be using `robotBodyDirectionDegrees`. As the `robotBodyDirectionDegrees` is the value that is linked to the direction drop-down-list from Echo (see picture 1)

**Screenshots**
image
image

mrica-equinor:
Closes #948


Christdej (approved):
LGTM

PR #959, opened 2023-08-09
Split status and heatbeat topic event handlers

aestene:
Flotilla will now handle the robot status separately from the heartbeat signal. Following https://github.com/equinor/isar/pull/483 ISAR will only publish robot status as an event upon change and not continuously as before.


Afonso-2403 (approved):
Looks good to me, just update from main as it is outdated and test again for any change.
Also the fourth commit and the first commit can be merged into just one.
This should be merged after [the isar changes](https://github.com/equinor/isar/pull/483)

PR #967, opened 2023-08-11
Change plant and installation to required

Eddasol:


Christdej:
Remember that changing columns to required will not migrate automatically if the columns are empty in the database.
Please check and update the dev and staging databases to have values in the now empty columns.

Eddasol:
> Remember that changing columns to required will not migrate automatically if the columns are empty in the database. Please check and update the dev and staging databases to have values in the now empty columns.

Checked and there are no empty columns entries for plant and installation

andchiind (approved):
LGTM

PR #970, opened 2023-08-12
Fix all warnings and add warning check to pipeline

Christdej:
Fix warnings from pipeline
Fix runtime warnings
Increase pageSize on test to ensure the counting of missions captures needed missions.
Make pipeline fail if there are warnings

Dont merge until https://github.com/equinor/flotilla/pull/967 is merged as many things will change then

### DIFF
--- a/backend/api/Controllers/AreaController.cs
+++ b/backend/api/Controllers/AreaController.cs
@@ -63,12 +63,13 @@ namespace Api.Controllers
                     "Succesfully created new area with id '{areaId}'",
                     newArea.Id
                 );
+
                 var response = new AreaResponse
                 {
                     Id = newArea.Id,
-                    DeckName = newArea.Deck.Name,
-                    PlantCode = newArea.Plant.PlantCode,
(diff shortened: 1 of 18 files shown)


backend/api/Controllers/AreaController.cs
```diff
@@ -63,12 +63,13 @@ public async Task<ActionResult<AreaResponse>> Create([FromBody] CreateAreaQuery
...
                 );
+
                 var response = new AreaResponse
                 {
                     Id = newArea.Id,
-                    DeckName = newArea.Deck.Name,
-                    PlantCode = newArea.Plant.PlantCode,
-                    InstallationCode = newArea.Installation.InstallationCode,
+                    DeckName = newArea.Deck?.Name ?? string.Empty,
+                    PlantCode = newArea.Plant?.PlantCode ?? string.Empty,
```
oysand, line 71: #967 Address making the plant and installation required and #970 address the warnings and this fix. Is it better to have the fix here as well or address it through the other two?

andchiind (approved):
LGTM

PR #971, opened 2023-08-13
Fix dropdowns in frontend

### LINKED ISSUE #929: Fix dropdowns in frontend
**Describe the bug**
When we have a dropdown in flotilla and it is rendered, it appears to the left initially, instead of being centered to the dropdown.

**To Reproduce**
Click the add mission button on the front page.

**Expected behavior**

**Screenshots**

Image

mrica-equinor:
Closes #929

### DIFF
--- a/frontend/src/components/Header/Header.tsx
+++ b/frontend/src/components/Header/Header.tsx
@@ -93,6 +93,8 @@ function InstallationPicker(page: string) {
                 if (mapKey !== undefined) switchInstallation(mapKey)
                 else switchInstallation('')
             }}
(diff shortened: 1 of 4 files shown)


andchiind (approved):
Works well 🙂 It's not usually ideal to hardcode pixel length in CSS, but in this case I don't really see an easy alternative, and I would be surprised if the mission dialog does not work on a tablet.

mrica-equinor:
> 

I choose the same dimensions as in the localization dialog and this has been tested on the tablets offshore and seems to look good

PR #973, opened 2023-08-15
Schedule missions by using events

aestene:
This PR removes the old MissionScheduler while-loop and instead schedules missions based on events
- MissionCreatedEvent
- RobotAvailableEvent


andchiind (approved):
I don't see any issues, but it may be good to get a second opinion. Up to you if you want to merge now.

PR #981, opened 2023-08-17
Remove old github action

aestene:
This action is no longer used and may be removed. It has started failing due to an old secret.


andchiind (approved):
LGTM

PR #989, opened 2023-08-30
Add obstacle detection reporting in Flotilla

UsamaEquinorAFK:


backend/api/Database/Models/Robot.cs
```diff
@@ -32,6 +32,8 @@ public class Robot
 
         public float BatteryLevel { get; set; }
 
+        public bool? ObstacleDectected { get; set; }
```
oysand, line 35: Is this a bit specific to add it as a column to the robot table? Is there a more general status object we want to use instead?

oysand (changes requested):
Consider handling obstacle status as part of robot status

UsamaEquinorAFK:
> Consider handling obstacle status as part of robot status

I thought about this I would like to keep the ObstacleStatus as a seperate entity in robot table in Database. It can be used in the future to pin point the obstacle location on the map by changing the message we send from isar-taurob in the future and that would be a very cool functionality to have in the future. Let me know what you think ?? :) 

I agree with sending the robot_status from isar and that can be implemented seperately. I can write up a issue link and work on that it in parallel.

oysand:
> > Consider handling obstacle status as part of robot status
> 
> I thought about this I would like to keep the ObstacleStatus as a seperate entity in robot table in Database. It can be used in the future to pin point the obstacle location on the map by changing the message we send from isar-taurob in the future and that would be a very cool functionality to have in the future. Let me know what you think ?? :)
> 
> I agree with sending the robot_status from isar and that can be implemented seperately. I can write up a issue link and work on that it in parallel.

I think is better to keep the implementation now as simple as possible tailored to the current issue and not have ObstacleStatus as part of the robot table. In general preparing for future implementations can be good, but this would add complexity to the system and we would need to ensure that ObstacleStatus is correctly set with regards to the robot status. 

If this is needed for obstacle location in the map I think we should wait with implementing this until we implement that feature.

Eddasol:
Covered by a different PR

PR #991, opened 2023-08-30
Fix mission definition filter and prevent duplicate mission definitions

### LINKED ISSUE #984: Do not create a new source blob if one with the same hash already exists
We want to store the hash in the source object for custom sources, maybe as the ID value. We can then check if there already exists a source object for the tasks we are trying to upload again.

This will prevent duplicate source objects, as well as duplicate blobs. 

N/A

andchiind:
Closes #984

### DIFF
--- a/backend/api.test/EndpointTest.cs
+++ b/backend/api.test/EndpointTest.cs
@@ -813,7 +813,7 @@ namespace Api.Test
             Assert.True(robots != null);
             var robot = robots[0];
             string robotId = robot.Id;
-            int echoMissionId = 95;
+            int echoMissionId = 1; // Corresponds to mock in EchoServiceMock.cs
 
             var query = new ScheduledMissionQuery
             {
--- a/backend/api/Controllers/MissionController.cs
+++ b/backend/api/Controllers/MissionController.cs
@@ -395,7 +395,7 @@ public class MissionController : ControllerBase
         {
             source = new Source
             {
-                SourceId = $"{scheduledMissionQuery.EchoMissionId}",
+                SourceId = $"{echoMission.Id}",
                 Type = MissionSourceType.Echo
             };
         }
--- a/backend/api.test/Mocks/EchoServiceMock.cs
+++ b/backend/api.test/Mocks/EchoServiceMock.cs
@@ -39,6 +39,7 @@ namespace Api.Test.Mocks
             {
                 Id = 1,
                 Name = "test",
+                InstallationCode = "JSV",
                 URL = new Uri("https://www.I-am-echo-stid-tag-url.com"),
                 Tags = new List<EchoTag>()
             };
--- a/backend/api.test/Mocks/StidServiceMock.cs
+++ b/backend/api.test/Mocks/StidServiceMock.cs
@@ -9,6 +9,7 @@ namespace Api.Services
 
         public async Task<Position> GetTagPosition(string tag, string installationCode)
         {
+            await Task.CompletedTask;
             return new Position(
                 x: 0,
                 y: 0,
--- a/backend/api/Services/MissionDefinitionService.cs
+++ b/backend/api/Services/MissionDefinitionService.cs
@@ -140,6 +140,7 @@ namespace Api.Services
         /// Filters by <see cref="MissionDefinitionQueryStringParameters.InstallationCode"/>
         /// and <see cref="MissionDefinitionQueryStringParameters.Area"/>
         /// and <see cref="MissionDefinitionQueryStringParameters.NameSearch" />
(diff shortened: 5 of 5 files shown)


backend/api/Controllers/MissionController.cs
```diff
@@ -384,14 +385,27 @@ public async Task<ActionResult<byte[]>> GetMap([FromRoute] string installationCo
...
+        var source = await _sourceService.CheckForExistingEchoSource(scheduledMissionQuery.EchoMissionId);
+        MissionDefinition? existingMissionDefinition = null;
+        if (source == null)
         {
-            Id = Guid.NewGuid().ToString(),
-            Source = new Source
+            source = new Source
             {
-                SourceId = $"{echoMission.Id}",
+                SourceId = $"{scheduledMissionQuery.EchoMissionId}",
```
andchiind, line 394: The returned echoMission is a mock when running the unit tests which is why we don't want to reference it here, but scheduledMissionQuery.EchoMissionId and echoMission.Id should always be the same here regardless.

backend/api.test/Mocks/CustomMissionServiceMock.cs
```diff
@@ -29,5 +32,20 @@ public string UploadSource(List<MissionTask> tasks)
             }
             return null;
         }
+
+        public string CalculateHashFromTasks(IList<MissionTask> tasks)
```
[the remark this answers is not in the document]
Afonso-2403: I see that it makes it easier to have it this way instead of then also mocking the retrieving of tasks in `GetMissionTasksFromSourceId`

backend/api.test/EndpointTest.cs
```diff
@@ -702,5 +706,141 @@ public async Task GetNextRun()
             Assert.Equal(missionRun3.MissionId, missionRun.MissionId);
             Assert.True(nextMissionRun.Id == missionRun2.Id);
         }
+
+        [Fact]
+        public async Task ScheduleDuplicateCustomMissionDefinitions()
```
Afonso-2403, line 711: For this test (and others) shouldn't `StidService` also be mocked?

backend/api/Controllers/MissionController.cs
Afonso-2403: I am not a big fan of altering functionality to fit tests, but agree that here there is no problem, this will be brought up in the discussion related to how we structure our tests

andchiind: I'm looking into making the mock handle this regardless.

andchiind: image
This approach lets us avoid this change on line 394, but I'm not sure if it is more elegant.

backend/api/Controllers/SourceController.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        return Ok(sources);
+    }
+
+    /// <summary>
+    /// Lookup a custom source by specified id.
+    /// </summary>
+    [HttpGet]
+    [Authorize(Roles = Role.Any)]
+    [Route("{id}/custom")]
```
Afonso-2403, line 78:
```suggestion
    [Route("custom/{id}")]
```
To be in accordance with the other controllers

backend/api/Services/SourceService.cs
```diff
@@ -0,0 +1,185 @@
...
+        public async Task<Source?> ReadBySourceId(string sourceId)
+        {
+            return await GetSources()
+                .FirstOrDefaultAsync(s => s.SourceId.Equals(sourceId));
+        }
+
+        public async Task<SourceResponse?> ReadByIdAndInstallationWithTasks(string id, string installationCode)
+        {
+            var source = await GetSources()
+                .FirstOrDefaultAsync(a => a.Id.Equals(id));
```
Afonso-2403, line 92:
```suggestion
                .FirstOrDefaultAsync(s => s.Id.Equals(id));
```

backend/api/Services/SourceService.cs
```diff
@@ -0,0 +1,185 @@
...
+                    return new SourceResponse(source, tasks);
+                default:
+                    return null;
+            }
+        }
+
+        public async Task<SourceResponse?> ReadByIdWithTasks(string id)
+        {
+            var source = await GetSources()
+                .FirstOrDefaultAsync(a => a.Id.Equals(id));
```
Afonso-2403, line 117:
```suggestion
                .FirstOrDefaultAsync(s => s.Id.Equals(id));
```

Afonso-2403 (approved):
Don't merge before #970 is merged, then fix the warnings of possible null reference

PR #993, opened 2023-09-01
Only show rerun failed missions when applicable

### LINKED ISSUE #978: Can rerun failed tasks when there are no failed tasks
**Describe the bug**
Its possible to choose rerun failed tasks even if there are no failed tasks. This leads to a mission without any tasks that will fail with bad request.

**To Reproduce**
See screenshots

**Expected behavior**
This option should not be available for missions without failed tasks

**Screenshots**
1. Mission without any failed tasks
image
2. Rerun failed tasks
image
3. Resulting crashed mission
Screenshot from 2023-08-16 08-22-41

andchiind:
Closes #978

### DIFF
--- a/frontend/src/components/Pages/FrontPage/MissionOverview/MissionRestartButton.tsx
+++ b/frontend/src/components/Pages/FrontPage/MissionOverview/MissionRestartButton.tsx
@@ -1,4 +1,4 @@
-import { Mission } from 'models/Mission'
+import { Mission, MissionStatus } from 'models/Mission'
 import { Button, EdsProvider, Icon, Menu, Tooltip } from '@equinor/eds-core-react'
 import { Icons } from 'utils/icons'
 import { tokens } from '@equinor/eds-tokens'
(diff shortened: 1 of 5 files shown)


frontend/src/components/Pages/FrontPage/MissionOverview/MissionRestartButton.tsx
```diff
@@ -77,9 +78,17 @@ export function MissionRestartButton({ mission }: MissionProps) {
...
                     </Menu.Item>
-                    <Menu.Item onClick={() => startReRun(ReRunOptions.ReRunFailed)}>
-                        {TranslateText('Re-run failed and cancelled tasks in the mission')}
-                    </Menu.Item>
+                    {(mission.status === MissionStatus.Failed ||
+                        mission.status === MissionStatus.Aborted ||
+                        mission.status === MissionStatus.Cancelled ||
+                        mission.status === MissionStatus.PartiallySuccessful) &&
+                        mission.tasks.some(
+                            (t) => t.status !== TaskStatus.PartiallySuccessful && t.status !== TaskStatus.Successful
```
Eddasol, line 86: Why check for mission status and not just check if some tasks are not successful?

OleDrange (approved):
LGTM

PR #1003, opened 2023-09-08
Add DefaultLocalizationArea to deck

oysand:


backend/api/Controllers/DeckController.cs
```diff
@@ -141,6 +144,53 @@ public async Task<ActionResult<Deck>> Create([FromBody] CreateDeckQuery deck)
             }
         }
 
+        /// <summary>
+        /// Set default localization are for a deck
```
aestene, line 148:
```suggestion
        /// Set default localization area for a deck
```

backend/api/Controllers/DeckController.cs
```diff
@@ -141,6 +144,53 @@ public async Task<ActionResult<Deck>> Create([FromBody] CreateDeckQuery deck)
...
+            var deck = await _deckService.ReadById(deckId);
+            if (deck == null)
+                return NotFound($"Could not find deck with id {deckId}");
+
+            var area = await _areaService.ReadById(areaId);
+            if (area == null)
+                return NotFound($"Could not find area with id {areaId}");
+
+            if (area.Deck == null)
+                return Forbid($"Area {areaId} is not linked to any deck");
```
aestene, line 176:
```suggestion
                return NotFound($"Area {areaId} is not linked to any deck");
```
Assuming `Forbid` means 403 Forbidde that should not be used here as it indicates that the user does not have access to the requested resource which is not correct.

backend/api/Controllers/DeckController.cs
```diff
@@ -141,6 +144,53 @@ public async Task<ActionResult<Deck>> Create([FromBody] CreateDeckQuery deck)
...
+
+            if (area.Deck.Id != deckId)
+                return Forbid($"Area {areaId} is not linked to deck {deckId}");
```
aestene, line 179:
```suggestion
                return NotFound($"Area {areaId} is not linked to deck {deckId}");
```

aestene: This specific one might be fore of a 409 Conflict.

backend/api/Controllers/DeckController.cs
```diff
@@ -141,6 +144,53 @@ public async Task<ActionResult<Deck>> Create([FromBody] CreateDeckQuery deck)
...
+            try
+            {
+                deck.DefaultLocalizationArea = area;
+                var updatedDeck = await _deckService.Update(deck);
+                return Ok(updatedDeck);
+            }
+            catch (Exception e)
+            {
+                _logger.LogError(e, "Error while setting a default localization area");
+                throw;
```
aestene, line 190: We do not want to throw again after handling this error as the stack trace will then be thrown straight to the user. Instead, consider sending an InternalServerError back with a sentence containing some information.

backend/api/Database/Context/FlotillaDbContext.cs
```diff
@@ -69,11 +69,13 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
             poseBuilder.OwnsOne(pose => pose.Orientation);
         });
-        modelBuilder.Entity<Area>().HasOne(a => a.Deck).WithMany();
+        modelBuilder.Entity<Area>().HasOne(a => a.Deck).WithOne(d => d.DefaultLocalizationArea);
         modelBuilder.Entity<Area>().HasOne(a => a.Installation).WithMany();
         modelBuilder.Entity<Area>().HasOne(a => a.Plant).WithMany();
         modelBuilder.Entity<Deck>().HasOne(d => d.Plant).WithMany();
+
         modelBuilder.Entity<Deck>().HasOne(d => d.Installation).WithMany();
+        modelBuilder.Entity<Deck>().HasOne(d => d.DefaultLocalizationArea).WithOne(a => a.Deck);
```
aestene, line 78: Does `d` here mean `deck` while `a` points to an `DefaultLocalizationArea` and not an `Area`?

aestene: Or I guess `DefaultLocalizationArea` is actually an `Area` object?

aestene (changes requested):
See comments

aestene (approved):

Issue #839, opened 2023-06-06
Visualise the inspection state for each asset deck


andchiind:
**Describe the new feature you would like to see**
The asset decks in the asset page should be colour coded based on how close they are to the next inspection, and whether or not an inspection has already been scheduled.

**Describe the solution you'd like**
The user should ideally also be able to see the latest run for the mission and all the upcoming missions, when clicking on a deck.

**How will this feature affect the current Threat Model?**

Eddasol:
Dependant on #864

aestene:
This issue has automatically been marked as stale as there has been no activity for 60 days.

Issue #944, opened 2023-08-07
Allow users to edit mission definition inspection frequency


andchiind:
**Describe the new feature you would like to see**
When the inspection page is ready, when a specific deck is selected, the user should be able to edit the inspection frequency of the mission definition.

**Describe the solution you'd like**
Since mission definitions are created when running echo or custom missions, the last feature that is needed to complete the inspection page is the ability to state how often a mission should be run. By letting the user set this when inspecting a mission definition in the inspection view the inspection overview is complete. Otherwise we will need another mission definition view where these details can be edited after letting the user see all mission definitions.

**How will this feature affect the current Threat Model?**
N/A

andchiind:
This is technically already implemented in main, but it lacks a proper UX design. Ideally it would also be nice to set it in echo.

andchiind:
The components seem to work well, so the only thing remaining is UX I think.

PR #1004, opened 2023-09-08
Add inspection page feature

mrica-equinor:
Closes https://github.com/equinor/flotilla/issues/839 https://github.com/equinor/flotilla/issues/944


backend/api/Controllers/DeckController.cs
```diff
@@ -81,14 +85,43 @@ public async Task<ActionResult<Deck>> GetDeckById([FromRoute] string id)
...
+        {
+            try
+            {
+                var deck = await _deckService.ReadById(id);
+                if (deck == null)
+                    return NotFound($"Could not find deck with id {id}");
+
+                var missionDefinitions = await _missionDefinitionService.ReadByDeckId(deck.Id);
+                var missionDefinitionResponses = missionDefinitions.FindAll(m => !m.IsDeprecated).Select(m => new CondensedMissionDefinitionResponse(m));
+                return Ok(missionDefinitionResponses);
```
aestene, line 118: Is it correct to have a `CondensedMissionDefinitionResponse` here or should it be a `MissionDefinitionResponse`?

andchiind: The return type here is inconsistent but I think it's still good to have the condensed one since returning a list of the uncondensed ones is currently very slow. Each mission definition requires a blob storage lookup. If the frontend wants a more detailed mission definition they can then make a request for a specific mission definition. In the future I do think though that we may want to move away from storing custom mission definition sources in the blob store and instead store them directly in the SQL for the Source table, as we do with MissionRuns

andchiind: I've at least made them consistent now

backend/api/Services/MissionDefinitionService.cs
```diff
@@ -125,6 +149,40 @@ public async Task<MissionDefinition> Update(MissionDefinition missionDefinition)
...
+                                        .Result;
+                                    return new MissionTask(t, tagPosition);
+                                }
+                            )
+                            .ToList(),
+                    MissionSourceType.Custom =>
+                        await _customMissionService.GetMissionTasksFromSourceId(source.SourceId),
+                    _ =>
+                        throw new MissionSourceTypeException($"Mission type {source.Type} is not accounted for")
+                };
```
aestene, line 177: Can this be extracted into a separate method to improve readability?

backend/api/Services/MissionDefinitionService.cs
```diff
@@ -125,6 +149,40 @@ public async Task<MissionDefinition> Update(MissionDefinition missionDefinition)
...
+                            .Select(
+                                t =>
+                                {
+                                    var tagPosition = _stidService
+                                        .GetTagPosition(t.TagId, installationCode)
+                                        .Result;
+                                    return new MissionTask(t, tagPosition);
+                                }
+                            )
+                            .ToList(),
```
aestene, line 172: Can this be extracted into a separate method to improve readability?

aestene (approved):
LGTM

PR #1014, opened 2023-09-20
Handle exceptions when uploading blob

### LINKED ISSUE #968: Backend crashes when trying to run custom mission
**Describe the bug**
The custom mission container is called custom-missions in Azure.
The custom mission container is called custommission in the appsettings.
This causes the backend to crash as the exception is not handeled.
This needs to be fixed.

**To Reproduce**
Run a custom mission in staging with the wrong container name

**Screenshots**
image
image

andchiind:
Closes #968 

Worth noting that in order to catch exceptions when uploading the blobs, the blob uploading needs to be awaited.

The container name fix is done in the robotics infrastructure, here we just handle the potential exception.

### DIFF
--- a/backend/api.test/Mocks/BlobServiceMock.cs
+++ b/backend/api.test/Mocks/BlobServiceMock.cs
@@ -22,7 +22,7 @@ namespace Api.Test.Mocks
             return pages;
         }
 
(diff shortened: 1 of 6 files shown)


Christdej (approved):
LGTM

PR #1015, opened 2023-09-20
Filter sensor types by distinct values

### LINKED ISSUE #974: Task has two thermal image and two thermal video inspections
Mission with echoMissionId 891 has Thermal image, Thermal Video, Thermal Video and Thermal Image as inspection type for the first task. Should this be possible?

**Describe the bug**
When running the Echo mission with ID 891 it displays as having the following data formats
- Thermal image
- Thermal video
- Thermal video
- Thermal image

The issue is to investigate why this happens and if it should indeed be possible. It may also be that we should just display one of each type even if multiple inspections are a part of that step. 

**To Reproduce**
Execute the echo mission with ID 891 locally using isar-robot. 

**Expected behavior**
Would be preferable if the same types don't turn up multiple times.  

**Screenshots**
Image

andchiind:
Closes #974

### DIFF
--- a/backend/api/Controllers/Models/EchoInspection.cs
+++ b/backend/api/Controllers/Models/EchoInspection.cs
@@ -37,4 +37,26 @@ namespace Api.Controllers.Models
             };
         }
     }
+
+    public class EchoInspectionComparer : IEqualityComparer<EchoInspection>
+    {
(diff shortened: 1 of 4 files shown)


andchiind:
Marked as draft as it is hard to test properly while echo is unavailable

Christdej (approved):
LGTM

PR #1017, opened 2023-09-21
Add new Assetselection start page

### LINKED ISSUE #976: Create asset page before flotilla's current front page
We have talked briefly about creating a page before the frontpage where the asset is selected and therefore can't be changed by miss-clicking the dropdown in the frontpage. Something similar to the echo page where you select the asset would be my suggestion for this. I believe this is a good solution since most of the operators will be using flotilla only for the asset they are operating on.

Creating a sub page with the specific platform in the asset, such as P1, could also be nice, and this is what has been implemented by echo as well. 

Also, like in Echo, we should remember the last facility that user used. Gives a smoother user experience, as 99% of user will only be on one facility.

OleDrange:
Closes #976


oysand (changes requested):
Check with UX design and await updated sketch

OleDrange:
Marianne made a sketch in figma: https://www.figma.com/file/YoZo53WJmjlrjdQmJSdoHX/Flotilla---Sandbox?type=design&node-id=198-208197&mode=design&t=0ITQKE3VpPKj8pSk-0

oysand (approved):
LGTM

PR #1022, opened 2023-09-22
Call echo on behalf of users

aestene:
The application will now call Echo on behalf of the logged in user rather than on behalf of the application

### DIFF
--- a/backend/api/Services/EchoService.cs
+++ b/backend/api/Services/EchoService.cs
@@ -5,17 +5,16 @@ using Api.Services.Models;
 using Api.Utilities;
 using Microsoft.Identity.Web;
 using Microsoft.IdentityModel.Tokens;
-
 namespace Api.Services
 {
     public interface IEchoService
     {
-        public abstract Task<IList<CondensedMissionDefinition>> GetAvailableMissions(string? installationCode);
+        public Task<IList<CondensedMissionDefinition>> GetAvailableMissions(string? installationCode);
 
-        public abstract Task<EchoMission> GetMissionById(int missionId);
+        public Task<EchoMission> GetMissionById(int missionId);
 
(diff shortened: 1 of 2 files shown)


Christdej (approved):
LGTM

backend/api/Services/EchoService.cs
```diff
@@ -5,17 +5,16 @@
 using Api.Utilities;
 using Microsoft.Identity.Web;
 using Microsoft.IdentityModel.Tokens;
-
 namespace Api.Services
 {
     public interface IEchoService
     {
-        public abstract Task<IList<CondensedMissionDefinition>> GetAvailableMissions(string? installationCode);
+        public Task<IList<CondensedMissionDefinition>> GetAvailableMissions(string? installationCode);
```
Afonso-2403, line 12: Why are these no longer abstract methods?

aestene: Simply because the IDE marked the abstract declaration we have for interfaces as redundant and not necessary. The auto-formatter then automatically removes them. If we look at the abstract definition I suspect what might have been reasonably is to mark the interface as abstract but I see no reason to do that for the methods within. Not 100% sure, following my IDE quite a bit on this one but there is atleast no use of the abstract identifier here. 

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/abstract

Afonso-2403: From my understanding of abstract classes and methods and the link you sent for C#, both the interface class and its methods should be abstract, as the interface class is never supposed to be instantiated and classes derived from it must provide an implementation for its methods

aestene: Found this which indicates that the abstract keyword is redundant as the default is that you have to implement methods for an interface anyway 

https://www.jetbrains.com/help/rider/RedundantAbstractModifier.html

PR #1037, opened 2023-10-03
Add default localization pose to deck

oysand:


backend/api/Database/Context/InitDb.cs
```diff
@@ -84,6 +86,7 @@ private static List<Deck> GetDecks()
             Id = Guid.NewGuid().ToString(),
             Plant = plants[0],
             Installation = plants[0].Installation,
+            DefaultLocalizationPose = null,
```
Christdej, line 89: We could set one to something usefull rather than "null", might be useful to have one for testing

Christdej (changes requested):
Can we also make a test, for instance in EndpointTest, that tries to update the default localization pose?

backend/api/Database/Context/InitDb.cs
oysand: Agreed, but thought to keep it as null until we started using it

Christdej (approved):
LGTM

PR #1041, opened 2023-10-04
Connect backend and frontend to signalR

### LINKED ISSUE #1020: Implement signalR for async communication between frontend and API
A signalR connection should be made between backend and frontend, and at least one heartbeat message that is currently used should be replaced by it. From this point we can then continue replacing repeating HTTP requests in the frontend.

This is just for testing for now so other real time messaging frameworks can be used as well.

We should experience improved responsiveness and easier debugging as we will no longer need to send many HTTP requests each second to keep our frontend updated.

SignalR is a proven technology that is integrated in Azure, so the threatmodel shouldn't be affected.

andchiind:
Closes #1020 #1063 

This PR connects the frontend and backend to SignalR, and then replaces polling for missionRuns and failed missions for the mission queue and failed missions banner.

It is important to note that in order to use signalR in a react context, that context provider must be nested within the signalR provider in flotillasite.tsx


andchiind:
The big challenge at the moment is the fact that receiving many events at the same time will cause their event handler calls to overlap, so each call to setState within the event handler will be using outdated state

andchiind:
Currently we have one event that is sent if a mission run is updated in any way. Since this means that an update to a step also implies an update due to a task update, we could potentially create more fine tuned events to avoid unnecessary events in the frontend. So if you are only interested in the status of a mission changing then you could subscribe just to that and ignore step and task updates, for example.

andchiind:
All frontend polling has been removed except the call to get mission runs within the history page filter. This one could be removed by having the GET request only be made when a "mission run updated/created" event is received which would not be filtered away, but since "mission run updated" is called each time a task is completed, this may create be a small improvement overall, for the price of more complicated code. It may still be worthwhile doing at some point though, especially as it would be satisfying to remove the last reference to "refreshInterval".

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -352,9 +352,9 @@ private async void OnMissionUpdate(object? sender, MqttReceivedArgs mqttArgs)
                 if (flotillaMissionRun.MissionId != null)
                 {
                     var missionDefinition = await missionDefinitionService.ReadById(flotillaMissionRun.MissionId);
-                    if (missionDefinition != null)
+                    if (missionDefinition != null && flotillaMissionRun.Status == MissionStatus.Successful)
                     {
-                        missionDefinition.LastRun = flotillaMissionRun;
+                        missionDefinition.LastSuccessfulRun = flotillaMissionRun;
```
aestene, line 357: Should this be moved into a scope service instead?

backend/api/Services/MissionRunService.cs
```diff
@@ -88,8 +90,10 @@ private IQueryable<MissionRun> GetMissionRunsWithSubModels()
 
         public async Task<MissionRun> Create(MissionRun missionRun)
         {
+            missionRun.Id ??= Guid.NewGuid().ToString(); // Useful for signalR messages
```
aestene, line 93: Does this mean we set the ID instead of leaving it to the database to set the ID? Is it done so the frontend can have the ID to make requests with?

aestene: If the signalR event is passed after `SaveChangesAsync` wouldn't `missionRun` contain the ID?

backend/api/Services/MissionRunService.cs
```diff
@@ -352,7 +353,10 @@ MissionStatus missionStatus
             missionRun.Status = missionStatus;
 
             await _context.SaveChangesAsync();
+            await Update(missionRun);
```
aestene, line 356: Why do we perform a `SaveChangesAsync` here and then make a call to `Update` which makes another call to `SaveChangesAsync`?

backend/api/Services/MissionRunService.cs
```diff
@@ -403,6 +407,7 @@ IsarTaskStatus taskStatus
             }
 
             await _context.SaveChangesAsync();
+            await Update(missionRun);
```
aestene, line 410: Same as above

backend/api/Services/MissionRunService.cs
```diff
@@ -449,6 +454,7 @@ IsarStepStatus stepStatus
             inspection.UpdateStatus(stepStatus);
 
             await _context.SaveChangesAsync();
+            await Update(missionRun);
```
aestene, line 457: Same as above

backend/api/Services/RobotService.cs
```diff
@@ -52,6 +54,7 @@ public async Task<Robot> CreateFromQuery(CreateRobotQuery robotQuery)
                 _context.Entry(robotModel).State = EntityState.Unchanged;
                 await _context.Robots.AddAsync(newRobot);
                 await _context.SaveChangesAsync();
+                _ = _signalRService.SendMessageAsync("robot list updated", GetEnabledRobotsWithSubModels());
```
aestene, line 57:
```suggestion
                _ = _signalRService.SendMessageAsync("Robot updated", GetEnabledRobotsWithSubModels());
```

Should the event be named after what has happened? I assume the "robot list updated" indicates for the frontend that its list updated but what has happened is that one robot has changed.

aestene: Should the names of these events be gathered in en enum like structure somewhere if they are important to be precise for the frontend?

backend/api/Services/MissionRunService.cs
andchiind: I tihnk we would have to read it in again from the database to get the generated ID, but I can do that

frontend/src/components/Contexts/MissionListsContext.tsx
```diff
@@ -31,9 +39,77 @@ const fetchMissions = (params: {
...
 }
-export const useMissions = (refreshInterval: number): MissionsResult => {
+export const useMissions = (): MissionsResult => {
     const [ongoingMissions, setOngoingMissions] = useState<Mission[]>([])
     const [missionQueue, setMissionQueue] = useState<Mission[]>([])
+    const { registerEvent, connectionReady } = useSignalRContext()
+
+    useEffect(() => {
+        if (connectionReady) {
+            registerEvent('mission run created', (username: string, message: string) => {
```
aestene, line 49: Would like to see the events be based on enums or similar rather than pure strings.

frontend/src/components/Contexts/MissionListsContext.tsx
```diff
@@ -31,9 +39,77 @@ const fetchMissions = (params: {
...
+                const newMission: Mission = JSON.parse(message)
+                if (missionQueue.find((m) => m.id === newMission.id))
+                    setMissionQueue((oldQueue) => [...oldQueue, newMission])
+                else
+                    setMissionQueue((oldQueue) => {
+                        let missionQueueCopy = [...oldQueue]
+                        missionQueueCopy = upsertList(missionQueueCopy, newMission)
+                        return [...missionQueueCopy]
+                    })
```
aestene, line 58: Is this if statement correct? By the first check it looks like you're checking if the new mission already exists and then instead of upserting you are adding a new version of it at the end of the list?

frontend/src/components/Contexts/MissionListsContext.tsx
```diff
@@ -31,9 +39,77 @@ const fetchMissions = (params: {
...
+                // This conversion translates from the enum as a number to an enum as a string
+                updatedMission.status = Object.values(MissionStatus)[updatedMission.status as unknown as number]
+
+                setMissionQueue((oldQueue) => {
+                    const oldQueueCopy = [...oldQueue]
+                    const existingMissionIndex = oldQueue.findIndex((m) => m.id === updatedMission.id)
+                    if (existingMissionIndex !== -1) {
+                        if (updatedMission.status !== MissionStatus.Pending)
+                            oldQueueCopy.splice(existingMissionIndex, 1)
+                        else oldQueueCopy[existingMissionIndex] = updatedMission
```
aestene, line 71: Can this be extracted into a method to hide the complex logic and rather have a good name for the method indicating what it does?

frontend/src/components/Contexts/MissionListsContext.tsx
```diff
@@ -31,9 +39,77 @@ const fetchMissions = (params: {
...
+                        } else {
+                            // Mission is ongoing and not in the queue
+                            return [...oldQueueCopy, updatedMission]
+                        }
+                    } else {
+                        if (existingMissionIndex !== -1) {
+                            // Mission is not ongoing and in the queue
+                            oldQueueCopy.splice(existingMissionIndex, 1)
+                        }
+                    }
```
aestene, line 94: Can this logic be hidden in a well named function?

frontend/src/components/Contexts/SignalRContext.tsx
```diff
@@ -0,0 +1,105 @@
...
+}
+
+const URL = config.BACKEND_API_SIGNALR_URL
+
+export const SignalRContext = createContext<ISignalRContext>(defaultSignalRInterface)
+
+export const SignalRProvider: FC<Props> = ({ children }) => {
+    const [connection, setConnection] = useState<signalR.HubConnection | undefined>(defaultSignalRInterface.connection)
+    const [connectionReady, setConnectionReady] = useState<boolean>(defaultSignalRInterface.connectionReady)
+    const isAuthenticated = useIsAuthenticated()
```
aestene:
```suggestion
    const isAuthenticated = userIsAuthenticated()
```

Should it be user?

frontend/src/components/Pages/FrontPage/MissionOverview/FailedMissionAlertView.tsx
```diff
@@ -100,27 +102,57 @@ export function FailedMissionAlertView({ refreshInterval }: RefreshProps) {
...
+                installationCode &&
+                (!newFailedMission.installationCode ||
+                    newFailedMission.installationCode.toLocaleLowerCase() !== installationCode.toLocaleLowerCase())
+            )
+                return
+            if (new Date(newFailedMission.endTime!) <= lastDismissTime) return
+            let isDuplicate = recentFailedMissions.filter((m) => m.id === newFailedMission.id).length > 0
+            if (isDuplicate) return
+            setRecentFailedMissions([...recentFailedMissions, newFailedMission])
+        }
```
aestene, line 154:
```suggestion
        if (newFailedMission) {
            goodFunctionNameHere()
        }
```

I find quite a few of these useEffect functions to be difficult to read without using quite a bit of effort to read them. Could it be an option to wrap each on in a sub function for better readability? It would also make the declarations shorter and thus the entire setup part easier to read.

frontend/src/components/Pages/InspectionPage/InspectionSection.tsx
```diff
@@ -67,22 +68,23 @@ export interface OngoingMissionType {
 export const compareInspections = (i1: Inspection, i2: Inspection) => {
     if (!i1.missionDefinition.inspectionFrequency) return 1
     if (!i2.missionDefinition.inspectionFrequency) return -1
-    if (!i1.missionDefinition.lastRun) return 1
-    if (!i2.missionDefinition.lastRun) return -1
+    if (!i1.missionDefinition.lastSuccessfulRun) return 1
+    if (!i2.missionDefinition.lastSuccessfulRun) return -1
```
aestene, line 72: Is it best practise for typescript to use -1 and 1 rather than none or bools?

aestene (changes requested):
See comments

backend/api/Services/RobotService.cs
andchiind: The reason I made it robot list updated is because it is a relative small and finite list. As a result it is cheap to update the whole list each time a change is detected. If we instead reacted to each update, creation and deletion we would require a lot more validation and temporary datastructure on the frontend. So I think this simplifies the code a lot without many obvious downsides (if needed we can always compare the lists to find what changed). The reason I did not do this for missionRuns is that there are too many of them and they are constantly increasing.

frontend/src/components/Contexts/SignalRContext.tsx
andchiind: This is an Msal provided context, we didn't write this

backend/api/EventHandlers/MqttEventHandler.cs
andchiind: Based on our recent discussions regarding the problem of switching between different contexts, I'm not sure this needs to be changed.

frontend/src/components/Pages/InspectionPage/InspectionSection.tsx
andchiind: This is the expected output of the function https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#parameters

frontend/src/components/Pages/InspectionPage/InspectionSection.tsx
```diff
@@ -106,8 +108,39 @@ export function InspectionSection({ refreshInterval }: RefreshProps) {
...
+                            deadline: mDef.lastSuccessfulRun
+                                ? getInspectionDeadline(mDef.inspectionFrequency, mDef.lastSuccessfulRun.endTime!)
+                                : undefined,
+                        }
+                        return {
+                            ...deckMissions,
+                            [relevantDeck]: { ...deckMissions[relevantDeck], inspections: inspections },
+                        }
+                    }
+                    return deckMissions
```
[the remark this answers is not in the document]
andchiind: I've divided it into another function now

frontend/src/components/Pages/FrontPage/MissionOverview/FailedMissionAlertView.tsx
andchiind: I've removed the need for this useEffect

frontend/src/components/Contexts/MissionListsContext.tsx
andchiind: I've split them into new functions, hopefully you find them well named

frontend/src/components/Contexts/MissionListsContext.tsx
andchiind: I think it should be better now

frontend/src/components/Contexts/MissionListsContext.tsx
andchiind: I've fixed it now, good catch

aestene (approved):
LGTM

PR #1042, opened 2023-10-04
Fix bug where robot and robot model is not properly linked

Christdej:
To test, might be worth trying to run flotilla, broker and isar-robot to see if behaviour is expected


Eddasol (approved):
LGTM :)

PR #1045, opened 2023-10-05
Add stop all robots button

mrica-equinor:


backend/api/Controllers/EmergencyActionController.cs
```diff
@@ -0,0 +1,167 @@
...
+
+            return Ok("Request to abort current mission and move all robots back to safe position received");
+        }
+
+        /// <summary>
+        ///     This query will clear the emergency state that is introduced by aborting the current mission and returning to a
+        ///     safe zone. Clearing the emergency state means that mission runs that may be in the robots queue will start."
+        /// </summary>
+        [HttpPost]
+        [Route("{robotId}/{installationCode}/{areaName}/clear-robot-emergency-state")]
```
oysand, line 107: Do you need to supply the areaName and the installationCode in addition to the robotId? Could we use the area field from the robot object instead?

backend/api/Database/Context/FlotillaDbContext.cs
```diff
@@ -70,6 +70,8 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
 
         modelBuilder.Entity<MissionRun>().OwnsOne(m => m.Map).OwnsOne(t => t.TransformationMatrices);
         modelBuilder.Entity<MissionRun>().OwnsOne(m => m.Map).OwnsOne(b => b.Boundary);
+        modelBuilder.Entity<MissionRun>().HasOne(m => m.Area).WithMany();
```
oysand, line 73: Do we need to define this in the FlotillaDbContext as well as the database models?

backend/api/Database/Context/FlotillaDbContext.cs
```diff
@@ -70,6 +70,8 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
+        modelBuilder.Entity<Robot>().HasOne(r => r.CurrentArea).WithMany();
```
oysand, line 74: Do we need to define this in the FlotillaDbContext as well as the database models?

backend/api/Database/Context/InitDb.cs
```diff
@@ -134,6 +134,9 @@ private static List<Area> GetAreas()
             MapMetadata = new MapMetadata(),
             DefaultLocalizationPose = new Pose { },
             SafePositions = new List<SafePosition>()
+            {
```
oysand, line 137: Why is this not needed for the other areas?

oysand: Is it needed?

backend/api/EventHandlers/MissionEventHandler.cs
```diff
@@ -102,111 +108,103 @@ private async void OnRobotAvailable(object? sender, RobotAvailableEventArgs e)
...
 
-            if (MissionRunQueueIsEmpty(MissionRunQueue(robot.Id)))
+            if (MissionScheduling.MissionRunQueueIsEmpty(MissionRunQueue(robot.Id)))
             {
                 _logger.LogInformation("The robot was changed to available but there are no mission runs in the queue to be scheduled");
                 return;
             }
 
-            var missionRun = MissionRunQueue(robot.Id).First(missionRun => missionRun.Robot.Id == robot.Id);
+            var missionRun = MissionRunQueue(robot.Id).FirstOrDefault(missionRun => missionRun.Robot.Id == robot.Id);
```
oysand, line 117: Should this be handled in an else statement for the if statement below? This is only necessary if MissionQueueFrozen is false

mrica-equinor:
In order to test the functionality atm it's necessary to add a safe position to an area and localize the robot in that area before clicking the stop all robots button. This will be addressed in another issue #1046

backend/api/EventHandlers/MissionEventHandler.cs
```diff
@@ -102,111 +108,103 @@ private async void OnRobotAvailable(object? sender, RobotAvailableEventArgs e)
...
-            if (!TheSystemIsAvailableToRunAMission(missionRun.Robot, missionRun).Result)
+            _logger.LogInformation("Triggered EmergencyButtonPressed event for robot ID: {RobotId}", e.RobotId);
+            var robot = await RobotService.ReadById(e.RobotId);
+            if (robot == null)
+            {
+                _logger.LogError("Robot with ID: {RobotId} was not found in the database", e.RobotId);
+                return;
+            }
+
+            var area = await AreaService.ReadById(e.AreaId);
```
oysand, line 152: Should we not use the area information from the robot variable?

backend/api/EventHandlers/MissionEventHandler.cs
```diff
@@ -102,111 +108,103 @@ private async void OnRobotAvailable(object? sender, RobotAvailableEventArgs e)
...
+            if (robot == null)
             {
-                _logger.LogWarning("Mission run {MissionRunId} was not started as the robot {RobotId} is not enabled", missionRun.Id, robot.Id);
-                return false;
+                _logger.LogError("Robot with ID: {RobotId} was not found in the database", e.RobotId);
+                return;
             }
-            if (missionRun.DesiredStartTime > DateTimeOffset.UtcNow)
+
+            var area = await AreaService.ReadById(e.AreaId);
```
oysand, line 194: Should we not use the area information from the robot variable?

backend/api/EventHandlers/MissionScheduling.cs
```diff
@@ -0,0 +1,343 @@
...
+        }
+
+        public async Task UnfreezeMissionRunQueueForRobot(Robot robot)
+        {
+            robot.MissionQueueFrozen = false;
+            await _robotService.Update(robot);
+            _logger.LogInformation("Mission queue for robot {RobotName} with ID {RobotId} was unfrozen", robot.Name, robot.Id);
+        }
+
+        public async Task UnfreezeMissionRunQueueForRobot(string robotId)
```
oysand, line 170: Is this the same function as line 163? If so, one should be removed

backend/api/EventHandlers/MissionScheduling.cs
```diff
@@ -0,0 +1,343 @@
...
+                throw;
+            }
+            catch (JsonException e)
+            {
+                string message = "Error while processing the response from ISAR";
+                _logger.LogError(e, "{Message}", message);
+                throw new MissionException(message, 0);
+            }
+
+            if (robot.CurrentMissionId != null)
```
oysand, line 218: Could we reverse these if statements to guard statements to avoid nested if statements?

backend/api/EventHandlers/MissionScheduling.cs
```diff
@@ -0,0 +1,343 @@
...
+                        "Mission '{Id}' failed because ISAR could not be reached",
+                        missionRun.Id
+                    );
+                }
+            }
+            robot.CurrentMissionId = null;
+            await _robotService.Update(robot);
+        }
+
+        public static Pose ClosestSafePosition(Pose robotPose, IList<SafePosition> safePositions)
```
oysand, line 314: We have quite a few supporting functions in this file. Should we split out a utility file for these type of functions?

backend/api/Services/EmergencyActionService.cs
```diff
@@ -0,0 +1,42 @@
...
+        public void TriggerEmergencyButtonDepressedForRobot(EmergencyButtonPressedForRobotEventArgs e)
+        {
+            OnEmergencyButtonDepressedForRobot(e);
+        }
+
+        public static event EventHandler<EmergencyButtonPressedForRobotEventArgs>? EmergencyButtonPressedForRobot;
+
+        protected virtual void OnEmergencyButtonPressedForRobot(EmergencyButtonPressedForRobotEventArgs e)
+        {
+            EmergencyButtonPressedForRobot?.Invoke(this, e);
```
oysand, line 32: Does the ? do more than suppress the type warning for the possible null reference? Is this guaranteed or should we do a check to verify it?

frontend/src/api/ApiCaller.tsx
```diff
@@ -472,4 +472,26 @@ export class BackendAPICaller {
         })
         return result.content
     }
+
+    static async postSafePosition(installationCode: string) {
```
oysand, line 476: Does this post a safe position since it calls the abort-current-missions-and-send-all-robots-to-safe-zone endpoint?

frontend/src/api/ApiCaller.tsx
```diff
@@ -472,4 +472,26 @@ export class BackendAPICaller {
...
+        const body = {}
+
+        const result = await this.POST<unknown, unknown>(path, body).catch((e) => {
+            console.error(`Failed to POST /${path}: ` + e)
+            throw e
+        })
+        return result.content
+    }
+
+    static async resetRobotState(installationCode: string) {
```
oysand, line 487: Would a more descriptive name be clearEmergencyState?

oysand (changes requested):
See comments

backend/api/Controllers/EmergencyActionController.cs
mrica-equinor: At the moment we are not using this endpoint. This endpoint sends only clears one robot from the safe zone. We are using the one below, where all robots are cleared from the safe zone and that one only gets the installation code to clear all robots in that installation. I kept the code because it might be useful in the future if we want to only clear one area or one deck for example, but can also remove it if you think that is better to avoid confusion

backend/api/Database/Context/FlotillaDbContext.cs
mrica-equinor: I didn't do any changes on this, my guess is that Arnt Erik's editor formatted this line and that's why this has been marked as changed (this probably happened in other areas of the code) @aestene

backend/api/Database/Context/InitDb.cs
mrica-equinor: Yeah, ideally we need the areas to have a safe zone for the functionality to work. This area is the one we have been using to test the functionality so we added a safe zone in the initDb to make the process easier

backend/api/EventHandlers/MissionEventHandler.cs
mrica-equinor: Don't have a strong opinion on this, both methods should work

backend/api/EventHandlers/MissionScheduling.cs
mrica-equinor: We can't change the ones that update the database to another file because that caused us to run into issues before, so it would be creating a file for only the last two functions, I believe. This file was originally much smaller but we had to change it. I can try to move those two and as many as possible from the other tho, if you think it's worth it

backend/api/Services/EmergencyActionService.cs
mrica-equinor: @aestene did the work on events, i'm not entirely sure about this

frontend/src/api/ApiCaller.tsx
mrica-equinor: No, will change the name

frontend/src/api/ApiCaller.tsx
mrica-equinor: Absolutely, will change

backend/api/EventHandlers/MissionEventHandler.cs
mrica-equinor: Ok so what happened when I tried this was the robot became available before the mission queue frozen was set to false and because isar only send one event with robot available, the mission queue didn't run the next mission. So the emergency status was cleared but the robot didn't start any mission from the queue.. So I think it's best to keep it this way, what do you think?

backend/api/EventHandlers/MissionScheduling.cs
mrica-equinor: This doesn't lead to problems if the mission queue is still frozen after clearing the emergency state when the robot gets available because no emergency state mission will be found so the robot will just keep the first mission in the queue and run it, that's why it works the way it has been implemented

backend/api/EventHandlers/MissionEventHandler.cs
mrica-equinor: This doesn't lead to problems if the mission queue is still frozen after clearing the emergency state when the robot gets available because no emergency state mission will be found so the robot will just keep the first mission in the queue and run it, that's why it works the way it has been implemented

mrica-equinor: It's a timing issue and I don't see another way of guaranteeing that isar will ping the robot status as available always before the mission queue frozen has been updated to false

backend/api/Database/Context/FlotillaDbContext.cs
aestene: I suspect a rebase might be needed here. These are not changes I made with these files. We made some changes with this file when reverting your previous fix @oysand and I think this branch might not contain them and it looks fishy.

Try a rebase @mrica-equinor and we will see after.

backend/api/Database/Context/FlotillaDbContext.cs
aestene: Same thing as above @mrica-equinor

frontend/src/language/no.json
```diff
@@ -174,5 +174,10 @@
     "Hours between inspections": "Timer mellom inspeksjoner",
     "Days between inspections": "Dager mellom inspeksjoner",
     "Confirm plant": "Bekreft anlegg",
-    "N": "N"
+    "N": "N",
+    "Send robots to safe zone": "Send roboter til trygg sone",
+    "Send robots to safe zone long text": "Du er i ferd med å sende robotene til en trygg sone og dermed stoppe alle pågående oppdrag. For å kjøre oppdragene igjen må robotene starte fra begynnelsen.",
+    "Send robots to safe confirmation text": "Er du sikker på at du vil sende alle robotene til en trygg sone?",
+    "Dismiss robots from safe zone long text": "Vanlig drift kan fortsette ved å starte oppdragene i køen eller ved å slette dem.",
+    "Dismiss robots from safe zone": "Slepp robotene ut av trygg sone"
```
Eddasol, line 182:
```suggestion
    "Dismiss robots from safe zone": "Slipp robotene ut av trygg sone"
```

frontend/src/components/Pages/FrontPage/MissionOverview/StopDialogs.tsx
```diff
@@ -0,0 +1,234 @@
...
+                                    variant="outlined"
+                                    color="danger"
+                                    onClick={() => {
+                                        setIsStopRobotDialogOpen(false)
+                                    }}
+                                >
+                                    {TranslateText('Cancel')}
+                                </Button>
+                                <Button variant="contained" color="danger" onClick={resetRobots}>
+                                    {TranslateText('Continue missions')}
```
Eddasol, line 225: Missing translation

frontend/src/models/Robot.ts
```diff
@@ -8,6 +8,7 @@ export enum RobotStatus {
     Available = 'Available',
     Offline = 'Offline',
     Busy = 'Busy',
+    SafeZone = 'Safe Zone',
```
Eddasol, line 11: Needs translation

frontend/src/components/Pages/FrontPage/MissionOverview/StopDialogs.tsx
```diff
@@ -0,0 +1,234 @@
...
+                                </Button>
+                                <Button variant="contained" color="danger" onClick={stopAll}>
+                                    {TranslateText('Send robots to safe zone')}
+                                </Button>
+                            </StyledDisplayButtons>
+                        </Dialog.Actions>
+                    </StyledDialog>
+                </>
+            )}
+            {safeZoneStatus == true && (
```
Eddasol, line 190:
```suggestion
            {safeZoneStatus && (
```

frontend/src/components/Pages/FrontPage/MissionOverview/StopDialogs.tsx
```diff
@@ -0,0 +1,234 @@
+import { Button, Dialog, Typography, Icon } from '@equinor/eds-core-react'
+import styled from 'styled-components'
+import { useLanguageContext } from 'components/Contexts/LanguageContext'
+import { Icons } from 'utils/icons'
+import { useState, useEffect } from 'react'
+import { tokens } from '@equinor/eds-tokens'
+import { Mission } from 'models/Mission'
+import { useMissionControlContext } from 'components/Contexts/MissionControlContext'
+import { BackendAPICaller } from 'api/ApiCaller'
+import { Robot } from 'models/Robot'
```
Eddasol, line 10:
```suggestion
```
Not in use

frontend/src/components/Pages/FrontPage/RobotCards/RobotStatusChip.tsx
```diff
@@ -11,10 +12,13 @@ enum StatusColors {
     Available = '#A1DAA0',
     Offline = '#F7F7F7',
     Busy = '#FFC67A',
+    SafeZone = '#FF0000',
 }
 
 export function RobotStatusChip({ status }: StatusProps) {
     const { TranslateText } = useLanguageContext()
+    const { safeZoneStatus, switchSafeZoneStatus } = useSafeZoneContext()
```
Eddasol, line 20:
```suggestion
    const { safeZoneStatus } = useSafeZoneContext()
```
Not in use

backend/api/EventHandlers/MissionEventHandler.cs
andchiind: Could you acquire the lock used to freeze the mission queue here? So that it will run as soon as the mission queue becomes unlocked? I haven't personally tried it, but hopefully these mutexes can be passed around.

andchiind: Hmm I now see that we do not use any mutexes in this process, which can cause problems regardless. I'll need to investigate this algorithm a bit more.

andchiind: Can we maybe trigger an event when the mission run queue becomes unfrozen? And then do this in that event handler if we have not already done so?

frontend/src/components/Pages/FrontPage/MissionOverview/StopDialogs.tsx
andchiind: Potensial translation: "Fortsett oppdragene" or "Fortsett oppdrag"

frontend/src/models/Robot.ts
andchiind: I think we use "trygg sone" as the translation btw @mrica-equinor, we're just missing this capitalisation

backend/api/EventHandlers/MissionEventHandler.cs
```diff
@@ -102,111 +108,103 @@ private async void OnRobotAvailable(object? sender, RobotAvailableEventArgs e)
...
+            if (area == null)
             {
-                _logger.LogInformation("Mission {MissionRunId} was put on the queue as the system may not start a mission now", missionRun.Id);
+                _logger.LogError("Could not find area with ID {AreaId}", e.AreaId);
                 return;
             }
 
+            await MissionSchedulingService.FreezeMissionRunQueueForRobot(e.RobotId);
```
andchiind, line 159: In this function we should either check to make sure the robot is not null, or we need to put this function call inside the try catch

backend/api/EventHandlers/MissionScheduling.cs
```diff
@@ -0,0 +1,343 @@
...
+        {
+            robot.MissionQueueFrozen = false;
+            await _robotService.Update(robot);
+            _logger.LogInformation("Mission queue for robot {RobotName} with ID {RobotId} was unfrozen", robot.Name, robot.Id);
+        }
+
+        public async Task UnfreezeMissionRunQueueForRobot(string robotId)
+        {
+            var robot = await _robotService.ReadById(robotId);
+            if (robot == null)
```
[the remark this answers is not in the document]
mrica-equinor: What if the robot for some weird reason gets deleted from the database while it is in the safe zone? I can delete it but this call won't be made that often so don't mind keeping it

backend/api/EventHandlers/MissionEventHandler.cs
mrica-equinor: We do this on line 146, is that what you were asking about?

mrica-equinor:
Anyone knows why I am getting these errors ?

image

backend/api/EventHandlers/MissionEventHandler.cs
andchiind: sorry, I missed that. I guess in theory the state could change in the mean time, but it's unlikely.

backend/api/EventHandlers/MissionScheduling.cs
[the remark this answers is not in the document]
mrica-equinor: I think might be good now, these functions were a bit odd sorry!

backend/api/EventHandlers/MissionEventHandler.cs
```diff
@@ -102,111 +108,103 @@ private async void OnRobotAvailable(object? sender, RobotAvailableEventArgs e)
...
-                _logger.LogInformation("Mission run {MissionRunId} was not started as there is already an ongoing mission", missionRun.Id);
-                return false;
-            }
-            if (robot.Status is not RobotStatus.Available)
-            {
-                _logger.LogInformation("Mission run {MissionRunId} was not started as the robot is not available", missionRun.Id);
-                return false;
-            }
-            if (!robot.Enabled)
+            _logger.LogInformation("Triggered EmergencyButtonPressed event for robot ID: {RobotId}", e.RobotId);
```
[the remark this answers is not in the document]
andchiind: It's probably not necessary since the actions seem to be idempotent.

backend/api/Database/Context/FlotillaDbContext.cs
```diff
@@ -71,6 +71,8 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
 
         modelBuilder.Entity<MissionRun>().OwnsOne(m => m.Map).OwnsOne(t => t.TransformationMatrices);
         modelBuilder.Entity<MissionRun>().OwnsOne(m => m.Map).OwnsOne(b => b.Boundary);
+        // modelBuilder.Entity<MissionRun>().HasOne(m => m.Area).WithMany();
```
oysand, line 74: Remove the commented ones

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -229,17 +243,18 @@ private async void OnMissionUpdate(object? sender, MqttReceivedArgs mqttArgs)
             await robotService.Update(robot);
             _logger.LogInformation("Robot '{Id}' ('{Name}') - completed mission {MissionId}", robot.IsarId, robot.Name, flotillaMissionRun.MissionId);
 
-            if (!flotillaMissionRun.IsCompleted) { return; }
-
-            await taskDurationService.UpdateAverageDurationPerTask(robot.Model.Type);
+            if (flotillaMissionRun.IsCompleted)
```
oysand, line 246: Change to guard statement

mrica-equinor:
There's a bug, if you localize the robot in an area without safe zone it's crashing the backend. We should not merge this until it is fixed. I am not sure if this bug was introduced today or not but will fix it asap

oysand (approved):
LGTM

PR #1050, opened 2023-10-11
Remove unnecessary use of var

Eddasol:


frontend/src/components/Pages/FrontPage/MissionOverview/CreateMissionButton.tsx
```diff
@@ -6,8 +6,8 @@ import { useInstallationContext } from 'components/Contexts/InstallationContext'
 
 export const CreateMissionButton = (): JSX.Element => {
     const { TranslateText } = useLanguageContext()
-    const [isPopoverOpen, setIsPopoverOpen] = useState<boolean>(false)
     const { installationCode } = useInstallationContext()
+    const [isPopoverOpen, setIsPopoverOpen] = useState<boolean>(false)
```
andchiind, line 10: Are these reorderings done to group the useStates away from the useContexts? Just curious

frontend/src/components/Pages/FrontPage/RobotCards/RobotStatusView.tsx
```diff
@@ -47,25 +48,24 @@ export function RobotStatusSection({ refreshInterval }: RefreshProps) {
...
-    var robotDisplay
-    if (installationCode === '') {
-        robotDisplay = robots.map(function (robot) {
-            return <RobotStatusCard key={robot.id} robot={robot} />
-        })
-    } else {
-        robotDisplay = filteredRobots.map(function (robot) {
-            return <RobotStatusCard key={robot.id} robot={robot} />
-        })
+    const robotDisplay = () => {
```
andchiind, line 59: What's the reasoning for making robotDisplay a function here?

frontend/src/components/Pages/MissionPage/VideoStream/VideoStreamCards.tsx
```diff
@@ -57,37 +55,39 @@ export function VideoStreamCard({ videoStream, toggleFullScreenMode, setFullScre
         </FullscreenButton>
     )
 
-    if (IsValidOvenPlayerType({ videoStream })) {
-        if (videoStream.shouldRotate270Clockwise) {
-            videoPlayer = (
-                <StyledVideoSectionRotated>
-                    <Rotate>
+    const videoPlayer = () => {
```
andchiind, line 58: Since this is now a function, should we give it a verb name?

frontend/src/components/Pages/FrontPage/RobotCards/RobotStatusView.tsx
andchiind: If we do keep if a function then the naming change mentioned below should also apply. I think it's best to reserve noun names to variables.

frontend/src/components/Pages/FrontPage/MissionOverview/CreateMissionButton.tsx
Eddasol: Yes, looks a lot better and I find it easier to read

frontend/src/components/Pages/FrontPage/RobotCards/RobotStatusView.tsx
Eddasol: Reasoning for making it a function is that I prefer having a constant instead of a variable that can be changed. Also prefer clustering it into a functions so its easier to see whats relevant for that variable. Open for discussion on this.

Would getRobotDisplay be a better name?

andchiind: Yes that seems like a good name in my opinion

andchiind (approved):
LGTM

PR #1054, opened 2023-10-12
Fail mission when ISAR connection is lost

### LINKED ISSUE #1044: Bad gateway returned if mission stopped after isar stops
**Describe the bug**
If Isar is closed during a mission the mission still appears to be running in Flotilla. If you then stop the mission in the flotilla frontend, a bad gateway response is returned.

**To Reproduce**
Same as the description.

**Expected behavior**
Closing Isar while a mission is running should make the mission fail. 

**Screenshots**
The response when trying to update the robot to stop the mission:
image
This loading icon never goes away:
image

andchiind:
Closes #1044 

This PR fixes stopping the ongoing mission when connection to ISAR is lost, as well as an error that occurs when ISAR is started again. Errors were occuring since db context objects were taken out of scoped services and then used again from hosted services, which causes database errors since they are unable to verify that the object being modified is the same as before.

We still get a failed POST if we stop the mission as soon as we lose ISAR connection. I would like to create a banner alert for this, but I will then first wait for the failed request alert banner PR to be merged https://github.com/equinor/flotilla/pull/1018 .

### DIFF
--- a/backend/api/Services/RobotService.cs
+++ b/backend/api/Services/RobotService.cs
@@ -15,6 +15,7 @@ namespace Api.Services
         public Task<Robot?> ReadByIsarId(string isarId);
         public Task<Robot> Update(Robot robot);
         public Task<Robot?> Delete(string id);
+        public Task<Robot?> DisableRobotByIsarId(string isarId);
     }
 
     [SuppressMessage(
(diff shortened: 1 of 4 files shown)


backend/api/Services/RobotService.cs
```diff
@@ -99,6 +101,28 @@ public async Task<Robot> Update(Robot robot)
             return robot;
         }
 
+        public async Task<Robot?> DisableRobotByIsarId(string isarId)
+        {
+            var existingRobot = await ReadByIsarId(isarId);
+            if (existingRobot == null) return null;
```
aestene, line 107: If this happens it should be logged with an error message

backend/api/Services/RobotService.cs
```diff
@@ -99,6 +101,28 @@ public async Task<Robot> Update(Robot robot)
...
+            existingRobot.CurrentMissionId = null;
+            var entry = _context.Update(existingRobot);
+            await _context.SaveChangesAsync();
+            return entry.Entity;
+        }
+
+        public async Task<Robot?> EnableRobotByIsarId(string isarId)
+        {
+            var existingRobot = await ReadByIsarId(isarId);
+            if (existingRobot == null) return null;
```
aestene, line 119: Should have a log error message

PR #1055, opened 2023-10-13
Improve appearance of mission definition page

### LINKED ISSUE #944: Allow users to edit mission definition inspection frequency

andchiind:
Closes #944 since I haven't been able to find any issues with the functionality.  This PR therefore just addresses appearances. These may change in the future as we get feedback from UX design, but this PR regardless makes it easier to edit this style in the future if needed.


frontend/src/components/Pages/MissionDefinitionPage/MissionDefinitionPage.tsx
```diff
@@ -48,21 +62,25 @@ const StyledMissionDefinitionPage = styled.div`
...
     return (
-        <>
+        <StyledFormItem>
             <Typography
                 variant="body_long_bold"
                 group="paragraph"
                 color={tokens.colors.text.static_icons__secondary.rgba}
             >
                 {left}
+                {onEdit && <Icon style={{ paddingLeft: '5px' }} name={Icons.Edit} size={16} onClick={onEdit}></Icon>}
```
Eddasol, line 83: If the icon is clickable its better to make it a button with an icon such that the pointer indicates that its clickable

andchiind:
This is how the page looks at the moment 
image

frontend/src/components/Pages/MissionDefinitionPage/MissionDefinitionPage.tsx
```diff
@@ -94,45 +123,83 @@ function MissionDefinitionPageBody({ missionDefinition, updateMissionDefinition
...
-            <Button
+            <StyledFormContainer>
+                <KeyValuePairDisplay
+                    left={TranslateText('Name')}
+                    right={missionDefinition.name}
+                    onEdit={onEdit('name')}
+                />
+                <KeyValuePairDisplay left={TranslateText('Area')} right={missionDefinition.area.areaName} />
+                <KeyValuePairDisplay left={TranslateText('Deck')} right={missionDefinition.area.deckName} />
+                <KeyValuePairDisplay left={TranslateText('Plant')} right={missionDefinition.area.plantCode} />
```
Eddasol, line 143: Should we use plantName instead of plantCode?

frontend/src/components/Pages/MissionDefinitionPage/MissionDefinitionPage.tsx
```diff
@@ -175,68 +243,78 @@ function MissionDefinitionEditButtons({ missionDefinition, updateMissionDefiniti
...
+                {getFormItem()}
+                <StyledButtonSection>
+                    <Button onClick={handleSubmit} variant="outlined" color="primary">
+                        {' '}
+                        {TranslateText('Update mission definition')}{' '}
+                    </Button>
+                    <Button onClick={() => closeEditDialog()} variant="outlined" color="primary">
+                        {' '}
+                        {TranslateText('Cancel')}{' '}
+                    </Button>
```
Eddasol, line 314: To keep the same style as other dialogs the cancel button should be on the left and the other button should not be outlined

frontend/src/components/Pages/MissionDefinitionPage/MissionDefinitionPage.tsx
```diff
@@ -7,16 +7,16 @@ import { BackendAPICaller } from 'api/ApiCaller'
...
 import { config } from 'config'
+import { Icons } from 'utils/icons'
 
-const StyledFormDialog = styled.div`
-    display: flex;
-    justify-content: space-between;
-`
-const StyledAutoComplete = styled(Card)`
+const boxBackgroundColour = tokens.colors.ui.background__light.hex
+const boxBorderColour = tokens.colors.interactive.primary__resting.hex
```
Eddasol, line 17:
```suggestion
```
Not in use

frontend/src/components/Pages/MissionDefinitionPage/MissionDefinitionPage.tsx
```diff
@@ -71,13 +93,20 @@ function KeyValuePairDisplay({ left, right }: { left: string; right: any }) {
...
+interface IMissionDefinitionPageBodyProps {
+    missionDefinition: CondensedMissionDefinition
+    updateMissionDefinition: (missionDefinition: CondensedMissionDefinition) => void
+}
+
+function MissionDefinitionPageBody({ missionDefinition, updateMissionDefinition }: IMissionDefinitionPageBodyProps) {
     const { TranslateText } = useLanguageContext()
     let navigate = useNavigate()
+    const [isEditDialogOpen, setIsEditDialogOpen] = useState<boolean>(false)
+    const [selectedField, setSelectedField] = useState<string>('')
```
Eddasol, line 109:
```suggestion
function MissionDefinitionPageBody({ missionDefinition, updateMissionDefinition }: IMissionDefinitionPageBodyProps) {
    const { TranslateText } = useLanguageContext()
    const [isEditDialogOpen, setIsEditDialogOpen] = useState<boolean>(false)
    const [selectedField, setSelectedField] = useState<string>('')
    let navigate = useNavigate()
```

frontend/src/components/Pages/MissionDefinitionPage/MissionDefinitionPage.tsx
```diff
@@ -48,21 +62,29 @@ const StyledMissionDefinitionPage = styled.div`
...
+    }
+`
 
-interface IProps {
-    missionDefinition: CondensedMissionDefinition
-    updateMissionDefinition: (missionDefinition: CondensedMissionDefinition) => void
-}
-
-function KeyValuePairDisplay({ left, right }: { left: string; right: any }) {
+function KeyValuePairDisplay({ left, right, onEdit }: { left: string; right: any; onEdit?: () => void }) {
```
Eddasol, line 74: Suggest renaming left and right as they aren't actually left and right. 
Maybe something like "key" and "value" or "heading" and "content"

frontend/src/components/Pages/MissionDefinitionPage/MissionDefinitionPage.tsx
Eddasol, line 109: On line 111 - 123:
Do we need minutes in line 116?

Also suggest removing hour cause I think we can assume that inspection frequency will be in days (can add it again if it turns out to be wrong)

Eddasol: Suggest that it says "Inspeksjon kreves hver 4. dag" instead
image

frontend/src/components/Pages/MissionDefinitionPage/MissionDefinitionPage.tsx
```diff
@@ -94,45 +123,83 @@ function MissionDefinitionPageBody({ missionDefinition, updateMissionDefinition
...
+                <KeyValuePairDisplay
+                    left={TranslateText('Comment')}
+                    right={missionDefinition.comment}
+                    onEdit={onEdit('comment')}
+                />
+                <KeyValuePairDisplay
+                    left={TranslateText('Inspection frequency')}
+                    right={displayInspectionFrequency(missionDefinition.inspectionFrequency)}
+                    onEdit={onEdit('inspectionFrequency')}
+                />
```
Eddasol, line 161: If we write something that isn't a number "NaN" will show. Suggest emptying such that the prompt stays empty and add a "invalid input" feedback to the user

Eddasol: image

frontend/src/components/Pages/MissionDefinitionPage/MissionDefinitionPage.tsx
```diff
@@ -48,21 +62,29 @@ const StyledMissionDefinitionPage = styled.div`
...
                 variant="body_long_bold"
                 group="paragraph"
                 color={tokens.colors.text.static_icons__secondary.rgba}
             >
                 {left}
+                {onEdit && (
+                    <Button style={{ padding: '5px' }} variant="ghost" onClick={onEdit}>
+                        <Icon name={Icons.Edit} size={16} />
+                    </Button>
+                )}
```
Eddasol, line 84: The ones with button becomes taller and doesn't align
image

frontend/src/components/Pages/MissionDefinitionPage/MissionDefinitionPage.tsx
```diff
@@ -175,68 +240,78 @@ function MissionDefinitionEditButtons({ missionDefinition, updateMissionDefiniti
...
-                                </Button>
-                            </StyledButtonSection>
-                        </StyledAutoComplete>
-                    </StyledDialog>
-                </StyledFormDialog>
-            )}
-        </>
+        <StyledDialog open={true}>
+            <StyledFormCard>
+                <Typography variant="h2">{TranslateText('Edit mission definition')}</Typography>
```
Eddasol, line 301: Don't think we should use the word mission definition in the user interface. How about having "Edit name", "Edit comment" .. and such add header instead? And changing the text on the confirm button to just "Update" instead of "update mission definition"

frontend/src/components/Pages/MissionDefinitionPage/MissionDefinitionPage.tsx
andchiind: I'll just not allow non-numeric input

frontend/src/components/Pages/MissionDefinitionPage/MissionDefinitionPage.tsx
andchiind: This wasn't in the backend models, but I've added it now

frontend/src/components/Pages/MissionDefinitionPage/MissionDefinitionPage.tsx
```diff
@@ -28,18 +25,32 @@ const StyledButtonSection = styled.div`
...
-    margin-left: auto;
-    margin-right: 0;
-    gap: 10px;
+    flex-wrap: wrap;
+    justify-content: flex-start;
+    flex-direction: row;
+    align: left;
+    align-items: flex-start;
+    align-content: flex-start;
+    width: 50%;
```
Eddasol, line 36: When the screen is slim (iPad mode) the table becomes  very small and wraps even if there is more room

frontend/src/components/Pages/MissionDefinitionPage/MissionDefinitionPage.tsx
```diff
@@ -48,91 +59,138 @@ const StyledMissionDefinitionPage = styled.div`
...
             <Typography
                 variant="body_long_bold"
                 group="paragraph"
                 color={tokens.colors.text.static_icons__secondary.rgba}
             >
-                {left}
+                {title}
+                {onEdit && (
+                    <StyledEditButton variant="ghost" onClick={onEdit}>
+                        <Icon name={Icons.Edit} size={16} />
```
Eddasol, line 88: image

The edit icon looks a bit weird when hovering

andchiind: Would you prefer if I disabled onHover for it?

mrica-equinor (changes requested):

frontend/src/components/Pages/MissionDefinitionPage/MissionDefinitionPage.tsx
```diff
@@ -48,21 +62,29 @@ const StyledMissionDefinitionPage = styled.div`
...
             >
                 {left}
+                {onEdit && (
+                    <Button style={{ padding: '5px' }} variant="ghost" onClick={onEdit}>
+                        <Icon name={Icons.Edit} size={16} />
+                    </Button>
+                )}
             </Typography>
             <Typography
                 variant="body_long_italic"
```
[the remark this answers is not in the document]
andchiind: I think it should be quite equivalent now

mrica-equinor (approved):
LGTM :)

PR #1057, opened 2023-10-13
Use target and pose directly from Echo request

aestene:
Note that the call to STID and the `InspectionTarget` in `MissionTask` has not been removed yet. It was decided to do that in another PR to avoid too many breaking changes in one.


backend/api/Controllers/Models/EchoInspection.cs
```diff
@@ -43,18 +47,23 @@ public class EchoInspectionComparer : IEqualityComparer<EchoInspection>
         public bool Equals(EchoInspection? e1, EchoInspection? e2)
         {
             if (ReferenceEquals(e1, e2))
+            {
```
andchiind, line 50: All these extra lines will slow down the program immensely 🤢

backend/api/Controllers/Models/EchoInspection.cs
```diff
@@ -43,18 +47,23 @@ public class EchoInspectionComparer : IEqualityComparer<EchoInspection>
...
 
             if (e2 is null || e1 is null)
+            {
                 return false;
+            }
 
             return e1.InspectionType == e2.InspectionType
-                && e1.TimeInSeconds == e2.TimeInSeconds;
+                   && e1.TimeInSeconds == e2.TimeInSeconds
+                   && e1.InspectionPoint == e2.InspectionPoint;
```
andchiind, line 61: Will this work even if the Position class does not override an Equals method?

backend/api/Services/Models/EchoPoseResponse.cs
andchiind, line 1: This file should probably be renamed

backend/api/Migrations/20231013121440_AddInspectionTargetToInspection.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Api.Migrations
+{
+    /// <inheritdoc />
+    public partial class AddInspectionTargetToInspection : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<float>(
```
andchiind, line 13: Will these columns be populated manually? (With an SQL script I mean)

andchiind (commented):
Looks good

backend/api/Controllers/Models/EchoInspection.cs
aestene: Good catch and I'm not sure. Decided to override the equals method the same way as has been done for orientation.

backend/api/Controllers/Models/EchoInspection.cs
aestene: Fixed

backend/api/Services/Models/EchoPoseResponse.cs
aestene: Done

andchiind (approved):
Looks good to me, and seems to work well.

backend/api/Controllers/Models/EchoInspection.cs
```diff
@@ -42,19 +46,18 @@ public class EchoInspectionComparer : IEqualityComparer<EchoInspection>
...
             return e1.InspectionType == e2.InspectionType
-                && e1.TimeInSeconds == e2.TimeInSeconds;
+                   && e1.TimeInSeconds == e2.TimeInSeconds
+                   && e1.InspectionPoint.Equals(e2.InspectionPoint);
         }
 
         public int GetHashCode(EchoInspection e)
         {
-            // We cannot incorporate TimeInSeconds here are SQL queries do not handle 
+            // We cannot incorporate TimeInSeconds here are SQL queries do not handle
```
Christdej, line 60:
```suggestion
            // We cannot incorporate TimeInSeconds here as SQL queries do not handle
```

Christdej (approved):
LGTM

PR #1058, opened 2023-10-13
Fix mission order

Eddasol:


Eddasol:
image
Removing orderBy fixes the ordering but causes this warning

Christdej (commented):
I think we should try to make the error go away some way, as they are normally there for a reason, if not we can add statements to ignore the warning

backend/api/Utilities/PagedList.cs
```diff
@@ -33,14 +33,12 @@ public PagedList(List<T> items, int currentPage, int pageSize, int totalCount)
...
         {
-            // Adding order by constant value 1, which does not change order of rows
-            // but stops the runtime warning
             int totalCount = await source.CountAsync();
             List<T> items;
             if (source.Expression.Type is IOrderedQueryable)
                 items = await source.Skip((pageNumber - 1) * pageSize).Take(pageSize).ToListAsync();
             else
-                items = await source.OrderBy(x => 1).Skip((pageNumber - 1) * pageSize).Take(pageSize).ToListAsync();
+                items = await source.Skip((pageNumber - 1) * pageSize).Take(pageSize).ToListAsync();
```
andchiind, line 41: we might as well remove the if-else if we make this change I guess

andchiind: I'm not quite sure when this else statement is reached for mission runs though, since in the ReadAll method we always return an OrderedQueryable. I think that's something we need to investigate. It seems like the only time we would reach this else statement is when we fetch Source objects.

backend/api/Utilities/PagedList.cs
```diff
@@ -37,7 +38,7 @@ int pageSize
             // but stops the runtime warning
             int totalCount = await source.CountAsync();
             List<T> items;
-            if (source.Expression.Type is IOrderedQueryable)
+            if (source.Expression.Type == typeof(IOrderedQueryable<MissionRun>))
                 items = await source.Skip((pageNumber - 1) * pageSize).Take(pageSize).ToListAsync();
             else
                 items = await source.OrderBy(x => 1).Skip((pageNumber - 1) * pageSize).Take(pageSize).ToListAsync();
```
Christdej, line 44: I think we can add a comment to explain why we are doing something else for MissionRuns

Christdej (changes requested):

Christdej (approved):

PR #1071, opened 2023-10-18
Implement localization as part of a mission

aestene:


backend/api/Database/Models/Inspection.cs
```diff
@@ -50,16 +53,24 @@ public Inspection(Inspection copy)
...
         public InspectionStatus Status
         {
             get => _status;
             set
             {
                 _status = value;
-                if (IsCompleted && EndTime is null) { EndTime = DateTime.UtcNow; }
-
-                if (_status is InspectionStatus.InProgress && StartTime is null) { StartTime = DateTime.UtcNow; }
+                if (IsCompleted && EndTime is null)
```
Afonso-2403, line 65: Why does it format into one line in other files and here it split into multiple?

backend/api/Services/LocalizationService.cs
```diff
@@ -0,0 +1,230 @@
...
+            _missionRunService = missionRunService;
+            _installationService = installationService;
+        }
+
+        public async Task EnsureRobotIsOnSameInstallationAsMission(Robot robot, MissionDefinition missionDefinition)
+        {
+            var missionInstallation = await _installationService.ReadByName(missionDefinition.InstallationCode);
+            var robotInstallation = await _installationService.ReadByName(robot.CurrentInstallation);
+
+            if (missionInstallation is null || robotInstallation is null)
```
Afonso-2403, line 37: This message might be misleading if the robot has no current Installation

backend/api/Services/LocalizationService.cs
```diff
@@ -0,0 +1,230 @@
...
+                throw new RobotLocalizationException(errorMessage);
+            }
+        }
+
+        public async Task<bool> RobotIsLocalized(string robotId)
+        {
+            var robot = await _robotService.ReadById(robotId);
+            if (robot != null)
+            {
+                return robot.CurrentArea is not null;
```
Afonso-2403, line 76: Any reason for having the return inside the if instead of the throwing of the exception?
Given that in most places we do the opposite this looks less readable to me

backend/api/Services/LocalizationService.cs
```diff
@@ -0,0 +1,230 @@
...
+
+            if (lastExecutedMissionRun.Status != MissionStatus.Successful)
+            {
+                string errorMessage =
+                    $"The localization mission {lastExecutedMissionRun.Id} failed and thus subsequent scheduled missions for deck {lastExecutedMissionRun.Area?.Deck} wil be cancelled";
+                _logger.LogError("{Message}", errorMessage);
+                throw new LocalizationFailedException(errorMessage);
+            }
+
+            await _robotService.SetCurrentArea(lastExecutedMissionRun.Area, robot.Id);
```
Afonso-2403, line 115: Probably in the rebase this function was removed?

backend/api/Services/LocalizationService.cs
```diff
@@ -0,0 +1,230 @@
...
+
+            var lastExecutedMissionRun = await _missionRunService.ReadLastExecutedMissionRunByRobot(robot.Id);
+            if (lastExecutedMissionRun is null)
+            {
+                string errorMessage = $"Could not find last executed mission run for robot with ID {robot.Id}";
+                _logger.LogError("{Message}", errorMessage);
+                throw new MissionNotFoundException(errorMessage);
+            }
+
+            if (lastExecutedMissionRun.Status != MissionStatus.Successful)
```
Afonso-2403, line 107: This means that if the previous mission was stopped (cancelled), for example, all the subsequent missions will be cancelled, is that intended behaviour?

backend/api/Services/MissionDefinitionService.cs
```diff
@@ -179,25 +187,75 @@ public async Task<MissionDefinition> Update(MissionDefinition missionDefinition)
             }
         }
 
+        public async Task<MissionDefinition> FindExistingOrCreateCustomMissionDefinition(CustomMissionQuery customMissionQuery, List<MissionTask> missionTasks)
```
Afonso-2403, line 190: This function seems to already exist in `CustomMissionSchedulingService.cs`

backend/api/Controllers/MissionSchedulingController.cs
```diff
@@ -74,19 +78,25 @@ ICustomMissionSchedulingService customMissionSchedulingService
...
+            if (robot is null) { return NotFound($"Could not find robot with id {scheduledMissionQuery.RobotId}"); }
 
             var missionDefinition = await _missionDefinitionService.ReadById(scheduledMissionQuery.MissionDefinitionId);
-            if (missionDefinition == null)
-            {
-                return NotFound("Mission definition not found");
-            }
+            if (missionDefinition == null) { return NotFound("Mission definition not found"); }
+
+            try { await _localizationService.EnsureRobotIsOnSameInstallationAsMission(robot, missionDefinition); }
```
Afonso-2403, line 86: Scheduling Echo missions in the `Create` function below in this file should also ensure that the robot is localized, right?

backend/api/Services/MissionDefinitionService.cs
aestene: This was removed in a rebase so no longer present.

backend/api/Controllers/MissionSchedulingController.cs
aestene: No, currently we have no functionality to say which area an Echo mission is in. For now the Echo mission scheduling will be removed (in the frontend) until we fix that and then we will also include localization for those missions.

backend/api/Services/LocalizationService.cs
```diff
@@ -0,0 +1,230 @@
...
+        {
+            var robot = await _robotService.ReadById(robotId);
+            if (robot == null)
+            {
+                string errorMessage = $"Robot with ID: {robotId} was not found in the database";
+                _logger.LogError("{Message}", errorMessage);
+                throw new RobotNotFoundException(errorMessage);
+            }
+
+            if (await _missionRunService.OngoingMission(robot.Id))
```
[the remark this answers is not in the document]
aestene: The assumption is that there should be no ongoing mission. However, this was implemented as a safety in case the reported result of the localization mission from ISAR was not reported before the OnRobotAvailable event was triggered.

backend/api/Services/LocalizationService.cs
aestene: Will change it

backend/api/Services/LocalizationService.cs
```diff
@@ -0,0 +1,247 @@
...
+                string errorMessage = $"Could not find installation for installation code {missionDefinition.InstallationCode} or the robot has no current installation";
+                logger.LogError("{Message}", errorMessage);
+                throw new InstallationNotFoundException(errorMessage);
+            }
+
+            if (robotInstallation != missionInstallation)
+            {
+                string errorMessage = $"The robot {robot.Name} is on installation {robotInstallation.Name} which is not the same as the mission installation {missionInstallation.Name}";
+                logger.LogError("{Message}", errorMessage);
+                throw new MissionException(errorMessage);
```
oysand, line 36: Consider a more descriptive exception

backend/api/EventHandlers/MissionEventHandler.cs
```diff
@@ -66,14 +71,27 @@ private async void OnMissionRunCreated(object? sender, MissionRunCreatedEventArg
                 return;
             }
 
+            string missionRunIdToStart = missionRun.Id;
+
+            _scheduleLocalizationSemaphore.WaitOne();
+
+            string? localizationMissionRunId = null;
+            try { localizationMissionRunId = await LocalizationService.EnsureRobotIsCorrectlyLocalized(missionRun.Robot, missionRun); }
+            catch (Exception ex) when (ex is AreaNotFoundException or DeckNotFoundException) { return; }
```
oysand, line 80: Why catch in this manner?

backend/api/EventHandlers/MissionEventHandler.cs
```diff
@@ -87,9 +105,38 @@ private async void OnRobotAvailable(object? sender, RobotAvailableEventArgs e)
                 return;
             }
 
+            if (!await LocalizationService.RobotIsLocalized(robot.Id))
+            {
+                try { await LocalizationService.EnsureRobotWasCorrectlyLocalizedInPreviousMissionRun(robot.Id); }
+                catch (Exception ex) when (ex is LocalizationFailedException or RobotNotFoundException or MissionNotFoundException or MissionException or TimeoutException)
+                {
+                    //TODO Handle failed localization - Cancel all the missions?
```
oysand, line 113: Handle todo in this PR

backend/api/EventHandlers/MissionEventHandler.cs
oysand: Based on the exception, we might want to cancel the mission

backend/api/Database/Context/InitDb.cs
```diff
@@ -51,9 +52,16 @@ private static List<Installation> GetInstallations()
                 InstallationCode = "JSV"
             };
 
+            var installation2 = new Installation
+            {
+                Id = Guid.NewGuid().ToString(),
+                Name = "Kårstø",
+                InstallationCode = "KAA"
```
oysand, line 59: Should we utilize Huldra or general names?

backend/api/Database/Context/InitDb.cs
```diff
@@ -67,9 +75,17 @@ private static List<Plant> GetPlants()
                 PlantCode = "P1"
             };
 
+            var plant2 = new Plant
+            {
+                Id = Guid.NewGuid().ToString(),
+                Installation = installations[0],
+                Name = "Kårstø",
+                PlantCode = "Kårstø"
```
oysand, line 83: Should we utilize Huldra or general names?

backend/api/Database/Models/MissionTask.cs
```diff
@@ -40,6 +41,32 @@ public MissionTask(CustomTaskQuery taskQuery)
             RobotPose = taskQuery.RobotPose;
             TaskOrder = taskQuery.TaskOrder;
             Status = TaskStatus.NotStarted;
+            Type = "inspection";
+        }
+
+        public MissionTask(Pose robotPose, string type)
+        {
+            switch (type)
```
oysand, line 49: Should this type be an enum?

backend/api/EventHandlers/MissionEventHandler.cs
```diff
@@ -87,9 +105,38 @@ private async void OnRobotAvailable(object? sender, RobotAvailableEventArgs e)
                 return;
             }
 
+            if (!await LocalizationService.RobotIsLocalized(robot.Id))
```
oysand, line 108: We have many checks to see that everything is in order, should we group these checks in a function?

backend/api/EventHandlers/MissionEventHandler.cs
```diff
@@ -87,9 +105,38 @@ private async void OnRobotAvailable(object? sender, RobotAvailableEventArgs e)
...
                 _logger.LogInformation("The robot was changed to available but there are no mission runs in the queue to be scheduled");
+
+                var lastExecutedMissionRun = await MissionService.ReadLastExecutedMissionRunByRobot(robot.Id);
+                if (lastExecutedMissionRun is null)
+                {
+                    _logger.LogError("Could not find last executed mission run for robot");
+                    return;
+                }
+
+                if (!lastExecutedMissionRun.IsDriveToMission())
```
oysand, line 129: Can we change this to a guard check?

backend/api/Database/Models/MissionTask.cs
oysand, line 49: What happens if the type provided is inspection?

backend/api/EventHandlers/MissionEventHandler.cs
```diff
@@ -87,9 +105,38 @@ private async void OnRobotAvailable(object? sender, RobotAvailableEventArgs e)
...
+                {
+                    try { await ReturnToHomeService.ScheduleReturnToHomeMissionRun(robot.Id); }
+                    catch (Exception ex) when (ex is RobotNotFoundException or AreaNotFoundException or DeckNotFoundException or PoseNotFoundException)
+                    {
+                        //TODO Handle failed return to home - Try again for a number of times?
```
oysand, line 134: Ensure to break the localization if this fails

backend/api/Services/LocalizationService.cs
```diff
@@ -0,0 +1,247 @@
...
+        {
+            var robot = await robotService.ReadById(robotId);
+            if (robot == null)
+            {
+                string errorMessage = $"Robot with ID: {robotId} was not found in the database";
+                logger.LogError("{Message}", errorMessage);
+                throw new RobotNotFoundException(errorMessage);
+            }
+
+            if (await missionRunService.OngoingMission(robot.Id)) { await WaitForLocalizationMissionStatusToBeUpdated(robot); }
```
oysand, line 88: Catch all relevant exceptions

backend/api/Services/LocalizationService.cs
```diff
@@ -0,0 +1,247 @@
...
+                string errorMessage = $"Could not find ongoing mission with ID {robot.CurrentMissionId}";
+                logger.LogError("{Message}", errorMessage);
+                throw new MissionNotFoundException(errorMessage);
+            }
+
+            if (!ongoingMissionRun.IsLocalizationMission())
+            {
+                string errorMessage = $"The currently executing mission for robot {robot.CurrentMissionId} is not a localization mission";
+                logger.LogError("{Message}", errorMessage);
+                throw new MissionException(errorMessage);
```
oysand, line 131: Consider a more descriptive exception name

backend/api/Services/LocalizationService.cs
```diff
@@ -0,0 +1,247 @@
...
+                DesiredStartTime = DateTime.UtcNow,
+                Tasks = new List<MissionTask>
+                {
+                    new(area.Deck.DefaultLocalizationPose.Pose, "localization")
+                },
+                Map = new MapMetadata()
+            };
+            logger.LogWarning("Starting localization mission");
+            await missionRunService.Create(localizationMissionRun, triggerCreatedMissionRunEvent: false);
+            await robotService.UpdateCurrentArea(robot.Id, localizationMissionRun.Area);
```
oysand, line 202: Only update current area once the localization mission has been successfully run. To avoid scheduling multiple localization missions we could check the mission queue for a pending or ongoing localization mission

backend/api/EventHandlers/MissionEventHandler.cs
```diff
@@ -66,14 +71,27 @@ private async void OnMissionRunCreated(object? sender, MissionRunCreatedEventArg
                 return;
             }
 
+            string missionRunIdToStart = missionRun.Id;
+
+            _scheduleLocalizationSemaphore.WaitOne();
+
+            string? localizationMissionRunId = null;
+            try { localizationMissionRunId = await LocalizationService.EnsureRobotIsCorrectlyLocalized(missionRun.Robot, missionRun); }
```
Afonso-2403, line 79: Change to something like if (ShouldScheduleLocalizationMission)

backend/api/EventHandlers/MissionEventHandler.cs
Afonso-2403: Changed to cancelling the mission for all exceptions

backend/api/EventHandlers/MissionEventHandler.cs
```diff
@@ -66,14 +71,27 @@ private async void OnMissionRunCreated(object? sender, MissionRunCreatedEventArg
...
+
             if (MissionScheduling.MissionRunQueueIsEmpty(await MissionService.ReadMissionRunQueue(missionRun.Robot.Id)))
             {
                 _logger.LogInformation("Mission run {MissionRunId} was not started as there are no mission runs on the queue", e.MissionRunId);
                 return;
             }
 
             _scheduleMissionSemaphore.WaitOne();
-            await MissionScheduling.StartMissionRunIfSystemIsAvailable(missionRun.Id);
+            await MissionScheduling.StartMissionRunIfSystemIsAvailable(missionRunIdToStart);
```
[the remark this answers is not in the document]
Afonso-2403: Changed to catch and just return since it is already logged that the missionRun Id could not be found

backend/api/Database/Context/InitDb.cs
Afonso-2403: This should be addressed in another PR

backend/api/Database/Models/MissionTask.cs
Afonso-2403: Changed to throwing an exception if the provided type is not Localization or DriveTo

backend/api/Services/LocalizationService.cs
Afonso-2403: The intended behaviour is that these are thrown by this function and caught by the caller of `WaitForLocalizationMissionStatusToBeUpdated`

backend/api/EventHandlers/MissionEventHandler.cs
Afonso-2403: `CurrentArea` is set to `null`, to handle in another PR the error handling behaviour of notifying the frontend

backend/api/Database/Context/InitDb.cs
andchiind: This has been addressed in another PR so we should now use Huldra here

backend/api/Services/AreaService.cs
```diff
@@ -70,11 +70,13 @@ public async Task<IEnumerable<Area>> ReadAll()
             var installation = await installationService.ReadByName(installationCode);
             if (installation == null) { return null; }
 
-            return await context.Areas.Where(a =>
-                    a.Installation != null && a.Installation.Id.Equals(installation.Id) &&
-                    a.Name.ToLower().Equals(areaName.ToLower())
-                ).Include(a => a.SafePositions).Include(a => a.Installation)
-                .Include(a => a.Plant).Include(a => a.Deck).FirstOrDefaultAsync();
+            return await context.Areas.Where(a => a.Installation.Id.Equals(installation.Id) && a.Name.ToLower().Equals(areaName.ToLower()))
```
andchiind, line 73: This should use GetAreas() since this currently ignores granular access.

backend/api/Services/LocalizationService.cs
```diff
@@ -0,0 +1,246 @@
...
+                string errorMessage = $"The currently executing mission for robot {robot.CurrentMissionId} is not a localization mission";
+                logger.LogError("{Message}", errorMessage);
+                throw new OngoingMissionNotLocalizationException(errorMessage);
+            }
+
+            logger.LogWarning(
+                "The RobotAvailable event was triggered before the OnMissionUpdate event and we have to wait to see that the localization mission is set to successful");
+
+            const int Timeout = 5;
+            var timer = new Stopwatch();
```
andchiind, line 137: There are ways of awaiting event handlers which we should use instead in the future, but this is probably fine for now.

backend/api/Services/MissionDefinitionService.cs
aestene: This was still left there. Removed the unused one in MissionDefinitionService

backend/api/Services/LocalizationService.cs
aestene: We should not be scheduling multiple localization missions now? I think this is a good improvement but would suggest adding it as an improvement in the next PR.

backend/api/EventHandlers/MissionEventHandler.cs
aestene: I'm not sure what we want with this comment? The function is in charge of scheduling the localization mission so the name ShouldScheduleLocalizationMission is not entirely correct. 

And we do use the exceptions to cancel the mission run if we fail localization?

backend/api/Database/Models/MissionRun.cs
```diff
@@ -188,6 +188,24 @@ public void SetToFailed()
...
             }
         }
+
+        public bool IsLocalizationMission()
+        {
+            if (Tasks.Count != 1)
+            {
+                return false;
+            }
+            return Tasks[0].Type == MissionTaskType.Localization;
```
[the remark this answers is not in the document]
aestene: Implemented

backend/api/EventHandlers/MissionEventHandler.cs
```diff
@@ -59,21 +64,45 @@ private async void OnMissionRunCreated(object? sender, MissionRunCreatedEventArg
...
             if (MissionScheduling.MissionRunQueueIsEmpty(await MissionService.ReadMissionRunQueue(missionRun.Robot.Id)))
             {
                 _logger.LogInformation("Mission run {MissionRunId} was not started as there are no mission runs on the queue", e.MissionRunId);
                 return;
             }
 
             _scheduleMissionSemaphore.WaitOne();
-            await MissionScheduling.StartMissionRunIfSystemIsAvailable(missionRun.Id);
+            try { await MissionScheduling.StartMissionRunIfSystemIsAvailable(missionRunIdToStart); }
+            catch (MissionRunNotFoundException) { return; }
```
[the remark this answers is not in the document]
aestene: The release of the semaphore should be in a finally statement, the return would lock the semaphore forever. Fixed it.

backend/api/Services/AreaService.cs
aestene: Fixed, and also fixed for other functions in the same service.

backend/api/EventHandlers/MissionEventHandler.cs
andchiind: The comment was that it would perhaps be better to have the function return a boolean instead of using exceptions to signal if the check failed or not. Using exceptions is useful if we use the returned exceptions to do different behaviour depending on the type of exception, but in this case the check is true or false.

andchiind (approved):
LGTM

aestene:
Creating new PR to manage migrations

PR #1074, opened 2023-10-20
Fix inconsistencies on the frontend

mrica-equinor:


frontend/src/components/Pages/FrontPage/MissionOverview/MissionQueueCard.tsx
```diff
@@ -33,12 +33,22 @@ const HorizontalNonButtonContent = styled.div`
 `
 
 const StyledConfirmDialog = styled.div`
```
andchiind, line 35: We can probably rename this to something other than a dialog since it’s not being used that way anymore

andchiind (commented):
LGTM but it’s probably best to wait for another approval to merge since I can’t actually test it atm

prasm313:
Looks good. Tested the improvements both in the ipad and computer view.

prasm313 (approved):
LGTM

PR #1080, opened 2023-10-25
Remove browser list in frontend README

anetteu:
This list is not maintained at the moment, and it is removed for now as no list is better than an outdated list. Consider to add a list of tested/supported browsers in the future.


eivindsjovold (approved):
LGTM

eivindsjovold (changes requested):
Remove commit message

PR #1082, opened 2023-10-26
Remove list of web browsers in README

anetteu:
A list with supported/tested web browsers is usefull, but it is  removed for now as it is not maintained.


eivindsjovold (approved):
LGTM

PR #1085, opened 2023-10-27
Fix database entity detached issues on mission update

aestene:


backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -85,26 +77,15 @@ private async void OnIsarRobotStatus(object? sender, MqttReceivedArgs mqttArgs)
...
-                );
+                _logger.LogInformation("Received message from unknown ISAR instance {Id} with robot name {Name}", isarRobotStatus.IsarId, isarRobotStatus.RobotName);
                 return;
             }
 
-            if (robot.Status == isarRobotStatus.RobotStatus)
-            {
-                return;
-            }
+            if (robot.Status == isarRobotStatus.RobotStatus) { return; }
```
andchiind, line 84: Could we maybe omit the braces here?

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -142,79 +121,43 @@ private async void OnIsarRobotInfo(object? sender, MqttReceivedArgs mqttArgs)
...
-                if (isarRobotInfo.VideoStreamQueries is not null)
-                {
-                    UpdateVideoStreamsIfChanged(isarRobotInfo.VideoStreamQueries, ref robot, ref updatedFields);
-                }
-
-                if (isarRobotInfo.Host is not null)
-                {
-                    UpdateHostIfChanged(isarRobotInfo.Host, ref robot, ref updatedFields);
-                }
+                if (isarRobotInfo.VideoStreamQueries is not null) { UpdateVideoStreamsIfChanged(isarRobotInfo.VideoStreamQueries, ref robot, ref updatedFields); }
```
andchiind, line 131: And here

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -142,79 +121,43 @@ private async void OnIsarRobotInfo(object? sender, MqttReceivedArgs mqttArgs)
...
+                if (isarRobotInfo.Host is not null) { UpdateHostIfChanged(isarRobotInfo.Host, ref robot, ref updatedFields); }
```
andchiind, line 132: And here

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -142,79 +121,43 @@ private async void OnIsarRobotInfo(object? sender, MqttReceivedArgs mqttArgs)
...
 
             var existingVideoStreams = robot.VideoStreams;
-            if (
-                updatedStreams.Count == robot.VideoStreams.Count
-                && updatedStreams.TrueForAll(stream => existingVideoStreams.Contains(stream))
-            )
-            {
-                return;
-            }
+            if (updatedStreams.Count == robot.VideoStreams.Count && updatedStreams.TrueForAll(stream => existingVideoStreams.Contains(stream))) { return; }
```
andchiind, line 160: And here, etc (there are more but I don’t want to spam you. Also this is subjective so if you prefer with the braces then that’s fine)

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -513,17 +344,10 @@ private async void OnPressureUpdate(object? sender, MqttReceivedArgs mqttArgs)
                 await timeseriesService.Create(
                     new RobotPressureTimeseries
                     {
-                        MissionId = robot.CurrentMissionId,
-                        Pressure = pressureStatus.PressureLevel,
-                        RobotId = robot.Id,
-                        Time = DateTimeOffset.UtcNow
+                        MissionId = robot.CurrentMissionId, Pressure = pressureStatus.PressureLevel, RobotId = robot.Id, Time = DateTimeOffset.UtcNow
```
andchiind, line 347: This is again personal preference but I do find splitting up these assignments across several lines to be more readable.

andchiind: At least when there’s more than a couple variables.

andchiind (commented):
This might be quite hard to do but would it maybe be possible to split away the change which fixes the database entity detached issue into a separate commit? Since currently it’s a bit hard to tell what changes are necessary in order to make sure the right context objects are used, and which changes were made to improve the code readability. No stress if the refactoring and fix are tightly coupled though.

backend/api/EventHandlers/MqttEventHandler.cs
aestene: This one is the one setting I can't find in Rider as I agree with you. Dotnet format also requires it so it will be fixed.

Eddasol (approved):
LGTM :)

aestene:
> This might be quite hard to do but would it maybe be possible to split away the change which fixes the database entity detached issue into a separate commit? Since currently it’s a bit hard to tell what changes are necessary in order to make sure the right context objects are used, and which changes were made to improve the code readability. No stress if the refactoring and fix are tightly coupled though.

The refactoring done in the second commit is basically the fix. So we wrap the different scoped services into a separate one so that we are not calling multiple different scoped services within the same event handler. So in a sense the second commit although it looks mostly like refactoring is the fix.

PR #1092, opened 2023-10-30
Fix concurrency issues for updating robot object

### LINKED ISSUE #1089: Resolve race condition in read and write to robot object
**Describe the bug**
When we read and write to the robot object we have a race condition in the sense that a write to the robot object in one thread may happen in-between a read and a write in a different thread. The example of this happening is whenever a mission finishes and the robot is set to `Status = Available`. If a telemetry event handler trigger simultaneously it may read the robot object and read the status as `Busy` and then write the status to `Busy` just after the `OnMissionUpdate` has just written the status to `Available`. As they are in different threads they do not share a context and will not be aware of the changes the other one has made.  

**To Reproduce**
Reproducing reliably can be difficult but by turning the publish intervals for telemetry in isar-robot to 1 second for all publishers you will see this behavior.

Once that has been done run a mission and observe what happens to the status whenever a mission finishes. It may also happen if you set the status directly through the API although it is not as often. 

**Expected behavior**
The robot object should be locked such that there is no possibility of reading and writing to it simultaneously across threads. We have to look at how we do read/writes throughout the system.

For example it may be smart to redesign a lot of our endpoints such that activities which require reading the robot don't read at the beginning and then update at the end but are smaller and closer to atomic actions. 

**Screenshots**

aestene:
Closes #1089

### DIFF
--- a/backend/api/EventHandlers/IsarConnectionEventHandler.cs
+++ b/backend/api/EventHandlers/IsarConnectionEventHandler.cs
@@ -94,7 +94,11 @@ namespace Api.EventHandlers
             timer.Start();
 
             if (_isarConnectionTimers.TryAdd(robot.IsarId, timer)) { _logger.LogInformation("Added new timer for ISAR '{IsarId}' ('{RobotName}')", robot.IsarId, robot.Name); }
-            else { _logger.LogWarning("Failed to add new timer for ISAR '{IsarId}' ('{RobotName})'", robot.IsarId, robot.Name); }
+            else
+            {
+                _logger.LogWarning("Failed to add new timer for ISAR '{IsarId}' ('{RobotName})'", robot.IsarId, robot.Name);
+                timer.Close();
+            }
         }
 
         private async void OnTimeoutEvent(IsarRobotHeartbeatMessage robotHeartbeatMessage)
--- a/backend/api/Services/RobotService.cs
+++ b/backend/api/Services/RobotService.cs
@@ -88,10 +88,7 @@ namespace Api.Services
         public async Task<Robot?> Delete(string id)
         {
             var robot = await GetRobotsWithSubModels().FirstOrDefaultAsync(ev => ev.Id.Equals(id));
-            if (robot is null)
-            {
-                return null;
-            }
+            if (robot is null) { return null; }
 
             _context.Robots.Remove(robot);
             await _context.SaveChangesAsync();
--- a/backend/api.test/Mocks/RobotControllerMock.cs
+++ b/backend/api.test/Mocks/RobotControllerMock.cs
@@ -2,17 +2,16 @@
 using Api.Services;
 using Microsoft.Extensions.Logging;
 using Moq;
-
 namespace Api.Test.Mocks
 {
     internal class RobotControllerMock
     {
-        public Mock<IIsarService> IsarServiceMock;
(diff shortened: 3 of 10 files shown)


backend/api/EventHandlers/IsarConnectionEventHandler.cs
```diff
@@ -94,7 +94,11 @@ private void AddTimerForRobot(IsarRobotHeartbeatMessage isarRobotHeartbeat, Robo
             timer.Start();
 
             if (_isarConnectionTimers.TryAdd(robot.IsarId, timer)) { _logger.LogInformation("Added new timer for ISAR '{IsarId}' ('{RobotName}')", robot.IsarId, robot.Name); }
-            else { _logger.LogWarning("Failed to add new timer for ISAR '{IsarId}' ('{RobotName})'", robot.IsarId, robot.Name); }
+            else
+            {
+                _logger.LogWarning("Failed to add new timer for ISAR '{IsarId}' ('{RobotName})'", robot.IsarId, robot.Name);
+                timer.Close();
```
oysand, line 93: How does the timer close for the if statement above

backend/api/Services/RobotService.cs
```diff
@@ -88,20 +97,66 @@ public async Task<Robot> Update(Robot robot)
...
             var robot = await GetRobotsWithSubModels().FirstOrDefaultAsync(ev => ev.Id.Equals(id));
+            if (robot is null) { return null; }
+
+            _context.Robots.Remove(robot);
+            await _context.SaveChangesAsync();
+
+            return robot;
+        }
+
+        public async Task<Robot> Update(string id, string? name = null, string? isarId = null, RobotModel? robotModel = null, string? serialNumber = null,
```
oysand, line 108: How do we set a value back to null?

backend/api/appsettings.Development.json
```diff
@@ -43,6 +43,6 @@
     "AccountName": "flotilladevsa"
   },
   "Database": {
-    "UseInMemoryDatabase": true
+    "UseInMemoryDatabase": false
```
oysand, line 46: Set back to true

backend/api/Services/RobotService.cs
```diff
@@ -88,20 +97,66 @@ public async Task<Robot> Update(Robot robot)
...
         }
 
-        private IQueryable<Robot> GetRobotsWithSubModels()
+        private IQueryable<Robot> GetRobotsWithSubModels(bool lockWithSemaphore = true)
         {
-            return _context.Robots.Include(r => r.VideoStreams).Include(r => r.Model).Include(r => r.CurrentArea);
+            IIncludableQueryable<Robot, Area?> robots;
+            if (lockWithSemaphore)
+            {
+                updateRobotSemaphore.WaitOne();
```
oysand, line 152: Consider removing this if it is not needed for the read step

backend/api/EventHandlers/IsarConnectionEventHandler.cs
aestene: Closes on the timeout event.

backend/api/Services/RobotService.cs
aestene: Removed that.

Christdej (approved):
LGTM

PR #1093, opened 2023-10-30
Update npm packages

Christdej:
Target JS language version is updatede to be able to run "await" from index.ts to accommodate newest msal version
Fixes
https://github.com/equinor/robotics-infrastructure/issues/334


FuzzyReality (approved):
As good as it gets.

Issue #1108, opened 2023-11-02
Align placeholders for ongoing missions and mission queue


Eddasol:
**Describe the improvement you would like to see**
Align the placeholders for ongoing missions and mission queue

Currently:
![Aligning](https://github.com/equinor/flotilla/assets/33518988/aa8f3db7-a3b5-4cdf-b038-27285a75f99e)

Desired solution:
![image](https://github.com/equinor/flotilla/assets/33518988/e0bf8f85-49e6-4e58-bf47-d568de9917d9)

**How will this change existing functionality?**
Will look better :)

**How will this improvement affect the current Threat Model?**
N/A

Issue #901, opened 2023-07-19
Consider fixing the pause and stop button 


anetteu:
**Describe the new feature you would like to see**
Consider to fix the pause and stop button such that they are always available and always in the same place also if the user is scrolling on the ongoing mission page. Consider if the buttons should be available also on other pages than the ongoing mission page. Please see the attached screenshot of how the buttons are placed today.
![IMG_0022](https://github.com/equinor/flotilla/assets/21355793/8cc72bd8-c6bb-4d84-9d77-0db7ed19d49a)

**Describe the solution you'd like**
If you have an idea of how to implement the feature please describe it here.

**How will this feature affect the current Threat Model?**
Leave blank for maintainers to fill out if you are uncertain about this.

anetteu:
Such a fxed stop button should stop/freeze all missions/robot activity. If only the current mission is stopped, the next one in the queue will immediately start which is probably not what is the intention of pressing the stop button.

PR #1117, opened 2023-11-04
Move stop all robots button

anetteu:
Solves #1108 

I'll suggest to place the stop button above the mission overview as shown on the attached screenshots, but open for other opinions. A suggestion in the future is also to fix the stop button such that it is always available in the same position on the screen, as suggested in #901.

Screenshot from 2023-11-04 17-10-55
Screenshot from 2023-11-04 17-11-32


mrica-equinor (approved):
LGTM  :)

PR #1119, opened 2023-11-06
Remove Populate Area function in tests

Christdej:


backend/api.test/Client/MissionTests.cs
```diff
@@ -0,0 +1,432 @@
...
+
+            // Arrange - Areas
+            string areaUrl = "/areas";
+            var areaResponse = await _client.GetAsync(areaUrl);
+            Assert.True(areaResponse.IsSuccessStatusCode);
+            var areaResponses = await areaResponse.Content.ReadFromJsonAsync<List<AreaResponse>>(_serializerOptions);
+            Assert.True(areaResponses != null);
+            var area = areaResponses[0];
+            string areaName = area.AreaName;
+            string installationCode = area.InstallationCode;
```
tsundvoll, line 147: Is it possible / would it make sense to reuse some code here? I see this snippet is recurring a few places

Christdej: It is possible yes, but I was not if I wanted to do that or not, to make the tests as atomic as possible and to avoid the problem we had with the PopulateAreaDb where several tests tried to add the same Area to the DB and we got unique constraint conflicts. But I am open to ideas.

tsundvoll: I see, good enough reason for me

tsundvoll (approved):

aestene (approved):
LGTM

I agree with Thomas however I think it's better to merge this PR now as the purpose of the PR was not to improve the internal parts of the tests themselves but to refactor and remove the populate area function.

We should do those improvements too but avoid scope creep in this PR.

PR #1125, opened 2023-11-07
Fix localization map

Eddasol:
Fixing mismatch between expected area in frontend and sent areaResponse in backend


oysand (approved):
LGTM

PR #1126, opened 2023-11-07
Refactor MissionScheduling

aestene:


backend/api/EventHandlers/MissionEventHandler.cs
```diff
@@ -154,12 +151,11 @@ private async void OnEmergencyButtonPressedForRobot(object? sender, EmergencyBut
                 return;
             }
 
-            await MissionSchedulingService.FreezeMissionRunQueueForRobot(e.RobotId);
+            try { await MissionSchedulingService.FreezeMissionRunQueueForRobot(e.RobotId); }
+            catch (RobotNotFoundException) { return; }
```
mrica-equinor, line 155: I am confused, we read the robot from the database and it wasn't null, why do we need this catch?

andchiind: With the current db inconsistencies it might be possible for the robot to be deleted during the execution of this event handler, so there's no harm in having this try/catch in my opinion.

backend/api/EventHandlers/MissionScheduling.cs
```diff
@@ -232,12 +232,10 @@ public void TriggerRobotAvailable(RobotAvailableEventArgs e)
         public async Task<bool> TheSystemIsAvailableToRunAMission(string robotId, MissionRun missionRun)
         {
             var robot = await _robotService.ReadById(robotId);
-            if (robot == null)
-            {
-                _logger.LogError("Robot with ID: {RobotId} was not found in the database", robotId);
-                return false;
-            }
-            return await TheSystemIsAvailableToRunAMission(robot, missionRun);
+            if (robot != null) { return await TheSystemIsAvailableToRunAMission(robot, missionRun); }
```
oysand: Revert

oysand (approved):
LGTM

PR #1138, opened 2023-11-08
Remove duplicate scoped event handler

aestene:
The additional scoped service which was used to handle the RobotAvailable event caused events to sometimes be handled by two different objects at the same time.


Afonso-2403 (approved):
lgtm

PR #1142, opened 2023-11-09
Fix confirmation button asset page

### LINKED ISSUE #1124: "Confirm installation" is disabled when the input is valid
**Describe the bug**
The "Confirm installation" button is initially disabled on assetSelectionPage. Even if the previously selected value is valid.

**To Reproduce**
1. Go to assetSelectionPage
2. If the selected value is invalid, select a valid one and repeat step 1
3. Note that the button "Confirm installation" is disabled
4. Click on the dropdown and note that the button is active

**Expected behavior**
The button should not be disabled if the value is valid, and it shouldn't be necessary to re-select the same value before confirming

**Screenshots**
Asset selection page with "Confirm installation" disabled - even though the selected value is valid
image

Asset selection page after clicking the dropdown. The value is not changed but the button becomes active
image

mrica-equinor:
Closes #1124

### DIFF
--- a/frontend/src/components/Pages/AssetSelectionPage/AssetSelectionPage.tsx
+++ b/frontend/src/components/Pages/AssetSelectionPage/AssetSelectionPage.tsx
@@ -117,10 +117,14 @@ function InstallationPicker() {
                     placeholder={TranslateText('Select installation')}
                     onOptionsChange={({ selectedItems }) => {
                         const selectedName = selectedItems[0]
-                        setSelectedInstallation(selectedName)
+                        validateInstallation(selectedName)
+                            ? setSelectedInstallation(selectedName)
+                            : setSelectedInstallation('')
                     }}
                     onInput={(e: React.ChangeEvent<HTMLInputElement>) => {
-                        setSelectedInstallation(e.target.value ?? '')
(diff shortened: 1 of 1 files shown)


frontend/src/components/Pages/AssetSelectionPage/AssetSelectionPage.tsx
```diff
@@ -137,7 +134,7 @@ function InstallationPicker() {
             </BlockLevelContainer>
             <Button
                 onClick={() => switchInstallation(selectedInstallation)}
-                disabled={!validateInstallation(selectedInstallation)}
+                disabled={!selectedInstallation}
```
andchiind, line 137: Without this check we can accept any non-empty string 
image

andchiind (changes requested):

frontend/src/components/Pages/AssetSelectionPage/AssetSelectionPage.tsx
mrica-equinor: Hum i dind't try that indeed, thanks!

andchiind (approved):
I was sadly not able to break it

PR #1145, opened 2023-11-09
Prevent all robots from showing while loading

Eddasol:
This PR will also hide all robots if invalid/empty installation


mrica-equinor (approved):
LGTM :)

PR #1150, opened 2023-11-09
Add InspectionFindings endpoint to Flotilla API

UsamaEquinorAFK:
Using IsarStepId to add InspectionFindings under Inspections which is owned by Tasks which is owned by MissionRun. 

The above statement is now a lie, they are no longer owned.


backend/api/Controllers/InspectionFindingsController.cs
```diff
@@ -0,0 +1,97 @@
...
+        private readonly IInspectionService _inspectionService;
+        private readonly ILogger<InspectionFindingsController> _logger;
+        public InspectionFindingsController(
+            ILogger<InspectionFindingsController> logger,
+            IInspectionService inspectionService
+        )
+        {
+            _logger = logger;
+            _inspectionService = inspectionService;
+
```
aestene, line 22:
```suggestion
```

backend/api/Controllers/InspectionFindingsController.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        /// <summary>
+        /// Add a new inspection finding - need isarStepId to run this
```
aestene, line 26:
```suggestion
        /// Associate a new inspection finding with the inspection corresponding to isarStepId
```

backend/api/Controllers/InspectionFindingsController.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// </summary>
+        /// <remarks>
+        /// <para> This query adds a new area to the database </para>
```
aestene, line 29:
```suggestion
```

backend/api/Controllers/InspectionFindingsController.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// </remarks>
+        [HttpPost("add-findings")]
+        [Authorize(Roles = Role.Admin)]
+        [ProducesResponseType(typeof(AreaResponse), StatusCodes.Status201Created)]
```
aestene, line 33: `AreaResponse` is not the correct return type for this request. Maybe an `Inspection` or creating an `InspectionFindingResponse` would be better?

backend/api/Controllers/InspectionFindingsController.cs
```diff
@@ -0,0 +1,97 @@
...
+        [ProducesResponseType(StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status409Conflict)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(StatusCodes.Status500InternalServerError)]
```
aestene, line 38: These status codes should be reflected by the actual status codes that might be returned. Remember that this is the documentation that will be displayed in the swagger and should be correct. 

For example, you return OK on line 48 which is a 200 not 201.

backend/api/Controllers/InspectionFindingsController.cs
```diff
@@ -0,0 +1,97 @@
...
+        public async Task<ActionResult<InspectionFindings>> AddFindings([FromBody] InspectionFindingsQuery inspectionFinding)
+        {
+            _logger.LogInformation("Updating inspection findings for inspection with isarStepId '{inspectionFinding.IsarStepId}'", inspectionFinding.IsarStepId);
```
aestene, line 41:
```suggestion
            _logger.LogInformation("Updating inspection findings for inspection with isarStepId '{Id}'", inspectionFinding.IsarStepId);
```

backend/api/Services/InspectionService.cs
```diff
@@ -0,0 +1,97 @@
...
+                }
+                else
+                {
+                    return null;
+                }
+            }
+            else
+            {
+                return null;
+            }
```
aestene, line 67:
```suggestion
            else { return null; }
```

backend/api/Services/InspectionService.cs
```diff
@@ -0,0 +1,97 @@
...
+                    _context.MissionRuns.Update(missionRun);
+                    // Save the changes to the database
+                    await _context.SaveChangesAsync();
+
+                    return inspection;
+                }
+                else
+                {
+                    return null;
+                }
```
aestene, line 62:
```suggestion
                else { return null; }
```

backend/api/Controllers/InspectionFindingsController.cs
UsamaEquinorAFK: I believe this is addressed now, hence resolving.

backend/api/Controllers/InspectionFindingsController.cs
UsamaEquinorAFK: I believe this is addressed now, hence resolving.

backend/api/Controllers/InspectionFindingsController.cs
UsamaEquinorAFK: I believe this is addressed now, hence resolving.

backend/api/Controllers/InspectionFindingsController.cs
UsamaEquinorAFK: I believe this is addressed now, hence resolving.

UsamaEquinorAFK: Maybe not, why are we doing this ?

backend/api/Services/InspectionService.cs
UsamaEquinorAFK: I believe this is addressed now, hence resolving.

backend/api/Services/InspectionService.cs
UsamaEquinorAFK: I believe this is addressed now, hence resolving.

backend/api/Controllers/InspectionFindingsController.cs
UsamaEquinorAFK: I believe this is addressed now, hence resolving.

backend/api/Controllers/InspectionFindingsController.cs
aestene: We are doing it for readability. There is not need to specify the entire thing within the string as it is automatically associated with the parameter you put afterwards. Therefore just `Id` or `IsarStepId` is a more suitable name.

UsamaEquinorAFK: Fixed now. Hence resolving.

backend/api/Database/Models/InspectionFindings.cs
```diff
@@ -1,22 +1,46 @@
...
+using Api.Controllers.Models;
 #pragma warning disable CS8618
 namespace Api.Database.Models
 {
-    [Owned]
     public class InspectionFindings
     {
+        [Key]
+        [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
+        public string Id { get; set; }
```
Christdej, line 11: In the Inspection models, I think we should have a reference to this such that an Inspection can contain a List<InspectionFindings>

backend/api/Database/Models/InspectionFindings.cs
```diff
@@ -1,22 +1,46 @@
...
 
-        public string InspectionId { get; set; }
-
-        public string FindingsTag { get; set; }
-
+        public string IsarStepId { get; set; }
+
+        public string Findings { get; set; }
+
+        public string MissionRunId { get; set; }
```
Christdej, line 23: If we have a link from Inspection, I dont think we need MissionRunId

backend/api/Program.cs
```diff
@@ -73,7 +73,7 @@
 builder.Services.AddScoped<ITaskDurationService, TaskDurationService>();
 builder.Services.AddScoped<IPoseTimeseriesService, PoseTimeseriesService>();
 builder.Services.AddScoped<ILastMissionRunService, LastMissionRunService>();
-
+builder.Services.AddScoped<IInspectionService, InspectionService>();
```
Christdej, line 76: This should be InspectionFindingsService right?

backend/api/Controllers/InspectionFindingsController.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// </remarks>
+        [HttpPost("add-findings")]
+        [Authorize(Roles = Role.Admin)]
+        [ProducesResponseType(typeof(InspectionFindingsResponse), StatusCodes.Status200OK)]
+        [ProducesResponseType(StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status409Conflict)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+        public async Task<ActionResult<InspectionFindings>> AddFindings([FromBody] InspectionFindingsQuery inspectionFinding)
```
Christdej, line 38: Lines 32 says it produces output of type InspectionFindingsResponse while the return type on line 38 says it returns InspectionFindings, this is incorrect yeah?

Christdej (changes requested):

backend/api/Program.cs
UsamaEquinorAFK: No we are using inspection service for both Inspection and InspectionFindings so far, as we only need to have an inspection finding when we have an inspection the current approach requires less code and not an additional file.

backend/api/Database/Models/InspectionFindings.cs
UsamaEquinorAFK: Removed, it was added earlier when we have owned entities. We dont need it anymore.

backend/api/Controllers/InspectionFindingsController.cs
UsamaEquinorAFK: Yes it was incorrect. Fixed now I believe.

backend/api/Database/Models/InspectionFindings.cs
UsamaEquinorAFK: But inspection model already contains an inspectionFindingsList

backend/api/Program.cs
UsamaEquinorAFK: Removed redundant line, hence resolved.

backend/api/Database/Models/InspectionFindings.cs
UsamaEquinorAFK: Agreed that we have them as list.

backend/api/Controllers/InspectionFindingsController.cs
```diff
@@ -0,0 +1,95 @@
...
+        [ProducesResponseType(StatusCodes.Status409Conflict)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+        public async Task<ActionResult<InspectionFindings>> GetInspections([FromRoute] string id)
+        {
+            _logger.LogInformation("Get inspection by ID '{inspectionFinding.InspectionId}'", id);
+            try
+            {
+                var inspection = await _inspectionService.ReadByIsarStepId(id);
```
Christdej, line 78: This returns an "Inspection" and in the ProducesResponse and return type of function you say "AreaResponse" and "InspectionFindings", probably a mistake?

backend/api/Controllers/InspectionFindingsController.cs
```diff
@@ -0,0 +1,95 @@
...
+                if (inspection != null)
+                {
+                    return Ok(inspection);
+                }
+
+            }
+            catch (Exception e)
+            {
+                _logger.LogError(e, "Error while finding an inspection with inspection id '{id}'", id);
+                throw;
```
Christdej, line 88: Should return InternalServerError or similar

backend/api/Controllers/InspectionFindingsController.cs
```diff
@@ -0,0 +1,95 @@
...
+            }
+            return NotFound();
```
Christdej, line 90: Probably add a message to the NotFound

backend/api/Controllers/InspectionFindingsController.cs
```diff
@@ -0,0 +1,95 @@
...
+        [ProducesResponseType(StatusCodes.Status409Conflict)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+        public async Task<ActionResult<InspectionFindings>> AddFindings([FromBody] InspectionFindingsQuery inspectionFinding)
+        {
+            _logger.LogInformation("Updating inspection findings for inspection with isarStepId '{Id}'", inspectionFinding.IsarStepId);
+            try
+            {
+                var inspection = await _inspectionService.ReadByIsarStepId(inspectionFinding.IsarStepId);
```
Christdej, line 43: This returns an "Inspection" and not an "InspectionFindings" as stated in the return type, should perhaps return the list of inspectionfindings on a isarStepId

Christdej (changes requested):

backend/api/Controllers/InspectionFindingsController.cs
UsamaEquinorAFK: Yup my mistake, Fixed now. Thanks.

UsamaEquinorAFK: Done. Resolved.

backend/api/Controllers/InspectionFindingsController.cs
UsamaEquinorAFK: Done now.

backend/api/Controllers/InspectionFindingsController.cs
UsamaEquinorAFK: Done.

backend/api/Controllers/InspectionFindingsController.cs
UsamaEquinorAFK: No, this is for fetching the inspection object based on the isarStepId and then we add inspectionfindings to this object.

UsamaEquinorAFK: I think it should return inspection as it is required in the next line of code.

UsamaEquinorAFK: The check on the next line ois based on the premise that there should not be a null inspection to avoid adding any Rogue inspection findings which are not associated with any inspections.

UsamaEquinorAFK: Done now. Hence resolving. We agreed that we return the finding for the post call and inspection for the getcall.

backend/api/Controllers/InspectionFindingsController.cs
```diff
@@ -0,0 +1,97 @@
...
+                if (inspection != null)
+                {
+                    inspection = await _inspectionService.AddFindings(inspectionFinding);
+                    return Ok(inspection?.InspectionFindings);
```
Christdej, line 47:
```suggestion
                    return Ok(inspection.InspectionFindings);
```
No need for a question mark here I think, as we have checked that inspection is not null.

backend/api/Controllers/InspectionFindingsController.cs
```diff
@@ -0,0 +1,97 @@
...
+                }
+
+            }
+            catch (Exception e)
+            {
+                const string Message = "Error while adding findings to inspection with IsarStepId '{Id}'";
+                _logger.LogError(e, "Message", inspectionFinding.IsarStepId);
```
Christdej, line 54: Will this not just log "Message"?

backend/api/Controllers/InspectionFindingsController.cs
```diff
@@ -0,0 +1,97 @@
...
+                return StatusCode(StatusCodes.Status500InternalServerError, Message);
```
Christdej, line 55: Dont think we have to include this message to the user

backend/api/Controllers/InspectionFindingsController.cs
```diff
@@ -0,0 +1,97 @@
...
+                if (inspection != null)
+                {
+                    return Ok(inspection);
+                }
+
+            }
+            catch (Exception e)
+            {
+                const string Message = "Error while finding an inspection with inspection id '{id}'";
+                _logger.LogError(e, "Message", id);
```
Christdej, line 89: This will also just log "Message", look at other controllers to see how to log a message with a variable

backend/api/Database/Models/InspectionFindings.cs
```diff
@@ -1,22 +1,42 @@
...
 namespace Api.Database.Models
 {
-    [Owned]
     public class InspectionFindings
     {
+        [Key]
+        [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
+        public string Id { get; set; }
 
         public string RobotName { get; set; }
```
Christdej, line 13: Not sure we need a string RobotName here. The inspection, task, missionRun already has information about which robot ran the mission I think.

backend/api/Database/Models/InspectionFindings.cs
```diff
@@ -1,22 +1,42 @@
...
 
         public string InspectionDate { get; set; }
```
Christdej, line 15: Should perhaps be a DateTime object

Christdej (changes requested):

backend/api/Controllers/InspectionFindingsController.cs
UsamaEquinorAFK: Nope it complains and throws a warning otherwise.

UsamaEquinorAFK: I will see again.

Christdej: Because you already have inspection on line 43, then you override it on line 46

backend/api/Controllers/InspectionFindingsController.cs
UsamaEquinorAFK: Yes my bad. Changed now.

backend/api/Controllers/InspectionFindingsController.cs
UsamaEquinorAFK: Yes it will, now fixed.

backend/api/Database/Models/InspectionFindings.cs
UsamaEquinorAFK: Done.

backend/api/Database/Models/InspectionFindings.cs
UsamaEquinorAFK: Yes I think it should be. Also highlight the correct parameter because this got me thinking why should I have "ID" as date time object :P . Now the InspectionDate is datetime object. =) .

backend/api/Controllers/InspectionFindingsController.cs
UsamaEquinorAFK: This is handled in the lattest ammends.

backend/api/Services/InspectionService.cs
```diff
@@ -61,5 +66,36 @@ private IQueryable<Inspection> GetInspections()
         {
             return _context.Inspections.Include(inspection => inspection.InspectionFindings);
         }
+
+        public async Task<MissionRun?> GetMissionRunFromIsarStepId(string isarStepId)
+        {
+            return await _context.MissionRuns.Where(m => m.Tasks.SelectMany(t => t.Inspections)
+            .Any(i => i.IsarStepId == isarStepId))
+            .FirstOrDefaultAsync();
+        }
```
Christdej, line 75: I cannot see where we use this function?

backend/api/Migrations/20231109194856_MakeInspectionNotOwned.cs
Christdej, line 1: Why are these files changed?

backend/api/Migrations/20231109195524_MakeMissionTaskNotOwned.cs
Christdej, line 1: Why are these files changed?

Christdej (changes requested):

Christdej (changes requested):
Fix the previous comments

backend/api/Services/InspectionService.cs
UsamaEquinorAFK: Removed, redundant now.

backend/api/Migrations/20231109194856_MakeInspectionNotOwned.cs
UsamaEquinorAFK: not changed now

backend/api/Migrations/20231109195524_MakeMissionTaskNotOwned.cs
UsamaEquinorAFK: not changed now

Christdej (approved):
LGTM

UsamaEquinorAFK:
Just rebased afterwards now new work added.

PR #1151, opened 2023-11-09
Fix status display

Eddasol:


frontend/src/components/Pages/FrontPage/MissionOverview/MissionStatusDisplay.tsx
```diff
@@ -42,6 +43,16 @@ export function displayIcon(status: MissionStatus) {
...
+            {displayIcon(status)}
+            <Typography>{TranslateText(status)}</Typography>
+        </StyledStatusIcon>
+    )
+}
+
+export function MissionStatusDisplayWithHeader({ status }: StatusProps) {
     const { TranslateText } = useLanguageContext()
     return (
         <StyledStatusDisplay>
```
anetteu, line 58: Good idea making a new function for the version with header! You can call the MissionStatusDisplay in this function to avoid duplicate code :)

anetteu (changes requested):

anetteu (approved):
LGTM 😄

PR #1152, opened 2023-11-09
Stop mission if robot is idle

Afonso-2403:
If the robot is reported as idle from ISAR, stopping a mission should not crash


mrica-equinor (approved):
LGTM

Afonso-2403:
This solution should be temporary, created an issue in [ISAR](https://github.com/equinor/isar/issues/503) to address this problem

PR #1155, opened 2023-11-09
Make `Inspection` and `MissionTask` not owned entities

### LINKED ISSUE #1141: Mission status is overwritten when setting to paused
**Describe the bug**
There are occasions when the mission status seems to behave weirdly and it is set to faulty values. For example that it is set to `Failed` but might be set to `Ongoing` afterwards.

**To Reproduce**
The easiest way to reproduce is by pausing a missions. Note that you have to be connected to the dev database as this does not happen for the in-memory database.
- Start a mission with Placebot
- Pause the mission
- Observe that the resume button flashes but then goes back to pause
- Note that the status goes to Ongoing again

**Expected behavior**
The robot should display as paused correctly and the status of the mission should be `Paused`. It is likely due to a timing conflict where the mission is being overwritten to the previous status. 

**Screenshots**

aestene:
As these were owned by `MissionRun` the only method of updating the status of each of them was to update the entire `MissionRun`. As this happens in three parallel event handlers these were overwriting eachother. 

Closes #1141

### DIFF
--- a/backend/api/EventHandlers/MqttEventHandler.cs
+++ b/backend/api/EventHandlers/MqttEventHandler.cs
@@ -206,15 +206,9 @@ namespace Api.EventHandlers
                 return;
             }
 
-            MissionRun? flotillaMissionRun;
+            MissionRun flotillaMissionRun;
             try { flotillaMissionRun = await missionRunService.UpdateMissionRunStatusByIsarMissionId(isarMission.MissionId, status); }
             catch (MissionRunNotFoundException) { return; }
-
-            if (flotillaMissionRun is null)
-            {
-                _logger.LogError("No mission found with ISARMissionId '{IsarMissionId}'. Could not update status to '{Status}'", isarMission.MissionId, status);
-                return;
-            }
 
             _logger.LogInformation(
                 "Mission '{Id}' (ISARMissionID='{IsarMissionId}') status updated to '{Status}' for robot '{RobotName}' with ISAR id '{IsarId}'",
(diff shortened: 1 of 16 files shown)


backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -206,7 +206,10 @@ private async void OnMissionUpdate(object? sender, MqttReceivedArgs mqttArgs)
                 return;
             }
 
-            var flotillaMissionRun = await missionRunService.UpdateMissionRunStatusByIsarMissionId(isarMission.MissionId, status);
+            MissionRun? flotillaMissionRun;
+            try { flotillaMissionRun = await missionRunService.UpdateMissionRunStatusByIsarMissionId(isarMission.MissionId, status); }
+            catch (MissionRunNotFoundException) { return; }
+
             if (flotillaMissionRun is null)
```
Afonso-2403, line 213: Shouldn't this logging be inside the above catch block, in case no mission was found?

backend/api/Services/InspectionService.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        public async Task<Inspection> UpdateInspectionStatus(string isarStepId, IsarStepStatus isarStepStatus)
+        {
+            var inspection = await ReadByIsarStepId(isarStepId);
+            if (inspection is null)
+            {
+                string errorMessage = $"Inspection with ID {isarStepId} could not be found";
+                _logger.LogError("{Message}", errorMessage);
+                throw new InspectionNotFoundException(errorMessage);
+            }
```
tsundvoll, line 38: Can we do the null-check inside the `ReadByIsarStepId` function?

backend/api/EventHandlers/MqttEventHandler.cs
aestene: Instead I removed the nullable on mission run here and changed the function as it can no longer return null. This means we could remove the check altogether.

backend/api/Services/InspectionService.cs
aestene: Ref. discussion on Teams.

Afonso-2403 (approved):
Let's migrate!

backend/api/Services/InspectionService.cs
tsundvoll: Ref. our discussion, we figured this is probably the C# way of doing it

PR #1157, opened 2023-11-10
Fix schedule mission button on inpection page

### LINKED ISSUE #1123: Text is larger than button when scheduling trough inspection section
**Describe the bug**
If scheduling missions trough inspection plan, when there are already scheduled missions and language is Norwegian, the text becomes bigger than the button

**To Reproduce**
1. Sett language to Norwegian
2. Queue a mission on a deck that has several missions
3. Press "Queue the missions" on the same deck
4. Result should be as in screenshot

**Expected behavior**
Text should be inside button

**Screenshots**
image

mrica-equinor:
Closes #1123


andchiind (approved):
LGTM

PR #1159, opened 2023-11-12
Improve concurrency behavior of Flotilla

aestene:


backend/api/Services/MissionSchedulingService.cs
```diff
@@ -210,6 +210,21 @@ public void TriggerRobotAvailable(RobotAvailableEventArgs e)
...
+            foreach (string missionRunId in interruptedMissionRunIds)
+            {
+                var missionRun = await _missionRunService.ReadById(missionRunId);
+                if (missionRun is null)
+                {
+                    _logger.LogWarning("{Message}", $"Interrupted mission run with Id {missionRunId} could not be found");
+                    continue;
+                }
+
+                missionRun.Status = MissionStatus.Pending;
```
andchiind: Since we have a SetCurrentMissionId in robotService, we could maybe consider having SetCurrentMissionStatus in missionRunService. But if this code already works consistently then this is not needed I suppose.

backend/api/Services/MissionSchedulingService.cs
```diff
@@ -221,8 +221,20 @@ private async Task MoveInterruptedMissionsToQueue(IEnumerable<string> interrupte
                     continue;
                 }
 
-                missionRun.Status = MissionStatus.Pending;
-                await _missionRunService.Update(missionRun);
+                var newMissionRun = new MissionRun
```
andchiind, line 213: Could this change maybe be rebased into the first commit?

backend/api/Controllers/RobotController.cs
```diff
@@ -732,10 +736,14 @@ public async Task<ActionResult> ResumeMission([FromRoute] string robotId)
...
 
-            robot.Status = RobotStatus.Busy;
-            robot.CurrentMissionId = missionRun.Id;
-            robot.CurrentArea = area;
-            await _robotService.Update(robot);
+            try
+            {
+                await _robotService.UpdateRobotStatus(robot.Id, RobotStatus.Busy);
+                await _robotService.UpdateCurrentMissionId(robot.Id, missionRun.Id);
+                await _robotService.UpdateCurrentArea(robot.Id, area);
```
andchiind, line 743: Is it safer to do in 3 requests instead of 1?

backend/api/Services/RobotService.cs
```diff
@@ -127,6 +134,15 @@ public async Task<Robot> Update(Robot robot)
             return robot;
         }
 
+        public async Task<IList<Robot>> ReadLocalizedRobotsForInstallation(string installationCode)
```
andchiind, line 137: This looks good, but if we keep adding more service functions like these then we could perhaps make a generic filter function like we did for missionRun and missionDefinition

andchiind (approved):
LGTM, only minor comments and questions, but it could be good to answer the one about using 3 requests instead of 1 before merging.

backend/api/Controllers/RobotController.cs
aestene: I think it would be better to do in one as we have additional database reads and writes although I don't think the difference is large as we have the semaphore now. I didn't want to make too many combination functions of type `UpdateRobotStatusAndCurrentMissionId...`.

For now will leave it like this and then we can consider improving it.

backend/api/Services/RobotService.cs
aestene: It might be relevant although the filter functions for mission run and mission definition are more linked to there being a lot of them and there is a need to be able to search specifically (for example by date etc.). As long as there are relatively few read functions for robots I think it's fine.

In my opinion this is also a lot more readable than having one function which we send a filter into. Easier to have a good, readable function name this way.

backend/api/Services/MissionSchedulingService.cs
aestene: Yes, I agree. Although not added as part of this PR.

backend/api/Services/MissionSchedulingService.cs
aestene: I agree. However, the merge conflicts that arose were sufficiently difficult that I decided against it.

PR #1163, opened 2023-11-13
Add new analytics user to the broker

tsundvoll:


UsamaEquinorAFK:
Just a question, is the password also stored somewhere on the keyvault ? :)

UsamaEquinorAFK (approved):
LGTM

PR #1177, opened 2023-11-14
Add localization verification on mission schedule

Eddasol:


aestene:

aestene (approved):
LGTM

I have tested it before the weekend but if you want someone to look into the detailed Typescript you should get another review.

frontend/src/utils/DeckMapView.tsx
```diff
@@ -0,0 +1,141 @@
...
+                    })
+                    .catch(() => {
+                        setImageObjectURL(NoMap)
+                    })
+            })
+            .catch(() => {
+                setMapMetadata(undefined)
+                setImageObjectURL(NoMap)
+            })
+        //.catch((e) => handleError(e))
```
UsamaEquinorAFK, line 88: Left this comment by mistake ?

backend/api/Controllers/DeckController.cs
```diff
@@ -220,5 +222,68 @@ public async Task<ActionResult<DeckResponse>> DeleteDeck([FromRoute] string id)
                 return NotFound($"Deck with id {id} not found");
             return Ok(new DeckResponse(deck));
         }
+
+        /// <summary>
+        /// Gets map metadata for localization poses belonging to deck with specified id
+        /// </summary>
+        [HttpGet]
+        [Authorize(Roles = Role.Any)]
```
UsamaEquinorAFK, line 230: Should it not be a user and then admin can do this automatically or does it not works that way ?

Eddasol: Seems that we have very mixed use of 'any' and 'user', and that 'any' is most used. Suggest that we make a card on being more consistent and solve that separately

UsamaEquinorAFK (approved):
lgtm

PR #1178, opened 2023-11-14
Fix filtering of missions by installation

Eddasol:
This is a partial fix. Also need to compare to installation filtered missions when deciding if placeholder should show.


eivindsjovold (approved):
LGTM

Issue #1060, opened 2023-10-13
Add area column and rerun button to mission history table


Eddasol:
**Describe the improvement you would like to see**
Add area column and "add to queue"-button to table in mission history. 

Design inspiration:
![image](https://github.com/equinor/flotilla/assets/33518988/24b26bef-ac4b-47fb-b941-5a8545b170f2)

**How will this change existing functionality?**

**How will this improvement affect the current Threat Model?**
N/A

prasm313:
Area column made in PR #1193.

prasm313:
Rerun button added in PR #1327.

PR #1193, opened 2023-11-16
Add missionRunResponse model to backend

prasm313:
MissionRunResponse.cs was made in addition to adjustments to missionRunController.cs and missionRunService.cs. 

Related to this, an area column was added to the Mission History Page (as suggested in #1060).

### DIFF
--- a/backend/api/Controllers/Models/MissionRunResponse.cs
+++ b/backend/api/Controllers/Models/MissionRunResponse.cs
@@ -1,0 +1,74 @@
+using Api.Database.Models;
+using System.Text.Json.Serialization;
+namespace Api.Controllers.Models
+{
+    public class MissionRunResponse
+    {
+        public string Id { get; set; }
+
+        public string? MissionId { get; set; }
+
+        public string? IsarMissionId { get; set; }
+
+        public string Name { get; set; }
+
+        public string? Description { get; set; }
+
+        public string? StatusReason { get; set; }
+
+        public string? Comment { get; set; }
+
+        public string InstallationCode { get; set; }
+
+        public AreaResponse? Area { get; set; }
+
+        public virtual RobotResponse Robot { get; set; }
+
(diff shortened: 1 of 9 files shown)


backend/api/Controllers/Models/MissionRunResponse.cs
```diff
@@ -0,0 +1,75 @@
+﻿using System.Text.Json.Serialization;
+using Api.Database.Models;
+#pragma warning disable CS8618
```
Eddasol, line 3: Which warning are we suppressing?

frontend/src/components/Pages/MissionHistoryPage/HistoricMissionCard.tsx
```diff
@@ -42,6 +42,9 @@ export function HistoricMissionCard({ index, mission }: IndexedMissionProps) {
                     {mission.name}
                 </Button>
             </Table.Cell>
+            <Table.Cell>
+                <Typography>{mission.area?.areaName}</Typography>
```
Eddasol, line 46: Do we want it to be empty when area is missing or do we want a placeholder value? If placeholder value is desired it can be achieved as shown below.

```suggestion
                <Typography>{mission.area ? mission.area.areaName : '-'}</Typography>
```

backend/api/Controllers/Models/RobotResponse.cs
```diff
@@ -1,43 +1,72 @@
-﻿using Api.Database.Models;
+﻿using System.Text.Json.Serialization;
+
+using Api.Database.Models;
 #pragma warning disable CS8618
```
Eddasol, line 4: Necessary?

Eddasol (approved):
LGTM :)

Additional improvement would be adding area to filters as well, but that can be done as a separate PR!

PR #1205, opened 2023-11-21
Add eslint to frontend pipeline

Christdej:


frontend/src/components/Contexts/MissionListsContext.tsx
```diff
@@ -114,7 +114,7 @@ export const useMissions = (): MissionsResult => {
                 })
             })
         }
-    }, [registerEvent, connectionReady])
+    }, [registerEvent, connectionReady, missionQueue])
```
aestene, line 117: Why has there been changes in this file?

Christdej: Because it caused a warning, and I dont think this create a circular dependency, so I guess it should be there.

andchiind (approved):
LGTM

PR #1211, opened 2023-11-22
Use token roles in controllers

andchiind:


andchiind:
A slight frontend refactor will be needed as we no longer want to send the whole list of robots with signalR, in order to only show the user the robots in the installations they can see. Instead of sending a list we want to send an update to each robot individually, and so more effort is needed to maintain this list on the frontend, similar to how we currently handle missions in the queue.
image

tsundvoll (approved):
I'd say we take this in now, and test it a bit afterwards to verify desired behaviour :)

backend/api.test/Mocks/HttpContextAccessorMock.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            get
+            {
+                var context = new DefaultHttpContext();
+                var tokenHandler = new JwtSecurityTokenHandler();
+                var claims = new List<Claim>
+                {
+                    new(ClaimTypes.Role, "Role.Admin"),
+                    new(ClaimTypes.Role, "Role.User"),
+                    new(ClaimTypes.Role, "Role.User.JSV")
```
oysand, line 23: Use Huldra for the test?

backend/api.test/Services/RobotService.cs
```diff
@@ -56,15 +59,23 @@ public async Task Read()
...
         [Fact]
         public async Task Create()
         {
-            var robotService = new RobotService(_context, _logger, _robotModelService, _signalRService);
+            var robotService = new RobotService(_context, _logger, _robotModelService, _signalRService, _accessRoleService);
+            var installationService = new InstallationService(_context, _accessRoleService);
+
+            var installation = await installationService.Create(new CreateInstallationQuery
+            {
+                Name = "Johan Sverdrup",
```
oysand, line 75: Huldra as test asset

backend/api/Database/Context/InitDb.cs
```diff
@@ -42,6 +43,27 @@ public static class InitDb
                 RobotPose = new Pose()
             };
 
+        private static List<AccessRole> GetAccessRoles()
+        {
+            var accessRole1 = new AccessRole
+            {
+                AccessLevel = RoleAccessLevel.USER,
+                RoleName = "Role.User"
```
prasm313, line 51: Replace Role.User with Role.User.HUAA. Add Huldra as installation.

backend/api/Database/Context/InitDb.cs
```diff
@@ -42,6 +43,27 @@ public static class InitDb
...
+            };
+
+            var accessRole2 = new AccessRole
```
prasm313, line 54: Remove accessRole2

backend/api/Services/AccessRoleService.cs
```diff
@@ -7,25 +7,37 @@ namespace Api.Services
...
             if (roles.Contains(SUPER_ADMIN_ROLE_NAME))
             {
-                return (await installationService.ReadAll()).Select((i) => i.InstallationCode).ToList();
+                return await context.Installations.Select((i) => i.InstallationCode).ToListAsync();
             }
             else
             {
                 return await context.AccessRoles.Include((r) => r.Installation)
-                    .Where((r) => roles.Contains(r.RoleName)).Select((r) => r.Installation != null ? r.Installation.InstallationCode : "").ToListAsync();
+                    .Where((r) => roles.Contains(r.RoleName)).Select((r) => r.Installation != null ? r.Installation.InstallationCode.ToUpperInvariant() : "").ToListAsync();
```
prasm313, line 40: Why invariant?

backend/api/Services/MissionRunService.cs
```diff
@@ -169,7 +169,7 @@ private IQueryable<MissionRun> GetMissionRunsWithSubModels()
                 .ThenInclude(robot => robot.Model)
                 .Include(missionRun => missionRun.Tasks)
                 .ThenInclude(task => task.Inspections)
-                .Where((m) => accessibleInstallationCodes.Result.Contains(m.Area.Installation.InstallationCode)); ;
+                .Where((m) => m.Area == null || accessibleInstallationCodes.Result.Contains(m.Area.Installation.InstallationCode.ToUpper())); ;
```
prasm313, line 173: Will KAA users get all missions from JSV without an area?

backend/api/Services/MissionDefinitionService.cs
```diff
@@ -169,7 +169,7 @@ private IQueryable<MissionDefinition> GetMissionDefinitionsWithSubModels()
                 .ThenInclude(missionRun => missionRun != null ? missionRun.Tasks : null)!
                 .ThenInclude(missionTask => missionTask.Inspections)
                 .ThenInclude(inspection => inspection.InspectionFindings)
-                .Where((m) => accessibleInstallationCodes.Result.Contains(m.Area.Installation.InstallationCode));
+                .Where((m) => m.Area == null || accessibleInstallationCodes.Result.Contains(m.Area.Installation.InstallationCode.ToUpper()));
```
prasm313, line 181: See comment on MissionRunService.cs.

backend/api/Services/RobotService.cs
```diff
@@ -62,9 +67,9 @@ public async Task<Robot> CreateFromQuery(CreateRobotQuery robotQuery)
                 };
                 context.Entry(robotModel).State = EntityState.Unchanged;
                 await context.Robots.AddAsync(newRobot);
-                await context.SaveChangesAsync();
-                _ = signalRService.SendMessageAsync("Robot list updated", GetEnabledRobotsWithSubModels().Select((r) => new RobotResponse(r)));
-                return newRobot;
+                await ApplyDatabaseUpdate(newRobot.CurrentInstallation);
+                _ = signalRService.SendMessageAsync("Robot added", newRobot?.CurrentInstallation, new RobotResponse(newRobot!));
```
prasm313, line 71: newRobot should either be followed by a "?" or a "!". Would argue that it should not be a question mark.

backend/api/Services/DeckService.cs
```diff
@@ -127,14 +133,25 @@ public async Task<Deck> Update(Deck deck)
             }
 
             context.Decks.Remove(deck);
-            await context.SaveChangesAsync();
+            await ApplyDatabaseUpdate(deck.Installation);
```
prasm313, line 136: Do we need to do the check? Only admin can access this endpoint.

backend/api.test/Services/RobotService.cs
andchiind: Then we will need to use Huldra in initDb as well I think, which may also change Dev, but that is probably fine

backend/api/Services/AccessRoleService.cs
andchiind: Invariant is done consistently regardless of culture. Since this is in an entity framework call it is being converted to SQL, which means that you are not allowed to specify culture. But on the other hand, omitting culture when using ToUpper will cause a linting error, unless we disable the linting for this line. So I though this was simplest. It will behave identically to ToUpper anyways

backend/api/Services/MissionRunService.cs
andchiind: If a missionRun does not have an area it also does not have an installation. We can enforce the use of Area in missionRun maybe but I think this will require some refactoring for the localisation code

backend/api/Services/RobotService.cs
andchiind: Yes I suppose we should assume that newRobot cannot be null

backend/api/Services/DeckService.cs
andchiind: Yes, since if we change the permissions in the endpoint at some point in the future, it is unlikely they will remember to make the change here as well. And this does no harm in my opinion, it is best to not delegate security just to the controllers.

backend/api/Services/MissionRunService.cs
```diff
@@ -133,8 +133,8 @@ public async Task<MissionRun> Update(MissionRun missionRun)
             if (missionRun.Area is not null) { context.Entry(missionRun.Area).State = EntityState.Unchanged; }
 
             var entry = context.Update(missionRun);
-            await context.SaveChangesAsync();
-            _ = signalRService.SendMessageAsync("Mission run updated", new MissionRunResponse(missionRun));
+            await ApplyDatabaseUpdate(missionRun.Area?.Installation);
```
oysand, line 136: Is this not updated from the mqtt? Do we need to check this?

backend/api.test/Mocks/HttpContextAccessorMock.cs
andchiind: This is not really possible unless we add huldra missions in echo as we use the installation code when fetching tag positions when running missions

andchiind: image

PR #1217, opened 2023-11-23
Modify function definitions

Eddasol:


frontend/src/components/Displays/MissionDisplays/MissionStatusDisplay.tsx
```diff
@@ -21,7 +21,7 @@ const StyledStatusIcon = styled.div`
     align-items: flex-end;
 `
 
-export function displayIcon(status: MissionStatus) {
+export const DisplayMissionStatusIcon = ({ status }: StatusProps) => {
```
andchiind, line 24: Could we just call it MissionStatusIcon? Since the "display" part is kind of a given

frontend/src/components/Pages/MissionPage/TaskOverview/TaskStatusDisplay.tsx
```diff
@@ -15,7 +15,7 @@ const StyledStatusDisplay = styled.div`
     align-items: flex-end;
 `
 
-function displayIcon(status: TaskStatus) {
+const DisplayTaskStatusIcon = ({ status }: StatusProps) => {
```
andchiind, line 18: Same thing here

frontend/src/components/Pages/MissionPage/VideoStream/VideoPlayerOvenPlayer.tsx
```diff
@@ -41,7 +41,7 @@ export function VideoPlayerOvenPlayer({ videoStream }: IVideoPlayerProps) {
     return <div id={videoStream.id} />
 }
 
-export function IsValidOvenPlayerType({ videoStream }: IVideoPlayerProps) {
+export const isValidOvenPlayerType = ({ videoStream }: IVideoPlayerProps) => {
```
andchiind, line 44: Since we only have one argument here, we could probably jsut have videoStream be a normal parameter instead of in an object.

frontend/src/utils/MapMarkers.tsx
```diff
@@ -31,14 +31,14 @@ export const PlaceTagsInMap = (mission: Mission, map: HTMLCanvasElement, current
...
-export const PlaceRobotInMap = (mapMetadata: MapMetadata, map: HTMLCanvasElement, robotPose: Pose) => {
+export const placeRobotInMap = (mapMetadata: MapMetadata, map: HTMLCanvasElement, robotPose: Pose) => {
     const pixelPosition = calculateObjectPixelPosition(mapMetadata, robotPose.position)
     const rad = calculateNavigatorAngle(robotPose)
     drawRobotMarker(pixelPosition[0], pixelPosition[1], map, 22)
     drawNavigator(pixelPosition[0], pixelPosition[1], map, rad)
 }
 
-export const InverseCalculatePixelPosition = (mapMetadata: MapMetadata, pixelPosition: ObjectPosition) => {
+export const inverseCalculatePixelPosition = (mapMetadata: MapMetadata, pixelPosition: ObjectPosition) => {
```
andchiind, line 41: I don't fully understand this name? Is it inverting it or is an inverse calculation something else? We could maybe call it getInversePixelPosition /calculateInversePixelPosition or something similar to make it easier to understand?

andchiind (commented):
Looks good otherwise

frontend/src/utils/MapMarkers.tsx
Eddasol: There already is a calculatePixelPosition function and this gives the inverse of that function

frontend/src/components/Pages/MissionPage/VideoStream/VideoStreamWindow.tsx
```diff
@@ -40,7 +40,7 @@ export function VideoStreamWindow({ videoStreams }: VideoStreamWindowProps) {
         toggleFullScreenMode()
     }
 
-    const videoCards = videoStreams.map(function (videoStream, index) {
+    const videoCards = videoStreams.map((videoStream, index) => {
         return (
```
andchiind, line 44: The { and the "return" is not needed here, but I suppose this is not the focus of this refactoring anyways

frontend/src/components/Pages/FrontPage/RobotCards/RobotStatusSection.tsx
```diff
@@ -51,9 +51,7 @@ export function RobotStatusSection() {
     }, [enabledRobots, installationCode, switchSafeZoneStatus])
 
     const getRobotDisplay = () => {
```
andchiind, line 53: Does this need to be a function?
```suggestion
    const robotDisplay = robots.map((robot) => <RobotStatusCard key={robot.id} robot={robot} />)
```

andchiind: In general if a local function has no arguments (or if the argument values never change) and no side effects, then it can be safely made into a variable

frontend/src/components/Pages/MissionPage/MapPosition/MissionMapView.tsx
```diff
@@ -67,7 +67,7 @@ export function MissionMapView({ mission }: MissionProps) {
     }
 
     const findCurrentTaskOrder = useCallback(() => {
-        mission.tasks.forEach(function (task) {
+        mission.tasks.forEach((task) => {
             if (task.status === TaskStatus.InProgress || task.status === TaskStatus.Paused) {
                 setCurrentTaskOrder(task.taskOrder)
```
andchiind, line 72: Unrelated to your PR, but this isn't necessarily safe. Sorry you can ignore this comment, this is moreso a reminder to myself :)

andchiind (approved):
LGTM, the comment is very minor and can be ignored if you want, to be honest

frontend/src/components/Pages/FrontPage/RobotCards/RobotStatusSection.tsx
Eddasol: There are several like this so I suggest fixing them in separate PR

PR #1258, opened 2023-12-13
Fix inspectionfinding test to become deterministic

UsamaEquinorAFK:


andchiind (approved):
This looks good assuming we don't expect to get more than one inspection within a minute

UsamaEquinorAFK:
> This looks good assuming we don't expect to get more than one inspection within a minute

I agree, there is a downside to this not having seconds but every finding carries a specific ID so the time is not the only distinguishing factor in any way :)

PR #1260, opened 2023-12-13
Prevent collisions in signalr messages in local development

### LINKED ISSUE #1256: Use signalR group for local development
We currently use groups to distinguish different access rights. We also want to make these even more specific when developing locally so that the group is unique to each local user. This can either be a generated ID or a value set in .env and read from appsettings.

It will avoid signalR conflicts in local development.

N/A, but don't use personal information to make the group name.

andchiind:
Closes #1256


andchiind:
Another PR is being made where a Local enviroment is being added. I will use this environment instead of Development, which is why this PR is still a draft

prasm313 (approved):
LGTM

PR #1262, opened 2023-12-14
Remove default robot constructor

### LINKED ISSUE #1220: Remove default Robot constructor
The default robot constructor in `Robot.cs` is not really in use and can be removed

Not much, but will make the code more readable

N/A

andchiind:
Closes #1220


andchiind:
It seems that entity framework wants there to be a default constructor. Someone else might be able to make this PR work, but I currently don't see the need for it

PR #1269, opened 2023-12-18
Pass NA in InspectionReport if some entities are null

UsamaEquinorAFK:


backend/api/EventHandlers/InspectionFindingEventHandler.cs
```diff
@@ -77,19 +81,32 @@ private async Task<List<Finding>> GenerateFindingsList(List<InspectionFinding> i
...
 
-                if (task?.TagId != null && missionRun?.Area?.Plant?.Name != null && missionRun?.Area?.Name != null)
+                if (task != null && missionRun != null && missionRun?.Area != null)
                 {
                     var finding = new Finding(
-                        task.TagId,
-                        missionRun.Area.Plant.Name,
-                        missionRun.Area.Name,
+                        task.TagId ?? "NA",
+                        missionRun.Area.Plant.Name ?? "NA",
```
Afonso-2403, line 88:
```suggestion
                        missionRun.Area?.Plant.Name ?? "NA",
```

backend/api/EventHandlers/InspectionFindingEventHandler.cs
```diff
@@ -77,19 +81,32 @@ private async Task<List<Finding>> GenerateFindingsList(List<InspectionFinding> i
...
+                        missionRun.Area.Name ?? "NA",
```
Afonso-2403, line 89:
```suggestion
                        missionRun.Area?.Name ?? "NA",
```

backend/api/EventHandlers/InspectionFindingEventHandler.cs
```diff
@@ -77,19 +81,32 @@ private async Task<List<Finding>> GenerateFindingsList(List<InspectionFinding> i
...
                         inspectionFinding.Finding,
                         inspectionFinding.InspectionDate,
-                        missionRun.Robot.Name
+                        missionRun.Robot.Name ?? "NA"
                     );
 
                     findingsList.Add(finding);
                 }
+                else if (task != null && missionRun?.Area == null && missionRun != null)
```
Afonso-2403, line 97: With the suggestions above I think you can drop this else

backend/api/EventHandlers/InspectionFindingEventHandler.cs
```diff
@@ -77,19 +81,32 @@ private async Task<List<Finding>> GenerateFindingsList(List<InspectionFinding> i
                 var missionRun = await InspectionFindingService.GetMissionRunByIsarStepId(inspectionFinding);
                 var task = await InspectionFindingService.GetMissionTaskByIsarStepId(inspectionFinding);
 
-                if (task?.TagId != null && missionRun?.Area?.Plant?.Name != null && missionRun?.Area?.Name != null)
+                if (task != null && missionRun != null && missionRun?.Area != null)
```
Afonso-2403, line 84:
```suggestion
                if (task != null && missionRun != null)
```

UsamaEquinorAFK: Agreed, done.

backend/api/EventHandlers/InspectionFindingEventHandler.cs
UsamaEquinorAFK: Agreed, done.

backend/api/EventHandlers/InspectionFindingEventHandler.cs
UsamaEquinorAFK: Agreed, done.

backend/api/EventHandlers/InspectionFindingEventHandler.cs
UsamaEquinorAFK: Agreed, done.

andchiind (approved):
This does imply we cannot have a plant/area/robot with the name "NA", so I'm not sure if this is ideal compared to a longer string that says that it's missing, or making the fields nullable in Finding. But for now we can leave it like this for testing

UsamaEquinorAFK:
> This does imply we cannot have a plant/area/robot with the name "NA", so I'm not sure if this is ideal compared to a longer string that says that it's missing, or making the fields nullable in Finding. But for now we can leave it like this for testing

I agree. What we mean by NA here is Not Available. Further more we need to review the "required" status of various attributes in the database models. For example why would TagID ever be nullable in MissionTask same goes for Area in MissionRun.

Issue #1075, opened 2023-10-23
Possibility to filter in the Area GET Endpoint


Christdej:
**Describe the improvement you would like to see**
The ability to filter area when GETing areas from the database, for instance on Installation, Deck and so on

**How will this change existing functionality?**
This will improve the usability of the API

aestene:
This issue has automatically been marked as stale as there has been no activity for 60 days.

PR #1273, opened 2023-12-18
Possibility to filter in the Area GET Endpoint

UsamaEquinorAFK:
[#1075](https://github.com/equinor/flotilla/issues/1075)

### DIFF
--- a/backend/api/Controllers/AreaController.cs
+++ b/backend/api/Controllers/AreaController.cs
@@ -1,6 +1,7 @@
 ﻿using Api.Controllers.Models;
 using Api.Database.Models;
 using Api.Services;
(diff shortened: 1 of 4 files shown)


andchiind (approved):
LGTM. I haven't tested it, but I assume the PageSize can be set in the query, since the query extends a class which does

UsamaEquinorAFK:
> LGTM. I haven't tested it, but I assume the PageSize can be set in the query, since the query extends a class which does

I have tested it. The main purpose it to be able filter over the ReadAll query in the Area endpoint. Right now we can filter over deck and installationcode. I understand pagination can be configured in the query.

PR #1274, opened 2023-12-18
Remove polling and refactor mission page

### LINKED ISSUE #1265: Remove polling in MissionPage.tsx
Use signalR instead of polling. Only one request needs to be sent and there is already a context which recieves events for missions.

It optimise the page and make it easier to debug.

N/A

andchiind:
Closes #1265


oysand (approved):
LGTM

PR #1278, opened 2023-12-20
Compare deck and installation using ID

aestene:


Afonso-2403 (approved):
This is why Huldra was not Huldra!

andchiind (approved):
LGTM

PR #1279, opened 2023-12-20
Update individual properties

aestene:


backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -135,13 +135,13 @@ private async void OnIsarRobotInfo(object? sender, MqttReceivedArgs mqttArgs)
 
                 List<string> updatedFields = [];
 
-                if (isarRobotInfo.VideoStreamQueries is not null) { UpdateVideoStreamsIfChanged(isarRobotInfo.VideoStreamQueries, ref robot, ref updatedFields); }
-                if (isarRobotInfo.Host is not null) { UpdateHostIfChanged(isarRobotInfo.Host, ref robot, ref updatedFields); }
+                if (isarRobotInfo.VideoStreamQueries is not null) UpdateVideoStreamsIfChanged(isarRobotInfo.VideoStreamQueries, ref robot, ref updatedFields);
```
andchiind, line 138: It could be good to also apply these checks and update strategy in the Update endpoint in the Robot controller, as this currently does not work. We can do this in another PR though, and we would need a new query class since passing a Robot object to the endpoint is not ideal.

andchiind: I have updated the mentioned endpoint in a different PR, so no worries

backend/api/Controllers/RobotController.cs
```diff
@@ -234,16 +231,10 @@ public async Task<ActionResult<RobotResponse>> DeleteRobot([FromRoute] string id
...
             try
             {
-                var updatedRobot = await robotService.Update(robot);
-                var robotResponse = new RobotResponse(updatedRobot);
-
-                logger.LogInformation("Successful PUT of robot to database");
-
-                return Ok(robotResponse);
+                await robotService.UpdateRobotStatus(id, robotStatus);
+                return Ok();
```
andchiind, line 237: RobotService.UpdateRobotStatus still seems to be returning a Robot object. What is the reason for not returning it here?

backend/api/Controllers/RobotController.cs
```diff
@@ -234,16 +231,10 @@ public async Task<ActionResult<RobotResponse>> DeleteRobot([FromRoute] string id
...
             }
 
-            robot.Status = robotStatus;
-
             try
             {
-                var updatedRobot = await robotService.Update(robot);
-                var robotResponse = new RobotResponse(updatedRobot);
-
-                logger.LogInformation("Successful PUT of robot to database");
```
andchiind, line 244: And what is the reason for not logging it?

backend/api/Services/RobotService.cs
```diff
@@ -98,29 +99,142 @@ public async Task<Robot> CreateFromQuery(CreateRobotQuery robotQuery)
...
-            return await UpdateRobotProperty(robotId, "Status", status);
+            var robotQuery = context.Robots.Where(robot => robot.Id == robotId).Include(robot => robot.CurrentInstallation);
+            var robot = await robotQuery.FirstOrDefaultAsync();
+
+            await VerifyThatUserIsAuthorizedToUpdateDataForInstallation(robot!.CurrentInstallation);
+
+            await robotQuery.ExecuteUpdateAsync(robots => robots.SetProperty(r => r.Status, status));
+
+            robot = await robotQuery.FirstOrDefaultAsync();
+            NotifySignalROfUpdatedRobot(robot!, robot!.CurrentInstallation!);
```
andchiind, line 112: Why are we stating in the function return value that the returned Robot can be null, whilst here we are stating that it won't be null?

backend/api/Services/RobotService.cs
```diff
@@ -98,29 +99,142 @@ public async Task<Robot> CreateFromQuery(CreateRobotQuery robotQuery)
...
+            ThrowIfRobotIsNull(robot, robotId);
+            NotifySignalROfUpdatedRobot(robot!, robot!.CurrentInstallation!);
+
+            return robot;
+        }
+        public async Task<Robot> UpdateCurrentArea(string robotId, Area? area) { return await UpdateRobotProperty(robotId, "CurrentArea", area); }
+        public async Task<Robot> UpdateMissionQueueFrozen(string robotId, bool missionQueueFrozen)
         {
-            return await GetRobotsWithSubModels().FirstOrDefaultAsync(robot => robot.Id.Equals(id));
+            var robotQuery = context.Robots.Where(robot => robot.Id == robotId).Include(robot => robot.CurrentInstallation);
```
andchiind, line 219: Now that the functions are not all one-liners, could we maybe add a line between them to space them out?

backend/api/Services/RobotService.cs
aestene: This function was missing the throw if null function calls. Fixed now.

backend/api/Controllers/RobotController.cs
aestene: I changed it back. The reason it was removed was because it should not return a robot object but I changed my mind midway through the PR.

backend/api/Controllers/RobotController.cs
aestene: Added logging back.

PR #1280, opened 2023-12-20
Add check for min pressure in robot when starting missions

### LINKED ISSUE #1019: Evaluate battery and pressure levels before starting missions
It should be clearly displayed if the battery and pressure levels are sufficient to start a mission or not. Consider a warning in the frontend when battery/pressure levels are starting to get low (#722). Furthermore, an error message, telling the user that the robot must be charged and/or pressurized before starting a mission, should be considered. 

A suggestion could be a pop-up explaining the required actions by the user before it is allowed to start a mission. 

andchiind:
Closes #1019 

Currently waiting on https://github.com/equinor/flotilla/pull/1279 to be merged, since we need to rewrite the update robot controller in order to implement this, which will need to use the new update service functions in that PR.


andchiind:
Here is what it looks like in the frontend: 
image

frontend/src/components/Pages/InspectionPage/ScheduleMissionDialogs.tsx
```diff
@@ -77,10 +77,12 @@ export const ScheduleMissionDialog = (props: IProps): JSX.Element => {
         if (!selectedRobot) return
 
         props.missions.forEach((mission) => {
-            BackendAPICaller.scheduleMissionDefinition(mission.id, selectedRobot.id).catch(() => {
+            BackendAPICaller.scheduleMissionDefinition(mission.id, selectedRobot.id).catch((e) => {
                 setAlert(
                     AlertType.RequestFail,
-                    <FailedRequestAlertContent message={`Failed to schedule mission ${mission.name}`} />
+                    <FailedRequestAlertContent
+                        message={TranslateText('Failed to schedule mission') + ` '${mission.name}'. ${e.message}`}
```
Eddasol, line 84: There is translation on the message here but not other places (e.g. on line 117)

backend/api/Services/RobotService.cs
```diff
@@ -224,6 +225,24 @@ public async Task<Robot> UpdateCurrentMissionId(string robotId, string? currentM
...
+        public async Task<Robot> UpdateCurrentInstallation(string robotId, Installation? installation)
+        {
+            var robotQuery = context.Robots.Where(robot => robot.Id == robotId).Include(robot => robot.CurrentInstallation);
+            var robot = await robotQuery.FirstOrDefaultAsync();
+            ThrowIfRobotIsNull(robot, robotId);
+
+            await VerifyThatUserIsAuthorizedToUpdateDataForInstallation(robot!.CurrentInstallation);
+            await VerifyThatUserIsAuthorizedToUpdateDataForInstallation(installation);
+
+            await robotQuery.ExecuteUpdateAsync(robots => robots.SetProperty(r => r.CurrentInstallation, installation));
```
Eddasol, line 237: Should UpdateCurrentInstallation modify/check currentArea as well so we don't get a mismatch for installation? Or maybe that should be handled elsewhere?

aestene: CurrentArea should be null unless to robot is localized and driving. Thus changing the installation while the robot is operating isn't something we should be doing. If you want to handle the edge case then one can check if the robot is available or not but I would say this is not necessary.

backend/api/Controllers/MissionSchedulingController.cs
```diff
@@ -351,6 +361,8 @@ ISourceService sourceService
 
             MissionRun? newMissionRun;
             try { newMissionRun = await customMissionSchedulingService.QueueCustomMissionRun(customMissionQuery, customMissionDefinition.Id, robot.Id, missionTasks); }
+            catch (RobotPressureTooLowException e) { return BadRequest(e.Message); }
+            catch (RobotBatteryLevelTooLowException e) { return BadRequest(e.Message); }
```
aestene, line 365:
```suggestion
            catch (Exception e) when (e is RobotPressureTooLowException or RobotBatteryLevelTooLowException) { return BadRequest(e.Message); }
```

Eddasol:
This works if scheduling echo missions but not if scheduling missions trough inspection page or re-running a mission

frontend/src/components/Pages/FrontPage/MissionOverview/ScheduleMissionDialog/SelectMissionsToScheduleDialog.tsx
```diff
@@ -51,10 +51,12 @@ export const SelectMissionsToScheduleDialog = ({ echoMissionsList, closeDialog }
...
 
         selectedEchoMissions.forEach((mission: EchoMissionDefinition) => {
-            BackendAPICaller.postMission(mission.echoMissionId, selectedRobot.id, installationCode).catch(() => {
+            BackendAPICaller.postMission(mission.echoMissionId, selectedRobot.id, installationCode).catch((e) => {
                 setAlert(
                     AlertType.RequestFail,
-                    <FailedRequestAlertContent message={`Failed to schedule mission ${mission.name}`} />
+                    <FailedRequestAlertContent
+                        message={TranslateText('Failed to schedule mission') + ` '${mission.name}'. ${e.message}`}
+                    />
```
Eddasol, line 59: There is translation on message before sending as input and inside FailedRequestAlertContent. Suggest removing translation on message inside FailedRequestAlertContent and instead set translation on all messages before sending as input. Perhaps also renaming message to translatedMessage

frontend/src/components/Pages/InspectionPage/ScheduleMissionDialogs.tsx
```diff
@@ -110,10 +110,12 @@ export const ScheduleMissionDialog = (props: IProps): JSX.Element => {
         if (!selectedRobot || !missionsToSchedule) return
 
         missionsToSchedule.forEach((mission) => {
```
Eddasol, line 112: Should there also be a frontend check such that the user get feedback before going trough the localization dialog. Like a check on line 101 where the user get a dialog or something when the pressure is low,  instead of actually scheduling?

Eddasol:
It prevents the user from scheduling the mission when the pressure is too low, but queued missions will still begin. Should there be some check on that as well?

andchiind:
> It prevents the user from scheduling the mission when the pressure is too low, but queued missions will still begin. Should there be some check on that as well?

This would involve having a check in the queue event handler which cancels missions if the pressure is too low. In effect it would cancel all missions in the queue and then send an error to the frontend for each mission. It is possible but I think we may need to have further UX discussions before implementing this. The scope of this PR was to provide user feedback when scheduling missions.

frontend/src/components/Pages/InspectionPage/ScheduleMissionDialogs.tsx
andchiind: This has been added, but I would like your feedback on it

backend/api/Controllers/RobotController.cs
```diff
@@ -173,6 +175,97 @@ public async Task<ActionResult<RobotResponse>> CreateRobot([FromBody] CreateRobo
             }
         }
 
+        /// <summary>
```
mrica-equinor, line 178: why is this part of the PR? Just wondering if it's intentional or something in the rebase because i thought we already had this functionality but maybe not in the robot controller

backend/api/Services/RobotService.cs
```diff
@@ -222,6 +223,24 @@ public async Task<Robot> UpdateCurrentMissionId(string robotId, string? currentM
...
```
mrica-equinor, line 235: why not use the updateRobotField here?

frontend/src/api/ApiError.tsx
```diff
@@ -0,0 +1,70 @@
+import { CustomError } from 'ts-custom-error'
```
mrica-equinor, line 1: Love this

frontend/src/components/Alerts/FailedRequestAlert.tsx
```diff
@@ -28,7 +28,7 @@ export const FailedRequestAlertContent = ({ message }: { message: string }) => {
             </StyledAlertTitle>
             <Indent>
                 <Button as={Typography} variant="ghost" color="secondary">
-                    {TranslateText(message)}
+                    {translatedMessage}
```
mrica-equinor, line 31: This is very picky but for consistency I would rather have 

`<FailedRequestAlertContent message={'Failed to rerun mission'} />`
```
<Button as={Typography} variant="ghost" color="secondary">
                    {TranslateText(message)}
```

andchiind: The issue is that the message cannot always be translated as a full string, instead part of it need to be translated separately. The message for instance can contain a custom message from the backend. We may also want to include robot names, etc.

mrica-equinor (approved):
LGTM :)

PR #1282, opened 2023-12-20
Fix mission queue not refreshing upon mission delete

### LINKED ISSUE #1181: Removing missions from the queue bug
**Describe the bug**
If we schedule a mission with shockwave for example, when I click on the remove mission from the queue button, the mission only disappears when I refresh the page. This does not happen for placebot.

**To Reproduce**
Schedule mission with shockwave and remove it from the queue.

**Expected behavior**
The mission should be removed without having to refresh the page.

**Screenshots**
In this picture I had already confirmed that I wanted to remove the mission from the queue and the dialog had been closed.

Image

andchiind:
Closes #1181

Also adds frontend error handling when failing to delete a mission

### DIFF
--- a/backend/api/appsettings.Local.json
+++ b/backend/api/appsettings.Local.json
@@ -44,5 +44,8 @@
   },
   "ApplicationInsights": {
     "ConnectionString": ""
+  },
+  "Database": {
+    "UseInMemoryDatabase": true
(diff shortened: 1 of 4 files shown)


oysand (approved):
LGTM

PR #1285, opened 2023-12-21
Add debug log to robot property update

oysand:


andchiind (approved):
LGTM for now, though we may want to remove some of these logs later since this log will be written a lot for change in battery/pressure levels

PR #1289, opened 2023-12-21
Bug stuck in busy

oysand:


aestene:

andchiind:
The robot gets stuck in busy even though the missionStatus is Idle and robot status is Available in ISAR. It occurs when the mission sent to Isar finishes instantly, eg. when Localisation is ignored, and Flotilla does not seem to update its own state based on the heartbeat info.

PR #1290, opened 2023-12-21
Refactor event handlers

Afonso-2403:


backend/api/EventHandlers/MissionEventHandler.cs
```diff
@@ -116,34 +116,21 @@ private async void OnRobotAvailable(object? sender, RobotAvailableEventArgs e)
...
 
-                var lastExecutedMissionRun = await MissionService.ReadLastExecutedMissionRunByRobot(robot.Id);
-                if (lastExecutedMissionRun is null)
-                {
-                    _logger.LogError("Could not find last executed mission run for robot");
-                    return;
-                }
-
-                if (!lastExecutedMissionRun.IsDriveToMission())
+                if (await ReturnToHomeService.IsRobotHome(robot.Id)) { await RobotService.UpdateCurrentArea(robot.Id, null); }
```
oysand, line 132: Add return here and have the functionality below withouth the else?

oysand: Consider removing the return on line 143 if this is done

backend/api/Services/ReturnToHomeService.cs
```diff
@@ -4,11 +4,24 @@ namespace Api.Services
...
+        {
+            var lastExecutedMissionRun = await missionRunService.ReadLastExecutedMissionRunByRobot(robotId);
+            if (lastExecutedMissionRun is null)
+            {
+                logger.LogInformation("Could not find last executed mission run for robot {RobotId}, can not guarantee that the robot is in its home", robotId);
+                return false;
+            }
+
+            if (lastExecutedMissionRun.IsDriveToMission()) { return true; }
+            else { return false; }
```
oysand, line 23:
```suggestion
            return lastExecutedMissionRun.IsDriveToMission();
```

backend/api/Services/RobotService.cs
```diff
@@ -163,6 +165,39 @@ public async Task<IList<Robot>> ReadLocalizedRobotsForInstallation(string instal
                 .ToListAsync();
         }
 
+        public async Task OnIsarUnavailable(string robotId)
```
oysand, line 168: Consider moving to an event handler. Could be RobotEventHandler or something descriptive. If it is to be in the service it should be renamed so id does not sound like something that handles an event

Afonso-2403: Given that it is not called in the sequence of an event I kept it in `RobotService` and renamed it to `SetRobotOffline`

oysand (approved):
LGTM

PR #1294, opened 2023-12-27
Add alert for failure to send robot to safe zone

### LINKED ISSUE #1291: The frontend should receive a warning the robot could not be returned home
When the scheduling of a return home mission fails in [`OnRobotAvailable`](https://github.com/equinor/flotilla/blob/23b00904d746a51691b5b44f662af1b5cb48389c/backend/api/EventHandlers/MissionEventHandler.cs#L109), a notification should be sent to the frontend so that an appropriate message can be shown to the user.
Currently we silently fail which might lead the user to believe that the robot is in its home.

A signalR topic?

This will make the system more robust given that the user will be warned if a return to home mission could not be scheduled and the robot should be physically handled.

andchiind:
Closes #1291

### DIFF
--- a/backend/api/Controllers/Models/AlertResponse.cs
+++ b/backend/api/Controllers/Models/AlertResponse.cs
@@ -1,0 +1,13 @@
+﻿using System.Text.Json.Serialization;
+namespace Api.Controllers.Models
+{
+    [method: JsonConstructor]
+    public class AlertResponse(string code, string name, string message, string installationCode, string? robotId)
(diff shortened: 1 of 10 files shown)


andchiind:
I would like to address this before making this not a draft: https://github.com/equinor/flotilla/issues/1291#issuecomment-1870263946

UsamaEquinorAFK (approved):
Did not test. Looks clean and good.

PR #1295, opened 2023-12-28
Add best frontend practises document

### LINKED ISSUE #1213: Create best practise document for the frontend
A document describing best practises for React development in the frontend.

Same as above, but also ideally it should be a .md file in the frontend folder. A docs folder can be used if we want to link several files together.

N/A

andchiind:
Closes #1213 although more information can always be added.


UsamaEquinorAFK (approved):
lgtm

PR #1297, opened 2024-01-02
Restructure "Add predefined mission" button

### LINKED ISSUE #1286: The Add Predefined Echo Missions button moves slightly when dialog is open
**Describe the bug**
When opening the dialog by pressing the button, the space between the two buttons widen. This is likely since the dialog is between these buttons in the React component, and the dialog has a min size. This can therefore be fixed by moving the dialog to the end of the component, or enforcing a component size of zero.

**To Reproduce**
Press the "Add Predefined Echo Missions" button.

**Expected behavior**
The button should remain in place.

**Screenshots**
image
image

andchiind:
Closes #1286 

This turned out to not be a good first issue, since solving it required refactoring.

It's worth noting that we now assume that an installation is set when we are on the main page. I think this is a fair assumption.


frontend/src/components/Pages/FrontPage/MissionOverview/ScheduleMissionDialog/SelectMissionsToScheduleDialog.tsx
```diff
@@ -1,7 +1,7 @@
 import { Autocomplete, Button, Card, Dialog, Typography } from '@equinor/eds-core-react'
 import styled from 'styled-components'
 import { useLanguageContext } from 'components/Contexts/LanguageContext'
-import { useState } from 'react'
+import React, { useState } from 'react'
```
Eddasol, line 4:
```suggestion
import {memo,  useState } from 'react'
```

Eddasol (approved):
LGTM :)

PR #1305, opened 2024-01-04
Remove unused functions and models

Eddasol:


andchiind (approved):
LGTM, I'm surprised ESlint didn't catch these before

Eddasol:
> LGTM, I'm surprised ESlint didn't catch these before

Seems like it doesn't catch unused variables if they are marked "export"

PR #1308, opened 2024-01-04
Filter away localisation/return missions

### LINKED ISSUE #1300: Do not show localisation missions in mission history
**Describe the bug**
The localisation missions are presented as normal missions on the history page

**To Reproduce**
Run a mission and look at the history page

**Expected behavior**
Only "real" missions should be there

**Screenshots**
image

andchiind:
Closes #1300


Eddasol (approved):
LGTM :)

PR #1309, opened 2024-01-04
Make more fields required

### LINKED ISSUE #1270: Review the Database models for attributes which should or should not be required

Some of the database models carry attributes should be checked if they are required or not required. For example TagID in MissionTask and Area in MissionRun. 

Make the code in services and eventhandlers better and would be lesser null checks.

NA.

andchiind:
Closes #1270

### DIFF
--- a/backend/api/Database/Models/Inspection.cs
+++ b/backend/api/Database/Models/Inspection.cs
@@ -50,11 +50,12 @@ namespace Api.Database.Models
         [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
         public string Id { get; set; }
 
+        [Required]
         [MaxLength(200)]
         // ReSharper disable once AutoPropertyCanBeMadeGetOnly.Local
-        public string? IsarStepId { get; private set; } = Guid.NewGuid().ToString();
+        public string IsarStepId { get; private set; } = Guid.NewGuid().ToString();
 
-        public Position InspectionTarget { get; set; }
+        public Position? InspectionTarget { get; set; }
 
         [Required]
         public InspectionStatus Status
--- a/backend/api/Controllers/Models/RobotResponse.cs
+++ b/backend/api/Controllers/Models/RobotResponse.cs
@@ -14,7 +14,7 @@ namespace Api.Controllers.Models
 
         public string SerialNumber { get; set; }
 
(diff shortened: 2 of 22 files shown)


andchiind:
A good way to pad my line contributions

backend/api/Database/Models/Inspection.cs
```diff
@@ -50,10 +50,12 @@ public Inspection(Inspection copy, InspectionStatus? inspectionStatus = null)
         [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
         public string Id { get; set; }
 
+        [Required]
         [MaxLength(200)]
         // ReSharper disable once AutoPropertyCanBeMadeGetOnly.Local
         public string? IsarStepId { get; private set; } = Guid.NewGuid().ToString();
```
Christdej, line 56: Can the question mark in `string?` be removed?

backend/api/Database/Models/Inspection.cs
```diff
@@ -50,10 +50,12 @@ public Inspection(Inspection copy, InspectionStatus? inspectionStatus = null)
...
 
+        [Required]
```
Christdej, line 58: Should this be required? For instance if you want to take a sound recording at a given spot, you dont need an InspectionTarget

Christdej (changes requested):

backend/api/Database/Models/Inspection.cs
andchiind: Yes actually. It was nullable since we use an empty value when generating hash values, but we now use an empty string anyways so this can be required in C# as well.

backend/api/Database/Models/Inspection.cs
andchiind: I tried to make it match the nullability of the C# object, but I can make them both nullable

Christdej: Think we make both nullable then?
`public Position? InspectionTarget`

backend/api/Database/Models/Inspection.cs
Christdej: Then I think we remove it

Christdej (approved):
LGTM =)

andchiind:
"/UpdateDatabase

PR #1317, opened 2024-01-08
Sort robots by status and name

Eddasol:


frontend/src/components/Pages/FrontPage/RobotCards/RobotStatusSection.tsx
```diff
@@ -35,7 +36,15 @@ export const RobotStatusSection = () => {
             (robot) =>
                 robot.currentInstallation.installationCode.toLocaleLowerCase() === installationCode.toLocaleLowerCase()
         )
-        .sort((robot, robotToCompareWith) => (robot.status! > robotToCompareWith.status! ? 1 : -1))
+        .sort((robot, robotToCompareWith) =>
+            robot.status! !== robotToCompareWith.status!
+                ? robot.status! > robotToCompareWith.status!
+                    ? 1
+                    : -1
```
tsundvoll, line 43: 1 and -1  are kind of magic numbers. Do they represent True / False, could we use that instead?

tsundvoll (changes requested):

UsamaEquinorAFK (approved):
lgtm but agree with Thomas  :p

frontend/src/components/Pages/FrontPage/RobotCards/RobotStatusSection.tsx
Eddasol: Kind of. Usually sorting is done by returning a - b but when sorting by something that isn't a number this is the common way to do it in typescript.

tsundvoll: OKey, good clarification, thank you

tsundvoll (approved):

andchiind (approved):
LGTM

PR #1324, opened 2024-01-09
Improve blocked alert

Eddasol:
* Moved to alertContext such that the banner will be updated regardless of viewed paged
* Removed setAlert from dependency in useEffect to fix warning (see linked issue)
* Only call setAlert when blockedRobotNames are changed such that the banner can be dismissed. Will reappear if changes to which robots are blocked or if page is refreshed
* Updated banner to display alternative text if more than on robot is blocked

### DIFF
--- a/frontend/src/components/Pages/FrontPage/RobotCards/RobotStatusSection.tsx
+++ b/frontend/src/components/Pages/FrontPage/RobotCards/RobotStatusSection.tsx
@@ -2,9 +2,7 @@ import { Typography } from '@equinor/eds-core-react'
 import { Robot } from 'models/Robot'
 import { useEffect } from 'react'
 import styled from 'styled-components'
-import { BlockedRobotAlertContent } from 'components/Alerts/BlockedRobotAlert'
 import { RobotStatusCard, RobotStatusCardPlaceholder } from './RobotStatusCard'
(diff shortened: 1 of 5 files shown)


frontend/src/components/Pages/FrontPage/RobotCards/RobotStatusSection.tsx
```diff
@@ -21,16 +19,12 @@ const RobotView = styled.div`
...
-    return robot.status === 'Blocked'
-}
-
 export const RobotStatusSection = () => {
     const { TranslateText } = useLanguageContext()
     const { installationCode } = useInstallationContext()
     const { enabledRobots } = useRobotContext()
     const { switchSafeZoneStatus } = useSafeZoneContext()
-    const { setAlert } = useAlertContext()
+
```
UsamaEquinorAFK, line 27: Is this an extra un-necessary blank line ?

UsamaEquinorAFK (approved):
tested - lgtm. Just look at that one comment.

frontend/src/components/Pages/FrontPage/RobotCards/RobotStatusSection.tsx
Eddasol: Its intentional to create some more air to increase readability

PR #1325, opened 2024-01-10
Change the title of inspection card

prasm313:


frontend/src/components/Pages/InspectionPage/DeckCards.tsx
```diff
@@ -70,10 +70,8 @@ const DeckCard = ({ deckData, deckName, setSelectedDeck, selectedDeck, handleSch
                                 </StyledDict.Content>
                             ))}
                     </StyledDict.TopDeckText>
-                    {deckData.areas && <Typography variant={'body_short'}>{formattedAreaNames}</Typography>}
```
mrica-equinor, line 73: Can you try to explain why we no longer need to check for deck.areas and deck.inspections?

Eddasol: They cannot be undefined so it will always be true. Can alternatively check if its an empty list but that will not be triggered by checking if deckData.areas is true

Eddasol: I think I hard-coded it to [] and the text still showed. Its been a few days, so don't fully remember

PR #1327, opened 2024-01-10
Add rerun button to history table

prasm313:
As suggested in #1060.
image


frontend/src/components/Pages/MissionHistoryPage/HistoricMissionCard.tsx
```diff
@@ -51,6 +58,9 @@ export const HistoricMissionCard = ({ index, mission }: IndexedMissionProps) =>
             <Table.Cell>
                 <MissionEndTimeDisplay mission={mission} />
             </Table.Cell>
+            <Table.Cell>
+                <MissionRestartButton mission={mission} hasFailedTasks={missionHasFailedTasks} />
```
Eddasol, line 62: Consider using AddOutlined as icon instead of Replay as that is what we have in figma

PR #1329, opened 2024-01-12
Fix missing pressure on new robots

Eddasol:
This is a fix for:
* If pressure is null it will never get updated and will stay null
* If upperPressureWarningThreshold or lowerPressureWarningThreshold is null, the pressure icon will show but the pressure value is missing

### DIFF
--- a/frontend/src/components/Pages/RobotPage/RobotPage.tsx
+++ b/frontend/src/components/Pages/RobotPage/RobotPage.tsx
@@ -63,18 +63,16 @@ export const RobotPage = () => {
                                 {selectedRobot.status !== RobotStatus.Offline && (
                                     <>
                                         <BatteryStatusDisplay itemSize={48} batteryLevel={selectedRobot.batteryLevel} />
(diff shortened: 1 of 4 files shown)


frontend/src/components/Pages/RobotPage/RobotPage.tsx
```diff
@@ -63,18 +63,16 @@ export const RobotPage = () => {
                                 {selectedRobot.status !== RobotStatus.Offline && (
                                     <>
                                         <BatteryStatusDisplay itemSize={48} batteryLevel={selectedRobot.batteryLevel} />
-                                        {selectedRobot.model.upperPressureWarningThreshold && (
```
andchiind, line 66: Could we not just replace selectedRobot.model.upperPressureWarningThreshold here with selectedRobot.pressureLevel? Since that would remove the need for the first if statement in the PressureStatusDisplay component. Then we could avoid rendering the PressureStatusDisplay component, and it would be visible in the robotPage, when the component would not display anything

andchiind (approved):
LGTM

PR #1335, opened 2024-01-17
Move robot controller start functionality to service

aestene:


backend/api.test/EventHandlers/TestMissionEventHandler.cs
```diff
@@ -198,6 +175,7 @@ public async void NewMissionIsStartedWhenRobotBecomesAvailable()
 
             // Act
             _mqttService.RaiseEvent(nameof(MqttService.MqttIsarRobotStatusReceived), mqttEventArgs);
+            Thread.Sleep(500);
```
andchiind, line 178: What happens without the sleeps, how did they work before without them and how did you settle on the specific timings (since some of them differ)?

backend/api/Services/MissionSchedulingService.cs
```diff
@@ -48,13 +47,18 @@ public async Task StartMissionRunIfSystemIsAvailable(string missionRunId)
...
+                    or IsarCommunicationException)
             {
                 const MissionStatus NewStatus = MissionStatus.Failed;
-                logger.LogWarning(
-                    "Mission run {MissionRunId} was not started successfully. Status updated to '{Status}'.\nReason: {FailReason}",
+                logger.LogError(
+                    ex,
+                    "Mission run {MissionRunId} was not started successfully due to {ErrorMessage}",
                     missionRun.Id,
-                    NewStatus,
```
andchiind, line 57: Why was the NewStatus part of the error removed? It could be nice to have in the log when debugging

andchiind (commented):
Looks good otherwise

backend/api/Services/MissionSchedulingService.cs
aestene: Mainly because it is not a dynamic variable at this point and always fixed as "Failed". Thus it seemed a bit redundant to me because we're already saying that the mission was not started successfully.

backend/api.test/EventHandlers/TestMissionEventHandler.cs
aestene: I will honestly say I'm not sure how they functioned before. My suspicion is that the mocking we had in place for the robot controller meant the tests were so quick that they managed to reliably perform the operations before the assert was called. After removing the mocks it was very clear that there is a timing issue which is not surprising giving the design of the system. 

The specific times are not scientific but rather taken as run the test a few times, see what is suitable but not so high that it impacts test runtime too much. I think sleeps here are okey as the tests aren't very complicated. It's mainly "nice the other thread some time".

andchiind (approved):
LGTM

PR #1343, opened 2024-01-18
Use Bar as unit for pressure limits

Eddasol:


frontend/src/components/Displays/RobotDisplays/PressureStatusDisplay.tsx
```diff
@@ -14,29 +14,26 @@ const StyledTypography = styled(Typography)<{ $fontSize?: 24 | 16 | 18 | 32 | 40
     font-size: ${(props) => props.$fontSize};
 `
 interface PressureStatusDisplayProps {
-    pressureInBar: number
+    pressure: number
```
andchiind, line 17: Is it not a good idea to provide the unit in the name? Otherwise we need to look it up in the backend

andchiind: I am not sure if I agree with this. I just mean that as a developer looking at this code, it is not ideal to need to look somewhere else to find out what the units are, especially if we may need to apply a transformation. I don't think most developers, including me, know what the SI units are.

mrica-equinor: I have to agree with @andchiind here, sorry but I studied pressure at university and even I wouldn't know directly that it's in Bar. Also there's no way of getting confirmation that it's in SI other than coming to you and ask what unit of measurement this is or having to pull my calculator and checking. There's another variable that says from bar to milibar which we could look for but think would be more clear to have InBar in the name

Eddasol: I still disagree. Specifying unit would indicate that the unit diverges from what we normally use. If the name is pressure I would assume that its the same unit as robot.pressure without giving unit a thought. If its called pressureInBar I would not make that assumption and would start looking for some information that will tell me what conversion of robot.pressure I would need to make before using the function.

If you still disagree I can change it

andchiind: I tihnk if this is a concern then we should instead change PressureLevel in Robot.cs to state the unit, rather than make the frontend model less descriptive.

mrica-equinor: Yeah I agree with this, think we should have some indication right away of what unit robot.pressure is in

prasm313 (approved):
LGTM

Issue #1264, opened 2023-12-14
Introduce global ordering of mission runs


andchiind:
**Describe the improvement you would like to see**
Each new mission run should be assigned an incremented integer.

**How will this change existing functionality?**
This needs to be done with a mutex. This number can then be used to order the mission in the mission queue on the backend. By sending this integer to the frontend we can then have a consistent ordering. Missions that need to run immediately should have the value set to zero as a special case. This integer also allows us to implement reordering of missions in the queue, though this also requires the use of mutexes. Using a large integer accomodates a very large number of missions without running out, but if this is something we're worried about (an unsigned long allows 18,446,744,073,709,551,615 mission runs) then we can allow the value to wrap. To find what the next integer should be the backend should store the counter in a database table. Updating it needs to be done with mutexes (both for reading the last value and writing the new value).

I am open to discussions on this topic. I understand it introduces more overhead and complexity, but I think in the long run we should have a solution for consistent ordering.

**How will this improvement affect the current Threat Model?**
N/A

tsundvoll:
Notes:
Plan to add variable on MissionRun
Semaphore in scoped service will not work as intended (needs to be in a hosted service)

UsamaEquinorAFK:
https://github.com/equinor/flotilla/pull/1354

UsamaEquinorAFK:
https://learn.microsoft.com/en-us/ef/core/modeling/generated-properties?tabs=data-annotations

UsamaEquinorAFK:
https://stackoverflow.com/questions/36155429/auto-increment-on-partial-primary-key-with-entity-framework-core

UsamaEquinorAFK:
https://github.com/orgs/equinor/projects/85/views/4?pane=issue&itemId=45945484

PR #1354, opened 2024-01-24
Add missionRun counter

UsamaEquinorAFK:
#1264

### DIFF
--- a/backend/api/Services/MissionRunService.cs
+++ b/backend/api/Services/MissionRunService.cs
@@ -58,13 +58,28 @@ namespace Api.Services
         ILogger<MissionRunService> logger,
         IAccessRoleService accessRoleService) : IMissionRunService
     {
+
+        private static readonly SemaphoreSlim missionRunCountSemaphore = new(1);
+
         public async Task<MissionRun> Create(MissionRun missionRun, bool triggerCreatedMissionRunEvent = true)
         {
             missionRun.Id ??= Guid.NewGuid().ToString(); // Useful for signalR messages
+
+            missionRunCountSemaphore.Wait();
+            try
+            {
+                int maxCount = await context.MissionRuns.MaxAsync(mr => (int?)mr.MissionRunCount) ?? 0;
+                missionRun.MissionRunCount = maxCount + 1;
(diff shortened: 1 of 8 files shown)


backend/api/Services/MissionRunService.cs
```diff
@@ -58,13 +58,28 @@ public class MissionRunService(
...
+
         public async Task<MissionRun> Create(MissionRun missionRun, bool triggerCreatedMissionRunEvent = true)
         {
             missionRun.Id ??= Guid.NewGuid().ToString(); // Useful for signalR messages
+
+            await missionRunCountSemaphore.WaitAsync();
+            try
+            {
+
+                missionRun.MissionRunCount = await context.MissionRuns.CountAsync() + 1;
```
andchiind, line 72: This could create issues if we ever delete a record. We have deleted records at some points when doing migrations

UsamaEquinorAFK: The issue could be that we jump over integers in the queue but the sequence or the order will be intact I think.

andchiind: If we have missions with indexes: [1, 2, 3, 4] and then we remove the second one, then add a new mission, we'll have missions with indexed: [1, 3, 4, 4]

UsamaEquinorAFK: Good point ! Will fix this.

backend/api/Services/MissionRunService.cs
```diff
@@ -58,13 +58,28 @@ public class MissionRunService(
         ILogger<MissionRunService> logger,
         IAccessRoleService accessRoleService) : IMissionRunService
     {
+
+        private static readonly SemaphoreSlim missionRunCountSemaphore = new(1);
```
andchiind, line 62: Does this work in a transient service? And is it the reason you made the service transient?

UsamaEquinorAFK: As discussed, seems to be working while testing

andchiind (approved):
LGTM

PR #1355, opened 2024-01-24
Fix localization failure bug and refactor

oysand:
Main fix: Current area is set based on result from localization mission.

Other changes:
- Add new event OnMissionCompleted
  - Needed in the case of localization mission. We need to know the status of this mission when it completes before we have enough information to decide which mission (if any) to run next.
- Refactoring

### DIFF
--- a/backend/api.test/EventHandlers/TestMissionEventHandler.cs
+++ b/backend/api.test/EventHandlers/TestMissionEventHandler.cs
@@ -10,6 +10,7 @@ using Api.Mqtt.Events;
 using Api.Mqtt.MessageModels;
 using Api.Options;
 using Api.Services;
+using Api.Services.ActionServices;
 using Api.Test.Database;
 using Api.Test.Mocks;
 using Microsoft.AspNetCore.Builder;
@@ -43,6 +44,8 @@ namespace Api.Test.EventHandlers
             var mapServiceLogger = new Mock<ILogger<MapService>>().Object;
             var mapBlobOptions = new Mock<IOptions<MapBlobOptions>>().Object;
             var returnToHomeServiceLogger = new Mock<ILogger<ReturnToHomeService>>().Object;
+            var missionDefinitionServiceLogger = new Mock<ILogger<MissionDefinitionService>>().Object;
+            var lastMissionRunServiceLogger = new Mock<ILogger<LastMissionRunService>>().Object;
 
             var configuration = WebApplication.CreateBuilder().Configuration;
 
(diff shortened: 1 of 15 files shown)


backend/api/Services/MissionSchedulingService.cs
```diff
@@ -53,37 +53,36 @@ public async Task StartNextMissionRunIfSystemIsAvailable(string robotId)
...
+            if (!await localizationService.RobotIsLocalized(robotId) && !missionRun.IsLocalizationMission())
             {
                 logger.LogError("Tried to schedule mission {MissionRunId} on robot {RobotId} before the robot was localized, scheduled missions will be canceled", missionRun.Id, missionRun.Robot.Id);
                 try { await CancelAllScheduledMissions(robot.Id); }
                 catch (RobotNotFoundException) { logger.LogError("Failed to cancel scheduled missions for robot {RobotId}", robot.Id); }
                 return;
             }
 
-            if (!missionRun.IsLocalizationMission() && !await localizationService.RobotIsOnSameDeckAsMission(missionRun.Robot.Id, missionRun.Area.Id))
+            if (!missionRun.IsLocalizationMission() && !await localizationService.RobotIsOnSameDeckAsMission(robot.Id, missionRun.Area.Id))
```
tsundvoll, line 84: Should `robotId` been used instead of `robot.Id`? For consistency, look over the whole function

backend/api.test/EventHandlers/TestMissionEventHandler.cs
```diff
@@ -194,14 +196,14 @@ public async void NewMissionIsStartedWhenRobotBecomesAvailable()
         }
 
         [Fact]
-        public async void NoMissionIsStartedIfQueueIsEmptyWhenRobotBecomesAvailable()
+        public async void ReturnToHomeMissionIsStartedIfQueueIsEmptyWhenRobotBecomesAvailable()
```
tsundvoll, line 214: Is this test testing what the name suggests?

Seems to be missing a check that a return to home mission actually is started when the robot becomes available and no missions is pending.

Consider making a new issue for this

oysand: See #1445

oysand:
See if this needs to be addressed #1446

oysand:
> See if this needs to be addressed #1446

Added a fix to the refactor commit: [Refactor OnRobotAvailable and localization feature](https://github.com/equinor/flotilla/pull/1355/commits/3f2558d656de68cb49cb97f038f903b3ca43d29f)

tsundvoll (approved):
When this is tested locally, I am happy with the changes

PR #1362, opened 2024-01-26
Remove use of Task.InspectionTarget

Eddasol:
* Task.InspectionTarget should be removed as it is replaced by Inspection.InspectionTarget, but first existing custom missions needs to be adapted to the change. It will temporary be set to optional to allow removal in custom missions.
* If a task has several inspections with different inspectionTargets, the current solution will display them all with the same task number in the mission map. Improvement for this can be solved as new issue.


andchiind (approved):
LGTM. It could be good to have error handling if a certain version of ISAR requires an inspectionTarget, but it might also be worthwhile to do this in the respective ISAR implementations.

PR #1363, opened 2024-01-26
Add functionality for resetting robots

### LINKED ISSUE #1348: Create "Set robot to idle"-endpoint
Endpoint that sets robot to available and current area to null. 

Consider also:
Stopping all ongoing missions and emptying the queue if there are

This would be good for getting back to a usable state if some bug occur

mrica-equinor:
Closes #1348

### DIFF
--- a/backend/api/Controllers/RobotController.cs
+++ b/backend/api/Controllers/RobotController.cs
@@ -13,6 +13,7 @@ namespace Api.Controllers
         ILogger<RobotController> logger,
         IRobotService robotService,
         IIsarService isarService,
+        IMissionSchedulingService missionSchedulingService,
         IRobotModelService robotModelService
     ) : ControllerBase
     {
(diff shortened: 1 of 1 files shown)


backend/api/Controllers/RobotController.cs
```diff
@@ -637,5 +639,86 @@ public async Task<ActionResult> ResumeMission([FromRoute] string robotId)
             return Ok(missionRun);
         }
 
+        /// <summary>
+        ///   Empies the mission queue for the robot, stops the ongoing mission and sets the robot to available. Current area is set to null
```
aestene, line 643:
```suggestion
        ///   Empties the mission queue for the robot, stops the ongoing mission, sets the robot to available and current area is set to null
```

backend/api/Controllers/RobotController.cs
```diff
@@ -637,5 +639,86 @@ public async Task<ActionResult> ResumeMission([FromRoute] string robotId)
...
+
+            try { await robotService.UpdateCurrentArea(robot.Id, null); }
+            catch (RobotNotFoundException)
+            {
+                string errorMessage = $"Failed to set current area to null for robot with id {robotId} because the robot was not found";
+                logger.LogWarning("{Message}", errorMessage);
+                return NotFound(errorMessage);
+            }
+
+            missionSchedulingService.TriggerRobotAvailable(new RobotAvailableEventArgs(robot.Id));
```
Eddasol, line 716: Doesn't this happen by itself when all missions are cancelled and status becomes available? Do we risk it running twice now?

mrica-equinor: I am not sure, can have a check on this later today

mrica-equinor: You were right so I removed it

backend/api/Controllers/RobotController.cs
```diff
@@ -637,5 +638,81 @@ public async Task<ActionResult> ResumeMission([FromRoute] string robotId)
...
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(StatusCodes.Status404NotFound)]
+        [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+        public async Task<ActionResult> ResetRobot(
+            [FromRoute] string robotId
+        )
+        {
+            Console.WriteLine(robotId);
+            Console.WriteLine("oooooooooo....................");
```
andchiind, line 661: Don't think these are load-bearing printouts so they can probably be deleted

andchiind (approved):
Looks good otherwise. Make sure to remove the printouts before merging.

backend/api/Controllers/RobotController.cs
mrica-equinor: Damn can't believe i forgot these

PR #1372, opened 2024-01-29
Fix confirm pressure/battery dialogs

Eddasol:
Assumes bar as unit for all pressure values. Is dependent on https://github.com/equinor/flotilla/pull/1343


Eddasol:
If pressure is too low:
image

If battery is too low:
image

Currently these dialogs won't pop up if scheduling echo missions, but will pop up if insufficient values when scheduling through inspection plan or rerunning mission

frontend/src/components/Displays/ConfirmScheduleDialogs/ConfirmScheduleDialog.tsx
```diff
@@ -1,8 +1,11 @@
 import { Robot } from 'models/Robot'
```
andchiind, line 1: Is the second commit really needed since it seems to mainly just fix the previous commit? Can they not be merged?

frontend/src/components/Displays/ConfirmScheduleDialogs/LocalizationVerification/ConfirmLocalizationDialog.tsx
```diff
@@ -0,0 +1,76 @@
...
+}: ConfirmLocalizationDialogProps) => {
+    const { TranslateText } = useLanguageContext()
+    const [isCheckboxClicked, setIsCheckboxClicked] = useState<boolean>(false)
+    const [newDeck, setNewDeck] = useState<Deck>()
+
+    useEffect(() => {
+        BackendAPICaller.getDecks().then(async (decks: Deck[]) => {
+            setNewDeck(decks.find((deck) => deck.deckName === newDeckName))
+        })
+    }, [newDeckName])
```
andchiind, line 31: These should be retrieved from the installationContext

frontend/src/components/Displays/ConfirmScheduleDialogs/ConfirmScheduleDialog.tsx
```diff
@@ -0,0 +1,58 @@
...
+        return <InsufficientPressureDialog robot={robot} cancel={closeDialog} />
+    } else {
+        return (
+            <ScheduleMissionWithLocalizationVerificationDialog
+                scheduleMissions={scheduleMissions}
+                closeDialog={closeDialog}
+                robotId={robot!.id}
+                missionDeckNames={missionDeckNames}
+            />
+        )
```
andchiind, line 56: Not sure I agree with removing the generic final page for starting the mission, since now we assume that we will always need to localise, which is not the case.

Eddasol: The uses has to click "queue the mission" prior to this so I don't see the point of an additional dialog if localization, battery and pressure is all good.

frontend/src/components/Displays/ConfirmScheduleDialogs/ConfirmScheduleDialog.tsx
Eddasol: Wasn't sure if we wanted "confirm"-dialog or "this is insufficient values"-dialog. Wanted to keep an easy way to revert. Can squash of current design is agreed on

frontend/src/language/no.json
```diff
@@ -207,23 +207,14 @@
...
-    "which is above the specified minimum": "som er høyere enn anbefalt minimum",
-    "which is above the suggested minimum": "som er mer enn anbefalt minimum",
-    "Failed to schedule mission": "Kunne ikke legge til oppdrag"
+    "Failed to schedule mission": "Kunne ikke planlegge oppdrag",
+    "Current pressure value for": "Nåværende trykk for",
+    "Current battery value for": "Nåværende batteri for",
+    "is": "er",
+    "This is below recommended lower limit of": "Dette er under anbefalt nedre grense på",
+    "This is above recommended upper limit of": "Dette er under anbefalt øvre grense på",
+    "pressure warning": "Trykk advarsel",
```
prasm313, line 216:
```suggestion
    "pressure warning": "Advarsel: Trykk",
```

frontend/src/language/no.json
```diff
@@ -207,23 +207,14 @@
...
+    "battery warning": "Batteri advarsel",
```
prasm313, line 217:
```suggestion
    "battery warning": "Advarsel: Batteri",
```

frontend/src/components/Displays/ConfirmScheduleDialogs/InsufficientValueDialogs.tsx
```diff
@@ -0,0 +1,111 @@
...
+            <Dialog.Content>
+                <VerticalContent>
+                    <Typography>{warningText}</Typography>
+                    <Typography>{actionText}</Typography>
+                    <Dialog.Actions>
+                        <HorizontalContent>
+                            <Button variant="outlined" onClick={cancel}>
+                                {TranslateText('Close')}
+                            </Button>
+                        </HorizontalContent>
```
mrica-equinor, line 105: Just picking here but probably HorizontalContent can be merged with a unique style of type Button, but this can be left for later refactoring as well

mrica-equinor:
> LGTM We should add a dialog that informs the user if both battery and pressure level are out of range. Can be solved in a different PR.

Agree with adding a third dialog so the operator doesn't go out to pressurize the robot and then when comes back realizes that has to go out again to charge it as well. But agree that can be done in another PR

frontend/src/components/Displays/ConfirmScheduleDialogs/InsufficientValueDialogs.tsx
Eddasol: HorizontalContent can be removed is it has been changed from two to one button!

PR #1373, opened 2024-01-29
Refactor naming of type of mission tasks for the case of return to home

### LINKED ISSUE #1350: Return to home mission cannot be cancelled
**Describe the bug**
Currently return to home mission cannot be cancelled. If it is cancelled the status will become available and a new return to home mission will be scheduled. 

**To Reproduce**
Cancel return to home mission

**Expected behavior**
If mission cannot be cancelled the stop button should not be available

**Screenshots**

mrica-equinor:
Closes #1350


backend/api/Database/Models/MissionTask.cs
```diff
@@ -203,6 +203,6 @@ public enum MissionTaskType
     {
         Inspection,
         Localization,
-        DriveTo
+        ReturnHome
```
andchiind, line 206: Does this remove the ability to have DriveTo tasks, ie. tasks where we tell the robot to go somewhere, but we don't give it a target? I think there are several use-cases for these, and DriveTo is already supported in some Isar implementations. Could we instead add a new task type?

mrica-equinor: I checked this with Afonso, DriveTo was created specifcally for the return to home functionality and it is not used anywhere in the code that is not related to return to home. At the time he wanted to change the name but went on vacation/offshore before he could do the refactor

andchiind: That is odd, since missions from echo use it in other parts of their missions. I think we may need to have a bigger discussion around this.

andchiind: Sorry, it seems I have misunderstood. I see now that this renaming will not affect the naming in the IsarTask

andchiind (approved):

PR #1374, opened 2024-01-29
Display correct number of tasks in MissionQueueCard

UsamaEquinorAFK:


frontend/src/components/Pages/FrontPage/MissionOverview/MissionQueueCard.tsx
```diff
@@ -120,7 +120,7 @@ export const MissionQueueCard = ({ order, mission, onDeleteMission }: MissionQue
     const [confirmDeleteDialogOpen, setConfirmDeleteDialogOpen] = useState<boolean>(false)
     const fillColor = tokens.colors.infographic.primary__energy_red_21.hex
     let numberOfTasks = 0
-    mission.tasks.forEach((task) => (numberOfTasks += task.inspections.length))
+    mission.tasks.forEach((task) => (numberOfTasks += 1))
```
Eddasol, line 123: just use tasks.length?

UsamaEquinorAFK: Nope- that is a loop my solution gives correct answer. You can test both ways.

UsamaEquinorAFK: Removed loop and assigned const directly now =)

Eddasol (approved):
LGTM :)

PR #1378, opened 2024-01-30
Removed quotation mark in environment files

tsundvoll:
A difference in Windows / Mac seems to interpret strings from environment variables differently. Removing quotation marks seems to work on both OS


mrica-equinor (approved):
LGTM :)

andchiind (approved):
😲

PR #1382, opened 2024-01-31
Use missionRunResponse in signalr messages

### LINKED ISSUE #1381: StartTime and currentArea on missionPage is sometimes wrong/missing
**Describe the bug**
Initially Area and Start time is correct but when the mission updates the area becomes undefined and start time becomes an hour off. 

Likely cause for area disappearing is that API call to get missionRun provides a MissionRunResponse (which is what the frontend expects), while signalR provides a MissionRun. 

**To Reproduce**
Start mission and go to mission page - Note area and start time
Wait for mission to update - Note area and start time

**Expected behavior**

**Screenshots**
Initially:
image

After update:
image

andchiind:
Closes #1381

### DIFF
--- a/backend/api.test/Database/DatabaseUtilities.cs
+++ b/backend/api.test/Database/DatabaseUtilities.cs
@@ -51,6 +51,7 @@ namespace Api.Test.Database
                 Status = MissionStatus.Pending,
                 DesiredStartTime = DateTime.Now,
                 Area = area,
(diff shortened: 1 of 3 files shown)


andchiind:
Seems to work fine now:
image
image
image

mrica-equinor (approved):
LGTM :)

PR #1390, opened 2024-02-05
Make installationCode lowercase upon access from database

UsamaEquinorAFK:


backend/api/Database/Models/Installation.cs
```diff
@@ -16,6 +18,12 @@ public class Installation
 
         [Required]
         [MaxLength(10)]
-        public string InstallationCode { get; set; }
+        public string InstallationCode
+        {
+            get => _installationCode?.ToLower(System.Globalization.CultureInfo.CurrentCulture)
```
andchiind, line 23: Since we currently mainly convert it to upper case when making comparisons, could it be good to convert it to upper case here and then make sure we never convert it to lower case anywhere? Or vice versa by always using lower case?

UsamaEquinorAFK: Every installationCode is uppercase now.

andchiind (approved):
LGTM

PR #1396, opened 2024-02-07
Fix flexibility and readability for error cards

prasm313:


prasm313:
End result:
image

mrica-equinor (approved):
LGTM :)

PR #1400, opened 2024-02-08
Remove Inspection Target

UsamaEquinorAFK:


prasm313 (commented):
LGTM, tested without any bugs. Suggest that @Afonso-2403 takes a look before it gets approved.

Afonso-2403 (approved):
LGTM, let's not merge yet until we have the workflow in place to update the existing missions

PR #1417, opened 2024-02-13
Fix localization bug

UsamaEquinorAFK:


backend/api/Services/MissionRunService.cs
```diff
@@ -114,9 +114,18 @@ public async Task<IList<MissionRun>> ReadMissionRunQueue(string robotId)
...
-                .FirstOrDefaultAsync(missionRun => missionRun.Robot.Id == robotId && missionRun.Status == MissionStatus.Pending);
+                .FirstOrDefaultAsync();
+
+
+            nextScheduledMissionRun ??= await GetMissionRunsWithSubModels()
+                    .Where(missionRun => missionRun.Robot.Id == robotId && missionRun.Status == MissionStatus.Pending)
+                    .OrderBy(missionRun => missionRun.DesiredStartTime)
+                    .FirstOrDefaultAsync();
+
+            return nextScheduledMissionRun;
```
andchiind, line 128: I like the code and the approach, but could it maybe be a bit more readable to have line 117 define a variable called something like "nextScheduledLocalisation" and then have an if statement instead of the ?? operator check if it is set? That way it would be clear that we first find the localisation mission, and if this does not exist we find the next mission run and return that.

UsamaEquinorAFK: Thanks- Discussed with @oysand and we will move the functionality one level up in the event handler instead. Will require more refactoring.

UsamaEquinorAFK: Will close this PR and make a new one with new approach.

PR #1427, opened 2024-02-19
Run emergency missions first even if queue is not frozen

UsamaEquinorAFK:


oysand:
To be handled in #1355

PR #1436, opened 2024-02-23
Update AdaptiveCardGeneration function

UsamaEquinorAFK:


backend/api.test/EventHandlers/TestInspectionFindingsEventHandler.cs
```diff
@@ -1,26 +0,0 @@
...
-using System.Globalization;
-using Api.EventHandlers;
-using Xunit;
-
-namespace Api.Test.EventHandlers
-{
-    public class TestInspectionFindingsEventHandler
-    {
-        [Fact]
-        public void TestGenerateReportFromFinding()
```
oysand, line 11: Why is the test removed?

backend/api/EventHandlers/InspectionFindingEventHandler.cs
```diff
@@ -103,54 +103,51 @@ private async Task<List<Finding>> GenerateFindingsList(List<InspectionFinding> i
...
+        public static string? GenerateAdaptiveCard(string title, int numberOfFindings, List<Finding> findingsReports)
         {
-            var reportBuilder = new StringBuilder("Findings Report:");
-            reportBuilder.Append(Environment.NewLine);
-            string dateFormat = "dd/MM/yyyy HH:mm";
-            var formatProvider = CultureInfo.InvariantCulture;
-            foreach (var finding in findingsReports)
+            var findingsJsonArray = new JArray();
+
+            for (int i = 0; i < findingsReports.Count; i++)
```
oysand, line 110: Could we use foreach(finding in findingsReports)?

backend/api/EventHandlers/InspectionFindingEventHandler.cs
```diff
@@ -103,54 +103,51 @@ private async Task<List<Finding>> GenerateFindingsList(List<InspectionFinding> i
...
-                                    ""msteams"": {{
-                                        ""displayStyle"": ""full""
-                                    }}
-                                }}
-                            }}
-                        ]
-                    }}";
+            var sections = new JArray(
+                new JObject(
+                    new JProperty("activityTitle", $"Inspection report for \"{findingsReports[0].PlantName}\""),
```
oysand, line 130: We do we use the [0] and not [i]?

oysand (changes requested):
See comments

backend/api/EventHandlers/InspectionFindingEventHandler.cs
```diff
@@ -103,55 +106,125 @@ private async Task<List<Finding>> GenerateFindingsList(List<InspectionFinding> i
...
+                cardBody.AppendLine("},");
+
             }
-            return reportBuilder.ToString();
-        }
 
-        public static string GenerateAdaptiveCard(string messageContent, ILogger<InspectionFindingEventHandler> logger)
-        {
-            logger.LogInformation("Message content for Adaptive card: {message}", messageContent);
-            string adaptiveCardJson = $@"{{
```
tsundvoll, line 125: This old way seems more readable to me, what is the reason we can not do it this way?

backend/api/EventHandlers/InspectionFindingEventHandler.cs
```diff
@@ -7,14 +7,16 @@
...
+using Newtonsoft.Json.Linq;
 
 namespace Api.EventHandlers
 {
     public class InspectionFindingEventHandler(IConfiguration configuration,
     IServiceScopeFactory scopeFactory,
     ILogger<InspectionFindingEventHandler> logger) : BackgroundService
     {
-        private readonly string _cronExpression = "30 14 * * * ";
+        private readonly string _cronExpression = "19 14 * * * ";
```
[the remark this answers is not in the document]
tsundvoll: This is a standard cron-job format, used in many other settings as well, so I think it is good to use it here as well. However, a comment above the line, explaining what the chosen setting means would be useful

backend/api.test/EventHandlers/TestInspectionFindingsEventHandler.cs
UsamaEquinorAFK: Because we are not using that function anymore, I can add more test for the existing functions.

backend/api/EventHandlers/InspectionFindingEventHandler.cs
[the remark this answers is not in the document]
UsamaEquinorAFK: Cronjob standard format,  do we need this in the config instead ?

backend/api/EventHandlers/InspectionFindingEventHandler.cs
UsamaEquinorAFK: Well it will always be the same plant per report but should be i I agree :)

UsamaEquinorAFK: Well this is actually outside the for loop :)

backend/api/EventHandlers/InspectionFindingEventHandler.cs
UsamaEquinorAFK: Done.

backend/api.test/EventHandlers/TestInspectionFindingsEventHandler.cs
UsamaEquinorAFK: Added new test.

backend/api/EventHandlers/InspectionFindingEventHandler.cs
UsamaEquinorAFK: Have a for each loop now instead, keeping this one as is or this line can also be removed later.

backend/api/EventHandlers/InspectionFindingEventHandler.cs
UsamaEquinorAFK: Agreed will add now.

backend/api/EventHandlers/InspectionFindingEventHandler.cs
```diff
@@ -7,14 +7,16 @@
 using Azure.Identity;
 using Azure.Security.KeyVault.Secrets;
 using NCrontab;
+using Newtonsoft.Json;
```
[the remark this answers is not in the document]
UsamaEquinorAFK: Can be handled in a seperate issue. Will mae an issue at the end of today.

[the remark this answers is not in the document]
UsamaEquinorAFK: Resolving for now.

backend/api/EventHandlers/InspectionFindingEventHandler.cs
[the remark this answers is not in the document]
oysand: Nah, I think it can be kept here :+1:

mrica-equinor (approved):
LGTM :)

PR #1455, opened 2024-03-07
Remove unnecessary icons in header

### LINKED ISSUE #1444: Remove unnecessary icons in Flotilla header
Suggest to remove the icons in the Flotilla header if we do not have specific plans to use them in the near future.

Image

mrica-equinor:
Closes #1444

### DIFF
--- a/frontend/src/components/Header/Header.tsx
+++ b/frontend/src/components/Header/Header.tsx
@@ -12,13 +12,6 @@ const StyledTopBar = styled(TopBar)`
     margin-bottom: 2rem;
     align-items: center;
     box-shadow: none;
-
-    @media (max-width: 700px) {
-        display: flex;
-        justify-self: start;
-        flex-direction: column;
-        height: 90px;
-    }
 `
 
 const StyledWrapper = styled.div`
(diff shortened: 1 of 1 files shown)


mrica-equinor:
image

image

prasm313:
When the dimensions of the screen is under 700px the icon jumps to the middle (see line 15 in Header.tsx). Not sure if any of our devices would be affected by this.

image

mrica-equinor:
> When the dimensions of the screen is under 700px the icon jumps to the middle (see line 15 in Header.tsx). Not sure if any of our devices would be affected by this.
> 
> image

Yeah there was a bit of code that I could have deleted, should be good now :)

prasm313:
Last comment resolved👍

- Would have to scroll to see the "NO" language selector (seems to me that this is because of the header).

- There is no margin for the "NO" on the right.

image

mrica-equinor:
> Last comment resolved👍
> 
> * Would have to scroll to see the "NO" language selector (seems to me that this is because of the header).
> * There is no margin for the "NO" on the right.
> 
> image

Yeah I tested with iPhone XR and didn't notice that the iPhone 12 screen is smaller. The margin left to the platform button was too big for the screen size. I updated this and think it's good now

oysand (approved):
LGTM

PR #1462, opened 2024-03-11
Only add quotations in dev

tsundvoll:
not for staging and prod


betaniat (approved):
not tested as part of the review process (for now)

PR #1465, opened 2024-03-11
Add dev prefix to dev image

tsundvoll:


oysand (approved):
LGTM, will be tested with the dev update this will trigger

PR #1471, opened 2024-03-13
Manage UTC/local time inconsistency

### LINKED ISSUE #1442: Time used is not correct during the mission
**Describe the bug**
During a mission, time used is not correct for all missions. Typically, it shows 60-80 minutes instead of 0-20 minutes. I have not confirmed it, but suspect that there might be an hour offset. The time used is shown correctly for the mission in the history page. 

**To Reproduce**
Steps to reproduce the behavior: Start a mission and monitor the time used on the mission page during the mission. 

**Expected behavior**
The time used should be correct. 

**Screenshots**

andchiind:
Closes #1442


frontend/src/components/Pages/MissionPage/MissionHeader/MissionHeader.tsx
```diff
@@ -100,18 +101,18 @@ const getStartUsedAndRemainingTime = (
...
         usedTimeInMinutes = mission.startTime
             ? differenceInMinutes(new Date(mission.endTime), new Date(mission.startTime))
             : 0
         remainingTime = 'N/A'
     } else if (mission.startTime) {
-        startTime = format(new Date(mission.startTime), 'HH:mm')
-        startDate = format(new Date(mission.startTime), 'dd/MM/yyy')
+        startTime = formatDateTime(new Date(mission.startTime), 'HH:mm')
+        startDate = formatDateTime(new Date(mission.startTime), 'dd/MM/yyy')
         usedTimeInMinutes = differenceInMinutes(Date.now(), new Date(mission.startTime))
```
Eddasol, line 116: Need to use formattedDateTime/convertUTCDateToLocalDate in this comparison as well to avoid used time having an an offset of on hour

image

Eddasol (approved):
LGTM :)

frontend/src/utils/StringFormatting.tsx
```diff
@@ -21,7 +24,9 @@ export const getInspectionDeadline = (
     return deadline
 }
 
-export const getDeadlineInDays = (deadlineDate: Date) => {
+export const getDeadlineInDays = (deadlineDate: Date): number =>
     // The magical number on the right is the number of milliseconds in a day
```
Eddasol, line 28: create a variable called milliSecondsInADay and use that instead?

PR #1476, opened 2024-03-20
Remove fix for UTC time. Not postgres compatible

andchiind:
This undoes the fix for UTC time in the frontend. Although the DateTime "kind" is set to unspecified when fetching from the database, we cannot set values in the database to any kind other than "UTC". When we send a datetime with kind "UTC" to the frontend however, it is interpreted as local time, which causes a time offset.

Reopens https://github.com/equinor/flotilla/issues/1442


oysand (approved):
LGTM

PR #1489, opened 2024-03-22
Rename enabled to isarConnected

Eddasol:


frontend/src/api/ApiCaller.tsx
```diff
@@ -134,7 +134,7 @@ export class BackendAPICaller {
     static async getEnabledRobots(): Promise<Robot[]> {
         const path: string = 'robots'
         const result = await BackendAPICaller.GET<Robot[]>(path).catch(BackendAPICaller.handleError('GET', path))
-        return result.content.filter((robot) => robot.enabled)
```
Eddasol, line 137: Note that the name of this function (getEnabledRobots) hasn't been changed, as this function will be rewritten to sort on deprecated instead in separate PR

backend/api/Services/RobotService.cs
```diff
@@ -268,7 +268,7 @@ public async Task<IList<Robot>> ReadRobotsForInstallation(string installationCod
                 .ToListAsync();
         }
 
-        public async Task SetRobotOffline(string robotId)
+        public async Task SetRobotToIsarDisconnected(string robotId)
```
oysand, line 271: This function does more than it says, we split the functionality or set the function to explicitly state what it does

Eddasol: Handled by issue https://github.com/equinor/flotilla/issues/1493

oysand (approved):
LGTM

PR #1497, opened 2024-03-25
Add support for arguments in translations

### LINKED ISSUE #1496: Consider allowing more dynamic translations
Currently the translations work by replacing a given string with another one. This allows us to handle most situations, but there are a couple of limitations. One limitation is that different languages place words in different parts of the sentences in their grammar, so if we want to translate that part separately then we currently instead have to make one of the translations sound awkward.

Here is an example of places where we need to create compound translations:
image

Another limitation is when we receive strings from the backend which we may want to translate, such as alerts. In this case it could for instance be convenient to simply paste the provided string into the TranslateText() function, but this will not work if the provided string contains variables, such as robot names.

What I think might be the simplest solution is to allow strings to make use of arguments using a typical string formatting syntax:

image

This allows us to change their location and ordering in a string, avoiding the need for translating parts of a sentence at a time.

N/A, the strings are anyways not used to run code. We don't even need regex.

andchiind:
Closes #1496


mrica-equinor (approved):
LGTM :)

PR #1499, opened 2024-03-25
Improve alert banners

### LINKED ISSUE #1495: Improve banners in the frontend
Currently, the frontend banners have designs that do not match. The error banner doesn't have a sketch in Figma and therefore the design should be aligned with the Safe Zone banners. 

To-do:
- Make the design of Error banner more similar to the Safe Zone banner.
- Change Safe Zone banner text to robot already at safe zone when robot reaches safe zone.
- Find lighter colors for the Safe Zone banner (until we get fnal colors from UX design, this has been talked in the meetings).
- Improve the hover colors on the buttons.

It's just CSS, the functionality will be the same.

mrica-equinor:
Closes #1495

### DIFF
--- a/backend/api.test/Mocks/SignalRServiceMock.cs
+++ b/backend/api.test/Mocks/SignalRServiceMock.cs
@@ -27,5 +27,10 @@ namespace Api.Services
         {
             return;
         }
+
+        public void ReportSafeZoneSuccessToSignalR(Robot robot, string message)
+        {
+            return;
+        }
     }
 }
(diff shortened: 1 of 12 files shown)


andchiind (approved):
The banners for "Failed to schedule return to home for robot x" replace each other, but I assume that this might be intentional. Seems to work fine otherwise.

mrica-equinor:
> The banners for "Failed to schedule return to home for robot x" replace each other, but I assume that this might be intentional. Seems to work fine otherwise.

I'm not sure if it's intentional, it was happening before so I don't think it's related to my PR. Can remember to take a look :)

PR #1504, opened 2024-04-02
Prevent many tasks being indexed with index 1

### LINKED ISSUE #1453: Mission task queue bug
**Describe the bug**
The tasks queue in a mission is not correctly numerated. Missing number 2 in the list.

**To Reproduce**
Run a mission with several tags.

**Expected behavior**
Should have correct numbering.

**Screenshots**
Image

andchiind:
Explanation for the bug:
When fetching missions from Echo, the first task is given the "PlanOrder"/"SortingOrder" 0 and all the rest get 2147483647. When sending this to ISAR we then currently ignore PlanOrder and just assume the order we get it from the backend is correct:
image
(screenshot from IsarMissionDefinition.cs, line 31)

PlanOrder/SortingOrder seems to only ever be used on the frontend to display the order. It might therefore be safe to ignore this value and just use the order of the items in the list of tasks, since that is what we do for ISAR anyways.

We could potentially also close issue #1453 if we just ignore "taskOrder" in the frontend and only use the index, but I would like some feedback on this first. UPDATE: we now ignore Echo taskOrder, so this PR will also close #1453 .

This is a draft until a more permanent solution is created, which maintains consistent indexing in the backend. This can either come from a solution in Echo, or from us setting the taskOrder as soon as we fetch a mission from echo.

### DIFF
--- a/backend/api/Services/EchoService.cs
+++ b/backend/api/Services/EchoService.cs
@@ -88,8 +88,16 @@ namespace Api.Services
         {
             var tags = new List<EchoTag>();
 
-            foreach (var planItem in planItems)
+            var indices = new HashSet<int>();
+            bool inconsistentIndices = false;
+
+            for (int i = 0; i < planItems.Count; i++)
             {
+                var planItem = planItems[i];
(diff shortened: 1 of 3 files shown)


Eddasol (approved):
LGTM :)

PR #1511, opened 2024-04-04
Buttons in tables and minor fixes

prasm313:


frontend/src/utils/InfoText.tsx
```diff
@@ -1,13 +1,19 @@
...
+    align-items: center;
+`
 
 export const SmallScreenInfoText = () => {
     const { TranslateText } = useLanguageContext()
     return (
-        <div id="SmallScreenInfoText">
-            <Icon name={Icons.Info} size={24}></Icon>
+        <StyledDiv id="SmallScreenInfoText">
+            <Icon name={Icons.Info}> </Icon>
```
andchiind, line 15:
```suggestion
            <Icon name={Icons.Info} />
```
We can also do this if you want to shorten it further, but it's entirely subjective.

andchiind (approved):
Tested it, seems to work well

frontend/src/utils/InfoText.tsx
prasm313: Thanks, I like this

PR #1513, opened 2024-04-04
Avoid localization mission if it is not supported

andchiind:


backend/api/EventHandlers/MissionEventHandler.cs
```diff
@@ -75,37 +75,41 @@ private async void OnMissionRunCreated(object? sender, MissionRunCreatedEventArg
...
+                if (missionRun.Robot.RobotCapabilities != null && !missionRun.Robot.RobotCapabilities.Contains(RobotCapabilitiesEnum.localize))
                 {
-                    _scheduleLocalizationSemaphore.Release();
-                    return;
-                }
-                if (await MissionService.OngoingLocalizationMissionRunExists(missionRun.Robot.Id))
-                {
-                    _scheduleLocalizationSemaphore.Release();
-                    return;
+                    await RobotService.UpdateCurrentArea(missionRun.Robot.Id, missionRun.Area);
```
mrica-equinor, line 80: This is just a personal preferente because i don't like else's but can leave it as is if you want
```suggestion
                    await RobotService.UpdateCurrentArea(missionRun.Robot.Id, missionRun.Area);
                     _scheduleLocalizationSemaphore.Release();
                    return;
```

mrica-equinor (approved):
LGTM :)

backend/api/EventHandlers/MissionEventHandler.cs
andchiind: Wait I think I misread. At this point in the code we haven't actually acquired the lock so we can't release it. We only acquire it if we do need to run the localisation mission

mrica-equinor: Can just do the return then

PR #1532, opened 2024-04-08
Enable user to delete mission run

oysand:


andchiind (approved):
LGTM. Since there is a access role check performed in the Delete function in MissionRunService (when calling GetMissionRunsWIthSubModels) then this should be safe to allow. If a user does not have access to the installation then they will get a 404 status code.

PR #1536, opened 2024-04-09
Return error when using unsupported sensors

### LINKED ISSUE #1534: Return an error when trying to start a mission containing tasks not supported by the given robot
Return an error when trying to queue a mission containing tasks not supported by the given robot. For instance if trying to take a thermal video with a robot that does not have a thermal camera. For simplicity we can focus on the backend part of this task in this issue, but as long as we return a 400 status code then this will anyways be reported in the frontend.

We can check the tasks in the mission and compare them to the RobotCapabilites field in Robot.cs. If the list is null then we should assume that it is allowed.

N/A

andchiind:
Closes #1534 

Here is how it currently looks: 
image
If we want we can catch it in the frontend to make a more unique error for it.

### DIFF
--- a/backend/api/Services/LocalizationService.cs
+++ b/backend/api/Services/LocalizationService.cs
@@ -99,8 +99,15 @@ namespace Api.Services
             };
             await mapService.AssignMapToMission(localizationMissionRun);
 
-            logger.LogWarning("Starting localization mission");
-            await missionRunService.Create(localizationMissionRun, triggerCreatedMissionRunEvent: false);
+            try
+            {
+                logger.LogWarning("Starting localization mission");
+                await missionRunService.Create(localizationMissionRun, triggerCreatedMissionRunEvent: false);
+            }
+            catch (UnsupportedRobotCapabilityException)
+            {
+                logger.LogError($"Unsupported robot capability detected when starting localisation mission for robot {localizationMissionRun.Robot.Name}. This should not happen.");
+            }
             return localizationMissionRun;
         }
 
--- a/backend/api/Services/MissionSchedulingService.cs
+++ b/backend/api/Services/MissionSchedulingService.cs
@@ -252,7 +252,14 @@ namespace Api.Services
                 Map = new MapMetadata()
             };
 
-            await missionRunService.Create(missionRun);
+            try
+            {
+                await missionRunService.Create(missionRun);
+            }
+            catch (UnsupportedRobotCapabilityException)
+            {
+                logger.LogError($"Unsupported robot capability detected when driving to safe position for robot {missionRun.Robot.Name}. This should not happen.");
+            }
         }
 
         public bool MissionRunQueueIsEmpty(IList<MissionRun> missionRunQueue)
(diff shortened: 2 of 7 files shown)


backend/api/Services/MissionSchedulingService.cs
```diff
@@ -252,7 +252,14 @@ public async Task ScheduleMissionToDriveToSafePosition(string robotId, string ar
...
             };
 
-            await missionRunService.Create(missionRun);
+            try
+            {
+                await missionRunService.Create(missionRun);
+            }
+            catch (UnsupportedRobotCapabilityException)
+            {
+                logger.LogWarning($"Unsupported robot capability detected when driving to safe position for robot {missionRun.Robot.Name}. This should not happen.");
```
Christdej, line 261: If this should not happen, we should maybe LogError?

backend/api/Services/MissionSchedulingService.cs
```diff
@@ -315,7 +322,14 @@ private async Task MoveInterruptedMissionsToQueue(IEnumerable<string> interrupte
...
                 };
 
-                await missionRunService.Create(newMissionRun);
+                try
+                {
+                    await missionRunService.Create(missionRun);
+                }
+                catch (UnsupportedRobotCapabilityException)
+                {
+                    logger.LogWarning($"Unsupported robot capability detected when restarting interrupted missions for robot {missionRun.Robot.Name}. This should not happen.");
```
Christdej, line 331: Not sure the service should be catching this? Can perhaps catch in higher levels?

Christdej (changes requested):

backend/api/Services/MissionSchedulingService.cs
andchiind: I would argue that we shouldn't handle it inside the Create function since then we will not be able to detect the specific cause of the error further down the stack. This is useful to report it to the user. The Create function also will not necessarily understand why it is that we do not expect this exception with this kind of mission run, since it doesn't treat Localisation and Return Home missions differently from normal missions which we do expect to potentially cause this exception.

backend/api/Services/LocalizationService.cs
```diff
@@ -99,8 +99,15 @@ public async Task<MissionRun> CreateLocalizationMissionInArea(string robotId, st
...
-            logger.LogWarning("Starting localization mission");
-            await missionRunService.Create(localizationMissionRun, triggerCreatedMissionRunEvent: false);
+            try
+            {
+                logger.LogWarning("Starting localization mission");
+                await missionRunService.Create(localizationMissionRun, triggerCreatedMissionRunEvent: false);
+            }
+            catch (UnsupportedRobotCapabilityException)
+            {
+                logger.LogWarning($"Unsupported robot capability detected when starting localisation mission for robot {localizationMissionRun.Robot.Name}. This should not happen.");
```
Christdej, line 109: LogError perhaps?

Christdej (changes requested):

backend/api/Services/MissionSchedulingService.cs
Christdej: Yeah, good point

Christdej (approved):

PR #1553, opened 2024-04-12
Fix mix between UTC and local time

Eddasol:
Time varies between having a z at the and not having it. If there is a z, typescript will assume that the time is given as UTCtime and if not it will assume that its local time

Console log of time values and corresponding new Date(). Note the marked time and the corresponding time on the line above
image

Best would be to fix such that backend always sends the same but this is an alternative solution

### DIFF
--- a/frontend/src/utils/StringFormatting.tsx
+++ b/frontend/src/utils/StringFormatting.tsx
@@ -2,8 +2,13 @@ import { format } from 'date-fns'
 
 const millisecondsInADay = 8.64e7
 
-export const convertUTCDateToLocalDate = (date: Date): Date =>
-    new Date(date.getTime() - date.getTimezoneOffset() * 60 * 1000)
+export const convertUTCDateToLocalDate = (date: Date): Date => {
+    // If lastChar is Z, typescript assumes the date to be UTC
+    const lastChar = date.toString().slice(-1)
+    if (lastChar === 'Z') return new Date(date)
+    // If not, typescript assumes the date to be local time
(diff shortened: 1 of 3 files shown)


frontend/src/utils/StringFormatting.tsx
```diff
@@ -2,8 +2,13 @@ import { format } from 'date-fns'
 
 const millisecondsInADay = 8.64e7
 
-export const convertUTCDateToLocalDate = (date: Date): Date =>
-    new Date(date.getTime() - date.getTimezoneOffset() * 60 * 1000)
+export const convertUTCDateToLocalDate = (date: Date): Date => {
+    // If lastChar is Z, typescript assumes the date to be UTC
+    const lastChar = date.toString().slice(-1)
+    if (lastChar === 'Z') return new Date(date)
```
mrica-equinor, line 8: I'm confused, this function is called convert from UTC to Local but if we get a date in UTC we don't do anything and if we receive one in local time then we change it. Shouldn't it then be change local time to UTC ?

Eddasol: We always get the date as UTC. If there is a Z typescript assumes (correctly) that the time is UTC and will handle the conversion to local time. If not it will assume that the date already is local time and we need to remove the offset to get the correct local time

mrica-equinor: I thought this could be the case, a bit confusing with the naming but can't really think of a better way to make it more explicit. Whoever looks at it will have to dig in the code

mrica-equinor (approved):
LGTM :)

PR #1554, opened 2024-04-15
Move arm triggers banner failed return home

### LINKED ISSUE #1466: Move robot arm triggers Return to Home Mission
**Describe the bug**
Currently when we schedule a move robot arm mission, the robot becomes busy and then goes back to available. As a consequence the backend will try to schedule a return to home mission because the mission queue is empty. Since the robot current area is null, an error message is sent to the frontend saying "unable to schedule return to home mission". 

**To Reproduce**
Schedule a move arm mission for robot.

**Expected behavior**
Moving the robot arm should not trigger a return to home mission.

Suggested solution: Create a new mission type: "Move arm". Also relevant for other mission types: "Return home", "Localization".

**Screenshots**

Image

mrica-equinor:
Closes #1466

### DIFF
--- a/backend/api/EventHandlers/MqttEventHandler.cs
+++ b/backend/api/EventHandlers/MqttEventHandler.cs
@@ -224,6 +224,7 @@ namespace Api.EventHandlers
             var taskDurationService = provider.GetRequiredService<ITaskDurationService>();
             var lastMissionRunService = provider.GetRequiredService<ILastMissionRunService>();
             var missionSchedulingService = provider.GetRequiredService<IMissionSchedulingService>();
+            var signalRService = provider.GetRequiredService<ISignalRService>();
 
             var isarMission = (IsarMissionMessage)mqttArgs.Message;
 
@@ -255,7 +256,13 @@ namespace Api.EventHandlers
 
             if (flotillaMissionRun.IsLocalizationMission())
             {
-                if (flotillaMissionRun.Status != MissionStatus.Successful) { await robotService.UpdateCurrentArea(robot.Id, null); }
+                if (flotillaMissionRun.Status != MissionStatus.Successful)
+                {
+                    await robotService.UpdateCurrentArea(robot.Id, null);
+
+                    signalRService.ReportGeneralFailToSignalR(robot, "Failed Localization Mission", $"Failed localization mission for robot {robot.Name}.");
+                    _logger.LogError("Localization mission for robot '{RobotName}' failed.", isarMission.RobotName);
+                }
                 else { await robotService.UpdateCurrentArea(robot.Id, flotillaMissionRun.Area); }
             }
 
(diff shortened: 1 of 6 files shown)


backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -255,7 +256,13 @@ private async void OnIsarMissionUpdate(object? sender, MqttReceivedArgs mqttArgs
 
             if (flotillaMissionRun.IsLocalizationMission())
             {
-                if (flotillaMissionRun.Status != MissionStatus.Successful) { await robotService.UpdateCurrentArea(robot.Id, null); }
+                if (flotillaMissionRun.Status != MissionStatus.Successful)
+                {
+                    await robotService.UpdateCurrentArea(robot.Id, null);
+
+                    signalRService.ReportGeneralFailToSignalR(robot, "Failed Localization Mission", $"Failed localization mission for robot {robot.Name}.");
```
oysand, line 263: Should we have a more specific fail reported? ReportLocalizationMissionFailedToSignalR or something a long that?

mrica-equinor: We used to but then the format that is being sent inside the funtion is the same for all cases, that's why I created this general one that we can call and personalize the content with the second and third field. Otherwise we would have like 3 or more functions with different names that do exactly the same thing

backend/api/Services/MissionSchedulingService.cs
```diff
@@ -55,10 +55,17 @@ public async Task StartNextMissionRunIfSystemIsAvailable(string robotId)
                 logger.LogInformation("The robot was ready to start mission, but no mission is scheduled");
                 if (robot.MissionQueueFrozen) { return; }
 
+                if (!await localizationService.RobotIsLocalized(robotId))
+                {
+                    string infoMessage = $"Not scheduling a return to home mission as the robot {robotId} is not localized.";
```
oysand, line 60: Can this be written directly in the LogInformation?

andchiind: I personally think it's good to reduce line length for readability.

mrica-equinor: Don't have a strong preference on this

oysand: Me neither, can keep it as it is then

backend/api/Services/MissionSchedulingService.cs
```diff
@@ -55,10 +55,17 @@ public async Task StartNextMissionRunIfSystemIsAvailable(string robotId)
                 logger.LogInformation("The robot was ready to start mission, but no mission is scheduled");
                 if (robot.MissionQueueFrozen) { return; }
 
+                if (!await localizationService.RobotIsLocalized(robotId))
```
oysand, line 58: Do we have this covered in a test or should we write one to see what happens if the robot is not localized and robot available is triggered?

mrica-equinor: Test has been added

andchiind (approved):
LGTM

PR #1557, opened 2024-04-16
Fix scheduling of missions during return to home

### LINKED ISSUE #1464: Scheduling a new mission while the “Return to home” mission is ongoing
When a mission is scheduled while the robot is returning to home, the robot returns home and then executes the mission that was added to the queue. 

Correct behavior would be cancelling the ongoing return home mission, execute new additional mission(s), completing the additional mission if battery and pressure are good and then scheduling a return home mission.  

Robot battery will be optimized.

mrica-equinor:
Closes #1464


backend/api/Services/MissionRunService.cs
```diff
@@ -153,6 +153,19 @@ public async Task<IList<MissionRun>> ReadMissionRunQueue(string robotId)
             return nextScheduledMissionRun;
         }
 
+        public async Task<MissionRun?> ReadPendingOrOngoingOrPausedReturnToHomeMissionRuns(string robotId)
```
andchiind, line 156:
```suggestion
        public async Task<MissionRun?> ReadReturnToHomeMissionRuns(string robotId, IList<MissionStatus>? filterStatuses = null)
```
Could we maybe make the function more dynamic by having the MissionStatus values we are filtering with be an argument?

mrica-equinor: We can, we don't need it for now, but might be usefull for the future

andchiind: I just figure it would make the code more readable, since it's a bit hard to understand the function based on its name.

backend/api/Services/MissionRunService.cs
```diff
@@ -153,6 +153,26 @@ public async Task<IList<MissionRun>> ReadMissionRunQueue(string robotId)
             return nextScheduledMissionRun;
         }
 
+        public async Task<MissionRun?> ReadReturnToHomeMissionRuns(string robotId, IList<MissionStatus>? filterStatuses = null)
+        {
+            var missionFilter = ConstructFilter(new MissionRunQueryStringParameters
+            {
+                Statuses = filterStatuses as List<MissionStatus>,
```
andchiind, line 160: Does this cause errors if filterStatuses is null?

backend/api/Controllers/MissionSchedulingController.cs
```diff
@@ -317,6 +327,15 @@ ISourceService sourceService
                 }
             }
 
+            IList<MissionStatus> missionStatuses = [MissionStatus.Ongoing, MissionStatus.Pending, MissionStatus.Paused];
+            var returnToHomeMission = await missionRunService.ReadReturnToHomeMissionRuns(robot.Id, missionStatuses);
+            if (returnToHomeMission != null)
+            {
+                await missionSchedulingService.StopCurrentMissionRun(robot.Id);
```
andchiind, line 334: I think we need to catch MissionRunNotFoundException here, since we do not lock this thread and the mission might therefore finish before we reach this. Or we need to lock this thread

backend/api/Controllers/MissionSchedulingController.cs
```diff
@@ -179,6 +179,23 @@ ISourceService sourceService
                 missionRun.CalculateEstimatedDuration();
             }
 
+            IList<MissionStatus> missionStatuses = [MissionStatus.Ongoing, MissionStatus.Pending, MissionStatus.Paused];
+            var returnToHomeMission = await missionRunService.ReadReturnToHomeMissionRuns(robot.Id, missionStatuses);
```
oysand, line 183: This function feels both quite specific and trying to be general. Could we make it even more specific or more general? For example "ReadUnfinishedReturnToHomeMissionsRuns" or "ReadMissionRuns" and mission type is an input?

backend/api/Services/MissionRunService.cs
```diff
@@ -153,6 +153,26 @@ public async Task<IList<MissionRun>> ReadMissionRunQueue(string robotId)
             return nextScheduledMissionRun;
         }
 
+        public async Task<MissionRun?> ReadReturnToHomeMissionRuns(string robotId, IList<MissionStatus>? filterStatuses = null)
+        {
+            var missionFilter = ConstructFilter(new MissionRunQueryStringParameters
+            {
+                Statuses = filterStatuses as List<MissionStatus> ?? new List<MissionStatus>([]),
```
oysand, line 160: if filterStatuses is null, do we get all statues?

backend/api/Controllers/MissionSchedulingController.cs
```diff
@@ -179,6 +179,23 @@ ISourceService sourceService
...
+            if (returnToHomeMission != null)
+            {
+                if (!await localizationService.RobotIsOnSameDeckAsMission(robot.Id, missionRun.Area.Id))
+                {
+                    return BadRequest($"The robot {robot.Name} is localized on a different deck so the mission was not scheduled.");
```
oysand, line 188: Should we use 409 Conflict instead of BadRequest? It can be argued that the request is valid, since it can work exactly the same with a different state

oysand (changes requested):
See comments

backend/api/Controllers/MissionSchedulingController.cs
andchiind: I personally like the latter.

mrica-equinor: Have adressed this

backend/api/Controllers/MissionSchedulingController.cs
```diff
@@ -350,6 +372,26 @@ ISourceService sourceService
                 await missionDefinitionService.Create(scheduledMissionDefinition);
             }
 
+            IList<MissionStatus> missionStatuses = [MissionStatus.Ongoing, MissionStatus.Pending, MissionStatus.Paused];
+            var missionRunType = MissionRunType.ReturnHome;
```
oysand, line 376: Can this be defined directly below in the function call?

backend/api/Controllers/MissionSchedulingController.cs
```diff
@@ -350,6 +372,26 @@ ISourceService sourceService
...
+            var missionRuns = await missionRunService.ReadMissionRuns(robot.Id, missionRunType, missionStatuses);
+
+            if (missionRuns.Count != 0)
```
oysand, line 379: Make a check for more than one

backend/api.test/EventHandlers/TestMissionEventHandler.cs
```diff
@@ -381,5 +382,71 @@ public async void ReturnHomeMissionNotScheduledIfRobotIsNotLocalized()
...
+            var updatedReturnHomeMission = await _missionRunService.ReadById(returnToHomeMission.Id);
+            Assert.True(updatedReturnHomeMission?.Status.Equals(MissionStatus.Cancelled));
+            Assert.True(updatedReturnHomeMission?.Tasks.FirstOrDefault()?.Status.Equals(TaskStatus.Cancelled));
+
+            // Act
+            // State we get after stopping mission in Isar
+            robot.Status = RobotStatus.Available;
+            await _robotService.Update(robot);
+            var robotAvailableArgs = new RobotAvailableEventArgs(robot.Id);
+            _missionSchedulingService.RaiseEvent(nameof(MissionSchedulingService.RobotAvailable), robotAvailableArgs);
```
oysand, line 416: If we trigger a robot status event from the mqtt service it would set it to available and trigger this. Then we would not need the status update above

backend/api.test/EventHandlers/TestMissionEventHandler.cs
```diff
@@ -381,5 +382,79 @@ public async void ReturnHomeMissionNotScheduledIfRobotIsNotLocalized()
...
+
+            Thread.Sleep(500);
+
+            // Assert 
+            var updatedReturnHomeMission = await _missionRunService.ReadById(returnToHomeMission.Id);
+            Assert.True(updatedReturnHomeMission?.Status.Equals(MissionStatus.Cancelled));
+            Assert.True(updatedReturnHomeMission?.Tasks.FirstOrDefault()?.Status.Equals(TaskStatus.Cancelled));
+
+            // Act
+            var robotAvailableArgs = new MqttReceivedArgs(new IsarRobotStatusMessage
```
oysand, line 412: Remove this part of the test

oysand (approved):
LGTM

PR #1564, opened 2024-04-19
Fix pressure display when value is zero

mrica-equinor:


frontend/src/components/Pages/RobotPage/RobotPage.tsx
```diff
@@ -95,7 +95,7 @@ export const RobotPage = () => {
                                             batteryLevel={selectedRobot.batteryLevel}
                                             batteryWarningLimit={selectedRobot.model.batteryWarningThreshold}
                                         />
-                                        {selectedRobot.pressureLevel && (
+                                        {selectedRobot.pressureLevel != null && (
```
Eddasol, line 98: Should we check for both null and undefined?

mrica-equinor: I've added it, thanks for the catch!

Eddasol (approved):
LGTM :)

PR #1568, opened 2024-04-23
Add return_to_home to capabilities

tsundvoll:


aestene (approved):

tsundvoll:
I have verified that no migration is needed, by running `dotnet ef migrations add AddTableNamePropertyName` and seeing that it is empty

PR #1570, opened 2024-04-25
Correct error message for robotNotAvailable

Eddasol:


backend/api/EventHandlers/MissionEventHandler.cs
```diff
@@ -102,9 +102,17 @@ or RobotNotFoundException
                         or IsarCommunicationException
                     )
                     {
-                        _logger.LogError("Mission run {MissionRunId} will be aborted as robot {RobotId} was not correctly localized", missionRun.Id, missionRun.Robot.Id);
+                        if (ex is RobotNotAvailableException)
```
andchiind, line 105: Instead of an if statement we could have "catch (RobotNotAvailableException ex)" before the current catch, and just exclude if from the list of exceptions in the current "catch" statement.

andchiind: Looks good otherwise

Eddasol: Was contemplating if it would be better to do like this, and avoid duplicating status = aborted and update part, or if I should duplicate those lines and split the exceptions. Can change to your suggestion

andchiind (approved):
LGTM

PR #1571, opened 2024-04-26
Allow tags in several decks in missions

### LINKED ISSUE #1567: Cannot handle tags that go over several decks
**Describe the bug**
For echo missions:
Tags can only be registered for one area and if a tag goes over several decks it can cause issues.

Assume a tag tag1 that is on both deckA and deckB but is registered as being on deckA. If we want to inspect a part that is on deckB there will be issues. If tag1 is one of several tags (assuming normal tags on just one deck) then the mission will fail to be scheduled. If tag1 is the only one then the mission will be added to the inspection plan for deckA. 

**To Reproduce**
Schedule mission with tags that are registered as being on a different deck than the belonging robot/inspection pose

**Expected behavior**
Should have a way to handle tags that spans more than one deck

**Screenshots**

andchiind:
Partly addresses #1567. Issues still remain if most tags are on a different deck, as this deck will be used as the "main" deck.

### DIFF
--- a/backend/api/Controllers/MissionSchedulingController.cs
+++ b/backend/api/Controllers/MissionSchedulingController.cs
@@ -285,7 +285,8 @@ namespace Api.Controllers
             {
                 missionDeck ??= missionArea.Deck;
 
-                if (missionDeck != missionArea.Deck) { return BadRequest("The mission spans multiple decks"); }
+                if (missionDeck != missionArea.Deck)
+                    logger.LogWarning($"Mission {echoMission.Name} has tags in both deck {missionDeck.Name} and {missionArea.Deck.Name}.");
             }
 
             Area? area = null;


backend/api/Controllers/MissionSchedulingController.cs
```diff
@@ -285,7 +285,8 @@ ISourceService sourceService
             {
                 missionDeck ??= missionArea.Deck;
 
-                if (missionDeck != missionArea.Deck) { return BadRequest("The mission spans multiple decks"); }
+                if (missionDeck != missionArea.Deck)
+                    logger.LogWarning($"Mission {echoMission.Name} has tags in both deck {missionDeck.Name} and {missionArea.Deck.Name}.");
```
Eddasol, line 289: Will this ensure that we select the deck with the most tags?

andchiind: No, but I can try implement that.

andchiind: Is it maybe not best to just use the area belonging to the area in the request body? I think it might get very confusing otherwise in the frontend, and it sounds like what you'd expect when using the endpoint. It is anyways what we do now and have done previously

Eddasol: Might be enough to just select area with the most tags. Will in most cases also be the deck with the most tags

Eddasol:
> Closes #1567

Suggest not closing or creating new issue that covers remaining part of the issue. This will solve issue for missions where the majority of tags are marked in right area but not for missions where that isn't the case

backend/api/Controllers/MissionSchedulingController.cs
andchiind: I have updated the issue to explain that I am not addressing this issue in this PR, so I will also resolve this conversation.

PR #1572, opened 2024-04-30
Pressure table on robot page

prasm313:


frontend/src/components/Pages/RobotPage/PressureTable.tsx
```diff
@@ -0,0 +1,76 @@
...
+                        <Table.Head sticky>
+                            <Table.Row>
+                                <Table.Cell>{TranslateText('Pressure level')}</Table.Cell>
+                                <Table.Cell>{TranslateText('Pressure limit')}</Table.Cell>
+                            </Table.Row>
+                        </Table.Head>
+                        <Table.Body>
+                            <StyledTableRow>
+                                <Table.Cell>{TranslateText('HighHigh')}</Table.Cell>
+                                <Table.Cell>{'> 90'}</Table.Cell>
```
mrica-equinor, line 52: You can think about moving these values to variables that are imported from another file/enum. But other than that it looks good

mrica-equinor (approved):
LGTM :) See the comment I made

PR #1602, opened 2024-06-04
Add a common mission pre-check function

prasm313:
The PR also refactors custom mission pre-checks and adds a pre-check for high pressure.

### DIFF
--- a/backend/api/Controllers/MissionSchedulingController.cs
+++ b/backend/api/Controllers/MissionSchedulingController.cs
@@ -26,7 +26,6 @@ namespace Api.Controllers
             ISourceService sourceService
         ) : ControllerBase
 
-
     {
         /// <summary>
         ///     Rerun a mission run, running only the parts that did not previously complete
@@ -48,7 +47,7 @@ namespace Api.Controllers
         )
         {
             Robot robot;
-            try { robot = await missionSchedulingService.MissionPreCheck(scheduledMissionQuery.RobotId); }
+            try { robot = await missionSchedulingService.GetRobotWithPreCheck(scheduledMissionQuery.RobotId); }
             catch (Exception e) when (e is RobotNotFoundException) { return NotFound(e.Message); }
             catch (Exception e) when (e is RobotPreCheckFailedException) { return BadRequest(e.Message); }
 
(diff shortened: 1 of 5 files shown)


backend/api/Services/MissionSchedulingService.cs
```diff
@@ -27,6 +28,9 @@ public interface IMissionSchedulingService
         public void TriggerRobotAvailable(RobotAvailableEventArgs e);
 
         public void TriggerMissionCompleted(MissionCompletedEventArgs e);
+
+        public Task<Robot> MissionPreCheck(string robotId);
```
andchiind, line 31: Could we call it PerformMissionPreCheck or another verb?

backend/api/Services/MissionSchedulingService.cs
```diff
@@ -543,5 +546,40 @@ private static float CalculateDistance(Pose pose1, Pose pose2)
...
+            }
+
+            if (!robot.IsRobotBatteryTooLow())
+            {
+                string errorMessage = $"The robot battery level on {robot.Name} is too low to start a mission";
+                logger.LogError("{Message}", errorMessage);
+                throw new RobotPreCheckFailedException(errorMessage);
+            }
+
+            return robot;
```
andchiind, line 582: Why does the function return the robot?

backend/api/Controllers/MissionSchedulingController.cs
```diff
@@ -6,25 +6,28 @@
...
             IEchoService echoService,
             ILogger<MissionSchedulingController> logger,
             IMapService mapService,
             IStidService stidService,
             ILocalizationService localizationService,
             ISourceService sourceService
         ) : ControllerBase
-    {
 
+
```
andchiind, line 29: Could we remove this extra line? 🙂

backend/api/Services/MissionSchedulingService.cs
andchiind: We might want to rename it to something where it would make sense for it to return the robot (eg. GetRobotWithPreCheck), otherwise we should not return the robot and just get the robot again in the controller function.

prasm313: It returns a robot because the endpoints in MissionSchedulingController.cs use the robot object after the pre-checks

andchiind (approved):
LGTM

PR #1603, opened 2024-06-05
Remove rerun button for localization and return to home missions

prasm313:


andchiind (commented):
I would personally argue that we don't need to show localisation and return to home missions in the historic mission page, but I know others will disagree with me on that. In that case though there is parameters that let us do that 
image

Otherwise though I think instead of checking the mission name here, we should instead be checking the contents of the mission. So for localisation we can check if it is a mission with just Localization type task, and for return to home it only has ReturnHome type
image

PR #1608, opened 2024-06-10
Limit height of dropdown menu for autocomplete

prasm313:


mrica-equinor (approved):
LGTM :) In the future we can try to get the dropdown height set based on screen size

PR #1609, opened 2024-06-11
Add exception handling in event handlers

andchiind:
If we have an unhandled exception in an event handler, then Flotilla crashes. This will prevent this in most cases. I have not added try/catch for all the database "Update" calls, even though these can return database exceptions, since we need a more systematic solution for these.

### DIFF
--- a/backend/api/Database/Models/MissionRun.cs
+++ b/backend/api/Database/Models/MissionRun.cs
@@ -114,7 +114,7 @@ namespace Api.Database.Models
             );
         }
 
-        public static MissionStatus MissionStatusFromString(string status)
+        public static MissionStatus GetMissionStatusFromString(string status)
         {
             return status switch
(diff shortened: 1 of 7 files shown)


backend/api/Database/Models/MissionRun.cs
```diff
@@ -114,7 +114,7 @@ public void UpdateWithIsarInfo(IsarMission isarMission)
             );
         }
 
-        public static MissionStatus MissionStatusFromString(string status)
```
Christdej, line 117: Why rename?

Christdej (approved):
LGTM

backend/api/Database/Models/MissionRun.cs
andchiind: Because function names should be verbs

PR #1613, opened 2024-06-11
Delete localisation missions

### LINKED ISSUE #1595: Remove localization missions
Currently we send an explicit localization mission to the robots that support localization. This is done in order to check that the robot is in the correct location before scheduling a full mission. However, robots will regardless need to localize in order to run the next mission. Therefore, as long as we are able to tell from the feedback we get from ISAR that the reason a mission failed was due to localization failing, then there is no need to have localization as a Flotilla feature. This would significantly simplify Flotilla and make the system more robust. In this case we would not need to check what missions are currently scheduled or running, or what mission just finished. Instead we could simply run a mission when the robot is available, and if the localization step fails then we report this to the user.

As described above, we will keep the prompt for the user in the frontend, telling them to ensure the robot is in the right place when starting a mission, but after that we will simply schedule the intended mission. The user will therefore not need to understand what a localization mission does, and they will only need to care about localization when it fails. Functionality will otherwise not change, but a major cleanup of the mission scheduling logic can then be done.

It will generally not change, but the resulting simplification of the code can improve the overall robustness of the system in the long run.

andchiind:
Closes #1595

### DIFF
--- a/backend/api.test/Database/DatabaseUtilities.cs
+++ b/backend/api.test/Database/DatabaseUtilities.cs
@@ -59,13 +59,6 @@ namespace Api.Test.Database
                 Map = new MapMetadata(),
                 InstallationCode = installationCode
             };
-            if (missionRunType == MissionRunType.Localization)
-            {
-                missionRun.Tasks = new List<MissionTask>
-                {
-                    new(new Pose(), MissionTaskType.Localization)
-                };
-            }
             if (missionRunType == MissionRunType.ReturnHome)
             {
                 missionRun.Tasks = new List<MissionTask>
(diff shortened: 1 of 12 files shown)


mrica-equinor:
Don't we want to keep the stepwise for a while? Should we maybe delay this getting merged a bit longer?

andchiind:
> Don't we want to keep the stepwise for a while? Should we maybe delay this getting merged a bit longer?

Stepwise does regardless not work right now, and when we get stepwise to work I anyways think that we should not be using localisation missions.

andchiind:
Some of the same changes can still be made even after we have now redone how localisation missions work (we no longer have standalone localisation missions), but this should be done in another PR.

PR #1616, opened 2024-06-12
Fix buttons disappearing on long maps

### LINKED ISSUE #1600: Confirm robot location button not reachable for larger images
**Describe the bug**
The card for checking that the robot is placed at the right location is not scrollable and goes past the bottom of the screen for larger images, making it hard to actually click the confirm button.

**To Reproduce**
Start a mission with a large map

**Expected behavior**
We should be able to scroll to the bottom of the card.

**Screenshots**
hard to confirm

Eddasol:
Closes #1600


haakonsf (approved):
LGTM

Issue #1537, opened 2024-04-09
Frontend screen sometimes goes white when scheduling multiple missions at once


oysand:
**Describe the bug**
Sometimes when scheduling all the missions on the mezzanine deck at Huldra in dev the screen is goes completely white. A refresh "fixes" the problem.

**To Reproduce**
Queue all the missions on the mezzanine deck at Huldra multiple times.

**Expected behavior**
Screen should not be blank

**Screenshots**
![image](https://github.com/equinor/flotilla/assets/32652118/cc1d2f94-9a69-4c05-ba37-e93037cebd0f)

oysand:
Experienced the bug with updated mission topic as well with the missionRunContext

mrica-equinor:
I haven't been able to reproduce this bug so we might consider waiting on this to see if it keeps happening

oysand:
New instance seen in staging: 
![image](https://github.com/equinor/flotilla/assets/32652118/c1580683-a302-4e21-9123-b4f53695c584)
Occured after opening staging and just leaving it on the side in a tab for approximatly 5-10 minutes. When revisiting the tab it was all white and had error on the robot status card

oysand:
New occurance: 
![image](https://github.com/equinor/flotilla/assets/32652118/89e1fdf2-0566-4dd7-aa1a-4ed43f6ce802)

mrica-equinor:
![Image](https://github.com/equinor/flotilla/assets/108514046/b5cbec08-fc17-4b4e-a2bf-46e545e20dfb)



![Image](https://github.com/equinor/flotilla/assets/108514046/8f0973d8-43f7-4804-bd87-dd3300119e0a)

Eddasol:
When adding this to signalRContext.tsx
![Image](https://github.com/equinor/flotilla/assets/33518988/3cb26b57-25d7-4777-be92-4e14b9de49d1)

the console.log message appears at the same time as these kind of "screen goes white"-events
![Image](https://github.com/equinor/flotilla/assets/33518988/1a09635d-b719-47fa-bc18-653de446a9e8)

Seems that the issue is caused by signalRMessages sometimes being empty

Eddasol:
Frontend part of this issue is solved by adding error handling in #1617. Backend issue and root cause is handled by #1604

oysand:
Reappered on robot status card
![image](https://github.com/equinor/flotilla/assets/32652118/43990393-f13f-4b18-928a-6de7d0808d58)

Issue #1604, opened 2024-06-05
Isar mission run sometimes is not yet created when receiving task status


andchiind:
**Describe the bug**
In the screenshot, the LogWarning line is sometimes reached. This is likely due to the mission run not being updated in the database at the point in time when the first task update is received from ISAR, causing it to not find the mission run. This causes it to send a value of null to signalR, in turn causing an error in the frontend. This should be handled better, and at the very least we should not send a null value for the mission run to the frontend.

**To Reproduce**
Running missions with isar_robot where the first task finishes quickly.

**Expected behavior**
The mission run value sent to signalR should never be null.

**Screenshots**
![image](https://github.com/equinor/flotilla/assets/20266900/3c60d2bc-3994-4b34-b7c9-58e9a4874cc7)

PR #1617, opened 2024-06-12
Prevent frontend from crashing on empty signalR message

Eddasol:
Solves #1537

### DIFF
--- a/frontend/src/components/Contexts/SignalRContext.tsx
+++ b/frontend/src/components/Contexts/SignalRContext.tsx
@@ -81,7 +81,14 @@ export const SignalRProvider: FC<Props> = ({ children }) => {
     }, [accessToken])
 
     const registerEvent = (eventName: string, onMessageReceived: (username: string, message: string) => void) => {
-        if (connection) connection.on(eventName, (username, message) => onMessageReceived(username, message))
+        if (connection)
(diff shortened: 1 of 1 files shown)


frontend/src/components/Contexts/SignalRContext.tsx
```diff
@@ -81,7 +81,14 @@ export const SignalRProvider: FC<Props> = ({ children }) => {
     }, [accessToken])
 
     const registerEvent = (eventName: string, onMessageReceived: (username: string, message: string) => void) => {
-        if (connection) connection.on(eventName, (username, message) => onMessageReceived(username, message))
+        if (connection)
+            connection.on(eventName, (username, message) => {
+                if (message === 'null') {
```
andchiind, line 86: Is the message a string with just "null" in it, and not a null string?

andchiind:
This PR solves the error checking in the frontend, which is good. The underlying backend issue is likely this: https://github.com/equinor/flotilla/issues/1604

frontend/src/components/Contexts/SignalRContext.tsx
Eddasol: Yes, the message is 'null' and its only after JSON.parse that we end up with null

oysand (approved):
LGTM

PR #1618, opened 2024-06-14
Allow echo missions with tags without area code

### LINKED ISSUE #1561: Entire echo mission fails if one tag is missing area information
**Describe the bug**
When scheduling echo mission: If one of the tags in the mission is lacking locationCode, it will throw an error and the entire mission will fail to be scheduled. This then leads to a "Request error"-banner in frontend with the message: Failed to schedule mission 'mission name'. An unexpected error occurred when handling the request.

**To Reproduce**
Start mission "Mission where tag is missing locationCode"

**Expected behavior**
Should the mission fail of one of the tags have problems? 
  If so: should the problematic tag still be included (thinking yes but unsure)?
  If not: There needs to be better information to the user

**Screenshots**

Eddasol:
Closes #1561

### DIFF
--- a/backend/api.test/Mocks/StidServiceMock.cs
+++ b/backend/api.test/Mocks/StidServiceMock.cs
@@ -12,7 +12,7 @@ namespace Api.Services
     {
         public const string ServiceName = "StidApi";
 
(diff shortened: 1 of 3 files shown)


mrica-equinor (approved):
LGTM :)

PR #1626, opened 2024-06-20
Move missionwise localisation inside semaphore

andchiind:


mrica-equinor (approved):
LGTM :)

backend/api/Services/MissionSchedulingService.cs
```diff
@@ -71,6 +71,7 @@ public async Task StartNextMissionRunIfSystemIsAvailable(string robotId, bool is
                     await robotService.UpdateCurrentArea(robot.Id, null);
                 }
                 if (missionRun == null) { return; }  // The robot is already home
+                isFirstMissionInQueue = false;
```
mrica-equinor, line 74: Treats the symptom but doesn't fix it

andchiind: Still a useful failsafe in my opinion, in case mistakes have been made elsewhere in the code. This will prevent invalid values being sent to ISAR in that case.

PR #1665, opened 2024-07-10
Fix removing a mission from queue that has started

### LINKED ISSUE #1548: Backend crashes if we remove a mission from the queue that has just started
**Describe the bug**
If we click the remove mission from queue button just when the mission is starting, the status will be ongoing instead of pending but the ongoing card and pending mission card haven't re-rendered. As a consequence, the backend crashes. 

This should not happen, the exception should be caught and a message to the frontend should say that the mission wasn't able to be removed because it already started. Possibly ask the user if it wants to stop the mission.

**To Reproduce**
Remove the first mission in the queue just when the localization is finishing.

**Expected behavior**

**Screenshots**

mrica-equinor:
Closes #1548

### DIFF
--- a/backend/api/Services/MissionRunService.cs
+++ b/backend/api/Services/MissionRunService.cs
@@ -282,6 +282,7 @@ namespace Api.Services
                 return null;
             }
 
+            context.Entry(missionRun.Status).State = EntityState.Unchanged;
             context.MissionRuns.Remove(missionRun);
             await ApplyDatabaseUpdate(missionRun.Area?.Installation);
             _ = signalRService.SendMessageAsync("Mission run deleted", missionRun?.Area?.Installation, missionRun != null ? new MissionRunResponse(missionRun) : null);


andchiind (commented):
If this solution works then we can merge it, but I was just wondering if you know why this fixes the issue?

mrica-equinor:
> If this solution works then we can merge it, but I was just wondering if you know why this fixes the issue?

Basically what is happening is that due to timing when this code runs the mission status has been updated from pending to ongoing. What happens currently is that the mission will remain ongoing and the user needs to click on it again to remove it by stopping it. 

Why isn't the mission actually removed? I honestly don't know. I expected the behaviour to be removing the mission completely but it doesn't. Could be that it can't remove something with an entity with status that is unchanged. 

Either way this fixes the bug, and the backend no longer crashes but I agree we could look into it more to fully understand what is happening.

andchiind (approved):
LGTM

PR #1668, opened 2024-07-16
Add missing error handling in frontend

andchiind:
It is worth noting that error messages that use the FailedRequestAlertContent component, do not update their translations when the selected language changes. I am looking into this, but it is not a large priority since this is a very niche issue,


mrica-equinor (approved):
LGTM :)

PR #1671, opened 2024-07-17
Use first mission task status to localize robot

### LINKED ISSUE #1655: Stopping first mission scheduled will set the current area to null
**Describe the bug**
If we click the stop button when the first mission in the queue is running, the current area will be set to null because the localization mission was not successful. As a consequence the returm to home mission or the following missions won't start. 

**To Reproduce**
Schedule a mission and stop it.

**Expected behavior**
The current area should be set when the start (localization) step of the first mission is successful, not when the hole mission is successful. 

It is suggested to check that the status of the first task in the mission becomes ongoing - this means the start step was successful and we can now update the current area.

mrica-equinor:
Closes #1655

### DIFF
--- a/backend/api/EventHandlers/MqttEventHandler.cs
+++ b/backend/api/EventHandlers/MqttEventHandler.cs
@@ -265,11 +265,9 @@ namespace Api.EventHandlers
                 return;
             }
 
-            if (flotillaMissionRun.Status == status) { return; }
-
             if (flotillaMissionRun.IsLocalizationMission())
             {
-                if (status == MissionStatus.Successful || status == MissionStatus.PartiallySuccessful)
+                if (flotillaMissionRun.Tasks.Any((task) => task.Status == Database.Models.TaskStatus.Successful || task.Status == Database.Models.TaskStatus.PartiallySuccessful))
                 {
                     try
                     {
@@ -281,7 +279,7 @@ namespace Api.EventHandlers
                         return;
                     }
                 }
-                else if (status == MissionStatus.Aborted || status == MissionStatus.Cancelled || status == MissionStatus.Failed)
+                else if (flotillaMissionRun.Tasks.All((task) => task.Status == Database.Models.TaskStatus.Cancelled || task.Status == Database.Models.TaskStatus.Failed) || flotillaMissionRun.Status == MissionStatus.Aborted)
                 {
                     try
                     {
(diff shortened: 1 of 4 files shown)


backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -281,7 +280,7 @@ private async void OnIsarMissionUpdate(object? sender, MqttReceivedArgs mqttArgs
                         return;
                     }
                 }
-                else if (status == MissionStatus.Aborted || status == MissionStatus.Cancelled || status == MissionStatus.Failed)
```
andchiind, line 284: Why do we no longer accept Aborted?

andchiind: I understand we might not want to send an error to the user in that case, but we also do want the robot to be unlocalized in that case.

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -265,11 +265,10 @@ private async void OnIsarMissionUpdate(object? sender, MqttReceivedArgs mqttArgs
                 return;
             }
 
-            if (flotillaMissionRun.Status == status) { return; }
-
             if (flotillaMissionRun.IsLocalizationMission())
             {
-                if (status == MissionStatus.Successful || status == MissionStatus.PartiallySuccessful)
+                var task = flotillaMissionRun.Tasks.FirstOrDefault();
+                if (task != null && (task.Status == Database.Models.TaskStatus.Successful || task.Status == Database.Models.TaskStatus.PartiallySuccessful))
```
andchiind, line 271: This probably works fine, but it could be a good idea to just check if any tasks have succeeded, in case the ordering is not consistent.

backend/api/EventHandlers/MqttEventHandler.cs
mrica-equinor: The reason is because there is no TaskStatus of type aborted, we could still check the mission status and if the mission has been aborted decide we want the robot to not be localized

backend/api/EventHandlers/MqttEventHandler.cs
mrica-equinor: I guess the first task could fail for some weird reason and then we move on to the second tag and if that one works then the localization worked, can change

andchiind (approved):
LGTM, can be merged once tests work

PR #1674, opened 2024-07-18
Verify installation for inspections in the inspection table

andchiind:
I noticed that we were only checking that deck names were matching, but not that the inspections belonged to the right installation. This solves that issue. 

A more permanent solution might be to limit all the mission runs available in the MissionRunsContext to only those of the relevant installation, but this makes it potentially more complicated to switch installations.


betaniat (commented):
LGTM :)

mrica-equinor (approved):
LGTM :)

PR #1676, opened 2024-07-18
Change front page background color

mrica-equinor:


andchiind (approved):
LGTM. Are we also planning on changing the background colour on other pages?

mrica-equinor:
> LGTM. Are we also planning on changing the background colour on other pages?

Yep but didn't want to do everything in one go

PR #1677, opened 2024-07-19
Filter robots by available or busy status

### LINKED ISSUE #1556: Only allow selecting an available robot when scheduling a mission
When scheduling a mission, a pop up with a drop down appears to select a robot. Currently, it is possible to select robots that are not available as they are not on the correct deck. I suggest to include all robots on the asset in the drop down menu, but only make the ones available for that particular mission selectable.

Provide more information and context to the robots when selecting them for a mission.

betaniat:
Closes #1556


mrica-equinor (approved):
LGTM :)

PR #1678, opened 2024-07-19
Add detailed signalR logs

### LINKED ISSUE #1563: Set up server side SignalR log
When a message is sent from the server using signalR, the message should be logged.

See: https://learn.microsoft.com/en-us/aspnet/core/signalr/diagnostics?view=aspnetcore-8.0

Should not be affected if we treat it as our other server logs.

andchiind:
Closes #1563

### DIFF
--- a/backend/api/appsettings.json
+++ b/backend/api/appsettings.json
@@ -24,7 +24,8 @@
     "ApplicationInsights": {
       "LogLevel": {
         "Default": "Warning",
-        "Api": "Information"
+        "Microsoft.AspNetCore.SignalR": "Debug",
+        "Microsoft.AspNetCore.Http.Connections": "Debug"
       }
     }
   },


mrica-equinor (approved):
LGTM :)

backend/api/appsettings.json
```diff
@@ -24,7 +24,8 @@
     "ApplicationInsights": {
       "LogLevel": {
         "Default": "Warning",
-        "Api": "Information"
+        "Microsoft.AspNetCore.SignalR": "Debug",
+        "Microsoft.AspNetCore.Http.Connections": "Debug"
```
Christdej, line 28: Do we want Debug to be the general level or is this something we set when we want to debug something?

andchiind: To be honest, the documentation is somewhat vague on what will be logged: https://learn.microsoft.com/en-us/aspnet/core/signalr/diagnostics?view=aspnetcore-8.0#server-side-logging
Here they only mention two options, debug and trace, and they state trace is more detailed than debug. As logn as we can see when messages are being sent then I think the log could be useful, but if this log is not detailed enough then another option would be to just use the normal logger in the SignalRService class, though this is theoretically not as reliable as we cannot guarantee that the messages are being sent properly when we call the function.

Christdej: Lets try with Debug then, and change it if it floods the logs with to much information

Christdej (approved):

PR #1680, opened 2024-07-19
Only schedule return home if supported by robot

### LINKED ISSUE #1672: Do not send return home mission if the robot does not support it
Although ISAR sends the robot capabilities which states whether return to home is supported, this is currently not used to determine whether such a mission should be sent. We should do this the same way we handle localisation missions. In this case we need to be careful to set the robot to delocalised when we would normally schedule a return to home mission.

We will no longer be sending unsupported requests to ISAR.

N/A

andchiind:
Closes #1672


Christdej (approved):
LGTM

PR #1684, opened 2024-07-23
Make robot go home when battery or pressure levels are outside safe limits

### LINKED ISSUE #1459: If robot battery is running low, stop the missions and send it home
We should estimate how much battery we need to run the robot back home before it reaches the baterry limit. For example, if the limit is 10% and from where we are it takes 5% to go back home, when the robot reaches 15% battery it should go home. 

Since we don't have a way of estimating the battery necessary to go back home at the moment, I believe we could use the lower baterry limit until another solution is found. 

If the robot reaches the lower battery limit, the ongoing mission should be stoped, none of the missions in the qeue are executed and the return to home mission starts.

### LINKED ISSUE #1469: Make return-to-home more robust
Return-to-home have several issues:
- [x] If arm is moved the robot will try to schedule a return-to-home mission. The mission will fail and do nothing as the robot is missing current area, but the scheduling attempt shouldn't happen at all. #1466 
- [ ] If robot is turned off without being allowed to complete return-to-home mission it will start a return-to-home when it gets started again. Should there be some timer that de-localizes the robot after a certain time if nothing happens (idle/offline?). Related to #1450 
- [x] #1464 

How can we make return-to-home behavior more robust?

In addition, I am adding this to the epic because I believe it is somehow related and would be good to get it all implemented in one go. 
- [x] #1459 

andchiind:
Closes #1459

### DIFF
--- a/backend/api.test/EventHandlers/TestMissionEventHandler.cs
+++ b/backend/api.test/EventHandlers/TestMissionEventHandler.cs
@@ -468,8 +468,8 @@ namespace Api.Test.EventHandlers
             Thread.Sleep(100);
 
             // Act
-            var eventArgs = new EmergencyButtonPressedForRobotEventArgs(robot.Id);
-            _emergencyActionService.RaiseEvent(nameof(EmergencyActionService.EmergencyButtonPressedForRobot), eventArgs);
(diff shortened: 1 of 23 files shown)


betaniat (commented):
LGTM :) I saw it working

betaniat (approved):

PR #1693, opened 2024-07-25
Store custom sources in SQL instead of blobs

### LINKED ISSUE #1577: Store custom missions in SQL table instead of blob storage
Instead of storing and reading custom missions as JSONs in blob storage, make a new table which contains a string of the missions as a JSON, as well as the hash of the mission. Then we can look up the hash of the current mission we are uploading to check if it already has a record, and otherwise we make a new record.

We would remove any reference to Azure blob storage, making it easier to use for other developers.

It will remove another dependency which might improve the Threat Model.

andchiind:
Closes #1577


betaniat (approved):
LGTM!

PR #1694, opened 2024-07-26
Move installation filtering to contexts

### LINKED ISSUE #1506: Predefined missions tab sometimes show missions from other installations
**Describe the bug**
If a mission definition is created it will show in predefined missions tab, regardless of which installation the user has selected. 

**To Reproduce**
1. Run new missions on a plant such that mission definitions are created
2. Change installation and select predefined missions tab. The missions definitions will appear once they are created. Reload and they will disappear again.

Alternatively:
1. Have two open browsers/tabs with flotilla open at different installations. 
2. Run new missions on one of the plants such that a mission definition is created.
3. Monitor mission predefined missions tab in the other browser/tab and note the new mission definitions.

This is most likely caused by there being no installation filter used for deciding if a new missionDefinition from signalR should be included in the context.

**Expected behavior**
Mission definitions should only be shown for the relevant installation

**Screenshots**

andchiind:
Closes #1506

### DIFF
--- a/frontend/src/components/Contexts/AlertContext.tsx
+++ b/frontend/src/components/Contexts/AlertContext.tsx
@@ -169,14 +169,7 @@ export const AlertProvider: FC<Props> = ({ children }) => {
                 const backendAlert: Alert = JSON.parse(message)
                 const alertType = alertTypeEnumMap[backendAlert.alertCode]
 
-                if (backendAlert.robotId !== null) {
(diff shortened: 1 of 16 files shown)


betaniat (commented):
LGTM, saw it working :)

betaniat (approved):
LGTM, saw it working :)

PR #1697, opened 2024-07-29
Fix last successful mission run update

### LINKED ISSUE #1555: Timestamp for last completed mission is also updated when the mission is canceled
In the mission menu, there is a timestamp for when the mission was last completed. This timestamp is updated every time an execution of that mission is completed. This happens also if the mission was cancelled and actually never or just partly executed.

On the history page, it makes sense to have this timestamp named "last compelted", as it also provide the context of that particular mission execution.

In the mission menu, it does not provide this context, and it refers to "last execution of that mission". Could it be an improvement to show the timestamp of the last _successfully_ completed mission, and not the last completed mission in the mission menu?

No change, just more accurate information

haakonsf:
Closes #1555


Christdej (approved):
lgtm

Issue #1696, opened 2024-07-29
Content on robot card overflows


Eddasol:
**Describe the improvement you would like to see**
When language is Norwegian and  status is connection issues, the card content overflow
<img width="357" alt="image" src="https://github.com/user-attachments/assets/c3a012bc-7e57-4f93-bbc1-37e52edb8ce6">

**How will this change existing functionality?**
It will look nicer

**How will this improvement affect the current Threat Model?**
NA

Eddasol:
Might be useful to solve this one and #1695 as one

PR #1698, opened 2024-07-30
Fix overflow and line break style on robot button cards

### LINKED ISSUE #1695: Status symbol on robot card doesn't center
When a robot has status as connection issues the status symbol doesn't center (height wise) with the status text 
image

It will look better

NA

haakonsf:
Closes #1695 and #1696


Eddasol:
Commit message is a bit long. Suggest something like "Fix overflow and line break on robot status cards"

Eddasol (approved):
Would rewrite commit message before merging but otherwise good :)

PR #1700, opened 2024-07-30
Avoid double-tracking robot entry

andchiind:
There was an issue where rerunning missions didn't work since an entity framework robot instance got tracked twice in the same thread. This seems to fix that. We now don't track objects we aren't planning on updating within the rerun function. We will want to do this in more places in the future.


Eddasol (approved):
LGTM :)

PR #1701, opened 2024-07-30
Prevent localisation mission in safe zone

### LINKED ISSUE #1683: Bug with the safe zone button
**Describe the bug**
It seems that if we click the send robots to safe zone button when the robos is not running a mission/localized and we schedule a mission afterwards, the mission starts. 

Think this was introduced because now the first mission in the queue as a localization mission so it will complete. Before it was assumed localization missions didn't have to be cancelled since they used to not have any tasks in them.

**To Reproduce**
Click the send robots to safe zone and then shedule a mission.

**Expected behavior**
The mission should either be queued or not added to the queue at all, but never moved to ongoing.

**Screenshots**

Image

andchiind:
Closes #1683


haakonsf (approved):
LGTM

PR #1708, opened 2024-08-06
All external mission loading put inside mission loader

Christdej:


frontend/src/api/ApiCaller.tsx
```diff
@@ -43,6 +41,7 @@ export class BackendAPICaller {
             method,
             headers,
             mode: 'cors',
+            // credentials: 'include', // TODO CHECK IF CORRECT
```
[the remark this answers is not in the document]
Christdej: Done =)

backend/api/Configurations/CustomServiceConfigurations.cs
```diff
@@ -124,5 +125,33 @@ IConfiguration configuration
...
             return services;
         }
+
+        public static IServiceCollection ConfigureMissionLoader(
+            this IServiceCollection services,
+            IConfiguration configuration
+        )
+        {
+            string? missionLoaderFileName = configuration["MissionLoader:FileName"];
+            Console.WriteLine(missionLoaderFileName);
```
andchiind, line 135: This should either be removed or logged with a proper logger

backend/api/Services/MissionLoaders/EchoMissionResponse.cs
```diff
@@ -81,7 +81,7 @@ public class SensorType
     public class EchoPose
     {
         [JsonPropertyName("poseId")]
-        public int PoseId { get; set; }
+        public int? PoseId { get; set; }
```
andchiind, line 84: Why is this nullable?

backend/api/Services/MissionLoaders/MissionLoaderInterface.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    public interface IMissionLoader
+    {
+        public Task<MissionDefinition?> GetMissionById(string sourceMissionId);
+
+        public Task<IQueryable<MissionDefinition>> GetAvailableMissions(string? installationCode);
+
+        public Task<List<MissionTask>> GetTasksForMission(string sourceMissionId);
+
+        public Task<List<PlantInfo>> GetPlantInfos(); // Facility service
```
andchiind, line 13: I am not sure if this comment is very helpful right now

frontend/best_practises.md
```diff
@@ -1,50 +1,51 @@
 # Flotilla frontend development best practises
 
-- [Flotilla frontend development best practises](#flotilla-frontend-development-best-practises)
```
andchiind, line 3: A lot of these changes to READMEs and whatnot are very nice, but would it not be good to seperate them into different commits?

frontend/best_practises.md
```diff
@@ -80,8 +82,9 @@ export const MissionTitleComponent = ({ missionName } : MissionTitleComponentPro
 ```
 
 Bad:
+
 ```
-export const RobotComponent = ({ mission, setColor } : {mission: CondensedMissionDefinition, setColor: (c: string) => void}) => {
```
andchiind, line 84: Though I understand that this is needed to be in the same commit as the other frontend changes

frontend/best_practises.md
andchiind: Only a suggestion, not a hard requirement. Just to generally split up the big commit somewhat

backend/api.test/Database/Models.cs
```diff
@@ -71,62 +71,5 @@ public void TestNegativaRotation()
                 3.0
             );
         }
-
-        [Fact]
-        public void AssertCoordinateConversion()
```
andchiind, line 76: Have these tests been moved somewhere or just deleted? Sorry it was hard to tell

andchiind (commented):
Looks good otherwise. Seems to work well when testing it

backend/api/Services/MissionLoaders/EchoMissionResponse.cs
Christdej: I saw it was null one time from Echo, and we have a check:
`if (planItem.PoseId is null)` after so that it throws a meaningfull error message instead of crashing when trying to load the json from Echo

backend/api/Services/MissionLoaders/MissionLoaderInterface.cs
Christdej: Probably not, ill remove it =)

backend/api.test/Database/Models.cs
Christdej: This tests only tests function that are in the test folder and not used in the rest of the application, this should probably have been a separate commit indeed explaining why they are deleted

backend/api/Configurations/CustomServiceConfigurations.cs
Christdej: Removed

andchiind (approved):

PR #1718, opened 2024-08-12
Fix dialogs during ongoing localization mission

### LINKED ISSUE #1463: Add missions to the queue before localization mission is finished
**Describe the bug**
When scheduling additional missions to the queue while there is an ongoing localization mission, the mission is not added. This is due to the fact that for the mission to be added to the queue, it has to be in the same deck as the robot is localized and since the localization mission hans't finished, the robot doesn't have a deck yet.

Suggested solution is to have a check when adding a new mission to the queue for ongoing localization missions and if the mission added is in the same deck as the ongoing localization mission, then it should be added in the queue.

**To Reproduce**
Add mission to the queue before the localization mission updated the robot current area.

**Expected behavior**
As long as the mission to be added is in the same deck as the ongoing localization mission or in the same deck as the robot current area, it should be added to the queue.

**Screenshots**

mrica-equinor:
Closes #1463


Eddasol (approved):
LGTM :)

PR #1733, opened 2024-09-03
Limit logging when fetching user info

andchiind:
The removed exception and its logWarning currently runs whenever a mission is created in an event handler, so it was bloating the log somewhat.


Christdej (approved):
LGTM

PR #1734, opened 2024-09-03
Various small backend refactors

andchiind:


backend/api/Database/Models/Robot.cs
```diff
@@ -20,7 +20,7 @@ public Robot()
             Pose = new Pose();
         }
 
-        public Robot(CreateRobotQuery createQuery, Installation installation, Area? area = null)
+        public Robot(CreateRobotQuery createQuery, Installation installation, RobotModel model, Area? area = null)
```
Christdej, line 23: Could consider putting this as part of CreateRobotQuery

Christdej (approved):
See comment, otherwise approved

backend/api/Database/Models/Robot.cs
andchiind: That's a bit difficult since it needs to be passed in the Swagger endpoint in that case

PR #1745, opened 2024-09-20
Stream media using LiveKit

sondreo:


backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -485,5 +487,25 @@ private async void OnIsarCloudHealthUpdate(object? sender, MqttReceivedArgs mqtt
...
         }
+
+        private async void OnIsarTelemetryUpdate(object? sender, MqttReceivedArgs mqttArgs)
+        {
+            var isarTelemetyUpdate = (IsarTelemetyUpdateMessage)mqttArgs.Message;
+
+            var robot = await RobotService.ReadByIsarId(isarTelemetyUpdate.IsarId);
+            if (robot == null)
+            {
+                _logger.LogInformation("Received message from unknown ISAR instance {Id} with robot name {Name}", isarTelemetyUpdate.IsarId, isarTelemetyUpdate.RobotName);
```
oysand, line 498: Consider logging this as an error

oysand: Follows general pattern

backend/api/MQTT/MessageModels/IsarTelemetryUpdate.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System.Text.Json.Serialization;
+using Api.Services.Models;
+
+namespace Api.Mqtt.MessageModels
+{
+#nullable disable
+    public class IsarTelemetyUpdateMessage : MqttMessage
```
oysand, line 7: Rename to something making it clear that this is the isar message updating the media config

backend/api/Services/Models/IsarMediaConfig.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Text.Json.Serialization;
+namespace Api.Services.Models
+{
+    public struct IsarMediaConfig
```
oysand, line 4: If this object is to pass between the frontend and backend over signalr. It should not be named with the Isar prefix

frontend/src/App.tsx
```diff
@@ -30,32 +31,34 @@ const App = () => (
...
-                                                </AuthenticatedTemplate>
-                                            </MissionControlProvider>
-                                        </MissionRunsProvider>
-                                    </SafeZoneProvider>
-                                </AlertProvider>
-                            </MissionRunsProvider>
-                        </RobotProvider>
-                    </MissionDefinitionsProvider>
-                </InstallationProvider>
+                <MediaStreamProvider>
```
oysand, line 34: Consider moving inside the the installation provider or however far inside makes sense

frontend/src/models/VideoStream.ts
```diff
@@ -1,8 +1,17 @@
...
+export interface MediaStreamConfig {
     url: string
-    type: string
-    shouldRotate270Clockwise: boolean
+    token: string
+    robotId: string
+    mediaConnectionType: MediaConnectionType
 }
+
+export type MediaStreamConfigAndTracks = MediaStreamConfig & { streams: MediaStreamTrack[] }
```
oysand, line 17: Consider splitting this into two objects to avoid needing to naming it as Thing1AndThing2

oysand (changes requested):
See comments

oysand:
Only able to access one video stream at the time. Should be able to open multiple connection in parallel

oysand:
Full screen view does not resize on window change

frontend/src/models/VideoStream.ts
andchiind: I am not quite sure I understand what you mean. Is it the name you disagree with or the expression on the right?

backend/api/EventHandlers/MqttEventHandler.cs
sondreo: Will keep as is for now.

backend/api/Services/Models/IsarMediaConfig.cs
sondreo: Good point. Renamed.

frontend/src/App.tsx
sondreo: Have moved the provider further down the chain.

frontend/src/models/VideoStream.ts
oysand: I think the name is good as it describes well the expression on the right. Though when the name includes *And* i suspect that it might be better suited as two distinct entities that is better combined when they are used. Though I am not familiar enough with webrtc to know if it makes good sense to combine it or if it is better to combine where needed

sondreo:
> Only able to access one video stream at the time. Should be able to open multiple connection in parallel

Will defer this to a [separate issue](#1754) for now.

sondreo:
> Full screen view does not resize on window change

Good catch! Have created a separate improvement [issue](#1754).

frontend/src/models/VideoStream.ts
sondreo: Have simplified this type.

oysand (approved):
LGTM

Issue #1528, opened 2024-04-07
Consider rewording of "safe zone"


anetteu:
**Describe the improvement you would like to see**
Consider renaming the button `Send robot to safe zone` with something like `Park all robots (outside escape route)` or similar. This to avoid any confusion about what a `safe zone` is. It this context, a `safe zone` can still be an ATEX sone, but it will be out of the way for escape routes and main walkways on the map. 

**How will this change existing functionality?**
No change in functionality, just rewording.

**How will this improvement affect the current Threat Model?**
Leave blank for maintainers to fill out if you are uncertain about this.

prasm313:
I agree with your arguments and really like "Park all robots". As UX-designer has mentioned earlier, a button should not have more than three words. Tried to think of other wordings: "Move out of main path", "Clear walkway", "Relocate robots", but they are not as descriptive.

prasm313:
Have talked with UX-designer. They want to have input from users before changing the wording on the button.

anetteu:
What about something like "Send robots home"? The current wording is five words, how important is it to have only three? Do we need to wait for the optimal wording and user input, or can we change it and then change it again after we have found the optimal wording?

Eddasol:
Already rewritten from "safe zone" to "dock"

PR #1759, opened 2024-10-11
Improve safezone stability

andchiind:
This PR removes a lot of failure points and instabilities in the safezone (and therefore also the charging) functionality. In a future PR I also want to implement this: https://github.com/equinor/flotilla/issues/1775 to further improve consistency. In this PR we now use localization poses if no safezone poses are found. In a future PR I would also like to handle this issue: https://github.com/equinor/flotilla/issues/1528 by removing safezone poses and only using localization poses.

### DIFF
--- a/backend/api/Services/MissionSchedulingService.cs
+++ b/backend/api/Services/MissionSchedulingService.cs
@@ -151,11 +151,11 @@ namespace Api.Services
             {
                 logger.LogError(
                     ex,
-                    "Mission run {MissionRunId} was not started successfully due to {ErrorMessage}",
+                    "Mission run {MissionRunId} was not started successfully. {ErrorMessage}",
                     missionRun.Id,
                     ex.Message
                 );
-                await missionRunService.SetMissionRunToFailed(missionRun.Id, $"Mission run '{missionRun.Id}' was not started successfully due to '{ex.Message}'");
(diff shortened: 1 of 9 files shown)


backend/api/Services/MissionSchedulingService.cs
```diff
@@ -376,6 +388,14 @@ private async Task MoveInterruptedMissionsToQueue(IEnumerable<string> interrupte
...
+                    .Where(t => !new List<Database.Models.TaskStatus>
+                        {Database.Models.TaskStatus.Successful, Database.Models.TaskStatus.Failed}
+                        .Contains(t.Status))
+                    .Select(t => new MissionTask(t)).ToList();
+
+                if (unfinishedTasks.Count == 0) continue;
+
                 var newMissionRun = new MissionRun
                 {
                     Name = missionRun.Name,
```
mrica-equinor, line 401: I would change the name to make sure it shows that this mission only has the unfinished tasks, might not be obvious

Name = missionRun.Name + " unfinished tasks"

mrica-equinor (approved):
LGTM :)

backend/api/Services/MissionSchedulingService.cs
andchiind: Sounds good. I might need to have the default be in Norwegian, but I'm not sure. I suppose it depends on what we make the standard for the mission names. I can ask around.

PR #1773, opened 2024-10-22
Improve frontend

mrica-equinor:


frontend/src/components/Pages/MissionPage/MissionHeader/MissionHeader.tsx
```diff
@@ -38,7 +37,7 @@ const StyledCard = styled(Card)`
     flex: 1 0 0;
     padding: 8px 16px;
     flex-direction: row;
-    background: ${tokens.colors.ui.background__light.hex};
+    background: ${tokens.colors.ui.background__default.hex};
```
Eddasol, line 40: When changing the background I would also change the background color of this button:
image

So it matches this:
image

andchiind:
There is now a mismatch in styles for running missions:
image
image
One is on the history page, the other on the inspections page.

frontend/src/components/Pages/MissionPage/MapPosition/MissionMapView.tsx
```diff
@@ -78,6 +79,9 @@ export const MissionMapView = ({ mission }: MissionProps) => {
         [mission.tasks]
     )
 
+    let displayedMapName = mission.map?.mapName.split('.')[0].replace(/[^0-9a-z-A-Z ]/g, ' ')
```
andchiind, line 82: What does this Regex do?

andchiind: Likewise for the line below. If the issue is in the map naming, might it be better to rename it at the source?

frontend/src/utils/DeckMapView.tsx
```diff
@@ -63,6 +73,9 @@ export const DeckMapView = ({ deck, markedRobotPosition }: DeckProps) => {
         return image
     }
 
+    let mapName = mapMetadata?.mapName.split('.')[0].replace(/[^0-9a-z-A-Z ]/g, ' ')
+    mapName = mapName ? mapName.charAt(0).toUpperCase() + mapName.slice(1) : ' '
```
andchiind, line 77: If we need to do this operation each time we get this data, might it be better to do it in getDeckMapMetadata? (Although I would personally prefer not doing this in the frontend, since it makes assumptions about the data sent from the backend/database)

frontend/src/language/en.json
```diff
@@ -265,5 +265,6 @@
     "No alerts": "No alerts",
     "minutes ago": "minutes ago",
     "Failed to retrieve previous mission runs": "Failed to retrieve previous mission runs",
-    "Documentation": "Documentation"
+    "Documentation": "Documentation",
+    "Map of": "Map of"
```
andchiind, line 269:
```suggestion
    "Map of {0}": "Map of {0}"
```

frontend/src/language/no.json
```diff
@@ -265,5 +265,6 @@
     "No alerts": "Ingen varsler",
     "minutes ago": "minutter siden",
     "Failed to retrieve previous mission runs": "Kunne ikke hente tildigere oppdragskjøringer",
-    "Documentation": "Dokumentasjon"
+    "Documentation": "Dokumentasjon",
+    "Map of": "Kart over"
```
andchiind, line 269:
```suggestion
    "Map of {0}": "Kart over {0}"
```

frontend/src/utils/DeckMapView.tsx
```diff
@@ -123,10 +136,15 @@ export const DeckMapView = ({ deck, markedRobotPosition }: DeckProps) => {
...
-                        <StyledMap id="deckMapCanvas" />
-                        {mapMetadata && <MapCompass />}
-                    </StyledMapCompass>
+                    <StyledCaption>
+                        <StyledMapCompass>
+                            <StyledMap id="deckMapCanvas" />
+                            {mapMetadata && <MapCompass />}
+                        </StyledMapCompass>
+                        <Typography italic variant="body_short">
+                            {TranslateText('Map of') + ' ' + mapName}
```
andchiind, line 145:
```suggestion
                            TranslateText('Map of {0}', [mapName])
```

frontend/src/components/Pages/MissionPage/MapPosition/MissionMapView.tsx
mrica-equinor: First line slpits the sentence after a "." and removes everything after and only keeps characters from a to z and 0 to 9. Second one capitalizes the first letter of the map name. So for example a map with name ac_002.png would be renamed to Ac 002.

andchiind (approved):
LGTM

PR #1781, opened 2024-10-30
Fix flotilla not reporting connection issues after broker goes down

### LINKED ISSUE #1359: Robot has status offline if the broker restarts
If the broker goes down and is restarted, the robots has "offline" as their status. If the robot is available or busy when the broker restarts this should be visible in Flotilla.

Robots will not be stuck in offline. Will not be necessary to manually set status as Available through swagger. The users would not need to contact us if the broker has restarted.  

N/A

andchiind:
Closes #1359 

Previously we were sending signalR messages that the robot had been updated with isarConnected being changed, but since the robotModel was null we were ignoring it in the frontend. This was occuring since we were not including the robot model in the database query we used when updating isarConnected. When I changed this to include the robot model we got an error since the ExecuteUpdateAsync function does not allow SQL JOINS (see https://learn.microsoft.com/en-us/ef/core/saving/execute-insert-update-delete#navigations-and-related-entities). I therefore normalised the code to use the context.Update function like we do in the rest of the RobotService. The only place that still uses ExecuteUpdateAsync is the UpdateRobotPose function as this one does not suffer from the same issues due to how the function is used (the return value is ignored), but it's worth keeping in mind.

### DIFF
--- a/backend/api/Services/RobotService.cs
+++ b/backend/api/Services/RobotService.cs
@@ -1,5 +1,6 @@
 ﻿using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
+using System.Linq.Dynamic.Core;
 using Api.Controllers.Models;
 using Api.Database.Context;
 using Api.Database.Models;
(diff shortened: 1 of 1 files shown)


mrica-equinor (approved):
LGTM :)

PR #1786, opened 2024-11-12
Remove unused functions

Eddasol:


frontend/src/components/Contexts/DockContext.tsx
mrica-equinor, line 1: This file is responsible for the context that was previously called safe zone contaxt, we've renamed it since it's more of a return to dock other than safe zone. This file is needed and should be used. Without this the safe zone functionality will be broken

frontend/src/components/Pages/FrontPage/MissionOverview/DockBanner.tsx
mrica-equinor, line 1: This file is responsible for the banner that was previously called safe zone banner, we've renamed it since it's more of a return to dock other than safe zone. This file is needed and should be used. Without this the safe zone functionality will be broken

PR #1826, opened 2024-11-28
Preselect robot if there is only one robot

Eddasol:


Eddasol:
Note: This adds default robot for inspection plan but not when fetching echo missions to schedule. Solving for echo missions is more demanding and will not be prioritized as its not used that much

PR #1829, opened 2024-11-29
Add ida user to mqtt broker

tsundvoll:
to read inspection results


olaals (approved):
Not sure how the password is encrypted but I trust that this is good

tsundvoll:
> Not sure how the password is encrypted but I trust that this is good

Valid concern. The passwords are generated and stored based on this: https://mosquitto.org/man/mosquitto_passwd-1.html

PR #1853, opened 2024-12-06
Add checklist to PRs

Eddasol:


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,10 @@
+## Ready for review checklist:
+- [ ] A self-review has been performed
+- [ ] All commits run individually
+- [ ] All TODO, console.log or other temporary changes have been removed
```
Christdej, line 4:
```suggestion
- [ ] Temporary changes have been removed, like console.log, etc.
- [ ] Unnecessary comments removed, like TODO, commented code, etc.
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,10 @@
...
+- [ ] The PR has been tested locally
+- [ ] A test have been written
+  - [ ] This change doesn't need a new test
+- [ ] Relevant issues are linked
+- [ ] Remaining work is documented in issues
+  - [ ] There is no remaining work from this PR that require new issues
```
Christdej, line 10: Should we add
`- [ ] All pipelines run successfully`
Or perhaps we cannot check this off when creating PR? (depends on if this list is used on PR create or PR ready for review)

Christdej (changes requested):

.github/PULL_REQUEST_TEMPLATE.md
anetteu, line 10: I suggest to add something similar to 

```
- [  ] The changes does not introduce dead code as unused imports, functions etc. 

```

Eddasol: No we need this when we also have the pipelines further down that gives the same check. Should be included in private repos though

Eddasol: image
The checklist will appear in the marked area and cannot be checked until after the pr is created

Christdej: Then I agreed we dont add this =)

Christdej (approved):

PR #1855, opened 2024-12-06
Let user configure custom zoom for tasks

andchiind:
This adds an endpoint where we can map an echo tag ID to a zoom value. This is not meant as a permanent solution, but it can be used until an Echo solution is created.

### DIFF
--- a/backend/api/Database/Context/FlotillaDbContext.cs
+++ b/backend/api/Database/Context/FlotillaDbContext.cs
@@ -1,4 +1,5 @@
 ﻿using Api.Database.Models;
+using Api.Services.MissionLoaders;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.ChangeTracking;
 using Microsoft.EntityFrameworkCore.Metadata.Builders;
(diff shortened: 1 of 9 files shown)


backend/api/Database/Models/EchoTagInspectionMetadata.cs
```diff
@@ -0,0 +1,14 @@
+﻿#nullable disable
+using System.ComponentModel.DataAnnotations;
+using Api.Services.Models;
+
+namespace Api.Services.MissionLoaders
+{
+    public class EchoTagInspectionMetadata
```
Christdej, line 7:
```suggestion
    public class TagInspectionMetadata
```

backend/api/Database/Context/FlotillaDbContext.cs
```diff
@@ -26,6 +27,7 @@ public FlotillaDbContext(DbContextOptions options) : base(options)
         public DbSet<DefaultLocalizationPose> DefaultLocalizationPoses => Set<DefaultLocalizationPose>();
         public DbSet<AccessRole> AccessRoles => Set<AccessRole>();
         public DbSet<UserInfo> UserInfos => Set<UserInfo>();
+        public DbSet<EchoTagInspectionMetadata> EchoTagInspectionMetadata => Set<EchoTagInspectionMetadata>();
```
Christdej, line 30: Avoid using Echo naming

backend/api/Controllers/EchoController.cs
Christdej, line 1: Change to be a "Inspection" or "InspectionMetadata" controller, dont want to lock into Echo naming

Christdej (changes requested):

backend/api/Database/Context/FlotillaDbContext.cs
andchiind: This object is linked to the EchoTag model which already exists in the codebase

Christdej: Not in the database, only in the service and controller models

Christdej (approved):
LGTM

PR #1889, opened 2024-12-16
Improve mission starting stability

andchiind:

### DIFF
--- a/backend/api/EventHandlers/MqttEventHandler.cs
+++ b/backend/api/EventHandlers/MqttEventHandler.cs
@@ -206,7 +206,7 @@ namespace Api.EventHandlers
                     if (isarRobotInfo.Capabilities is not null) UpdateRobotCapabilitiesIfChanged(isarRobotInfo.Capabilities, ref robot, ref updatedFields);
                     if (updatedFields.Count < 1) return;
 
-                    robot = await RobotService.Update(robot);
+                    await RobotService.Update(robot);
                     _logger.LogInformation("Updated robot '{Id}' ('{RobotName}') in database: {Updates}", robot.Id, robot.Name, updatedFields);
                 }
                 finally
@@ -354,7 +354,7 @@ namespace Api.EventHandlers
             var robot = await BatteryTimeseriesService.AddBatteryEntry(batteryStatus.BatteryLevel, batteryStatus.IsarId);
             if (robot != null && robot.BatteryState != batteryStatus.BatteryState)
             {
-                robot = await RobotService.UpdateRobotBatteryState(robot.Id, batteryStatus.BatteryState);
+                await RobotService.UpdateRobotBatteryState(robot.Id, batteryStatus.BatteryState);
             }
 
             _updateRobotSemaphore.Release();
--- a/backend/api/EventHandlers/MqttEventHandler.cs
+++ b/backend/api/EventHandlers/MqttEventHandler.cs
@@ -100,15 +100,16 @@ namespace Api.EventHandlers
             _updateRobotSemaphore.WaitOne();
             _logger.LogDebug("Semaphore acquired for updating robot status");
 
-            var updatedRobot = await RobotService.UpdateRobotStatus(robot.Id, isarStatus.Status);
+            await RobotService.UpdateRobotStatus(robot.Id, isarStatus.Status);
+            robot.Status = isarStatus.Status;
 
             _updateRobotSemaphore.Release();
             _logger.LogDebug("Semaphore released after updating robot status");
 
-            _logger.LogInformation("Updated status for robot {Name} to {Status}", updatedRobot.Name, updatedRobot.Status);
+            _logger.LogInformation("Updated status for robot {Name} to {Status}", robot.Name, robot.Status);
 
 
-            _logger.LogInformation("OnIsarStatus: Robot {robotName} has status {robotStatus} and current inspection area {areaName}", updatedRobot.Name, updatedRobot.Status, updatedRobot.CurrentInspectionArea?.Name);
+            _logger.LogInformation("OnIsarStatus: Robot {robotName} has status {robotStatus} and current inspection area {areaName}", robot.Name, robot.Status, robot.CurrentInspectionArea?.Name);
 
             if (isarStatus.Status == RobotStatus.Available)
             {
--- a/backend/api/Services/MissionSchedulingService.cs
+++ b/backend/api/Services/MissionSchedulingService.cs
@@ -45,7 +45,7 @@ namespace Api.Services
                 return;
             }
 
-            if (robot.MissionQueueFrozen && missionRun != null && !missionRun.IsEmergencyMission())
+            if (robot.MissionQueueFrozen && missionRun != null && !(missionRun.IsEmergencyMission() || missionRun.IsReturnHomeMission()))
             {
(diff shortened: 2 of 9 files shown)


backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -100,15 +100,16 @@ private async void OnIsarStatus(object? sender, MqttReceivedArgs mqttArgs)
             _updateRobotSemaphore.WaitOne();
             _logger.LogDebug("Semaphore acquired for updating robot status");
 
-            var updatedRobot = await RobotService.UpdateRobotStatus(robot.Id, isarStatus.Status);
+            await RobotService.UpdateRobotStatus(robot.Id, isarStatus.Status);
+            robot.Status = isarStatus.Status;
```
andchiind, line 104: This is the main fix in the PR

backend/api/Services/MissionSchedulingService.cs
```diff
@@ -114,6 +114,9 @@ or MissionRunNotFoundException
                 );
                 await missionRunService.SetMissionRunToFailed(missionRun.Id, $"Mission run '{missionRun.Id}' was not started successfully. '{ex.Message}'");
             }
+            catch (RobotBusyException)
```
andchiind, line 117: In this case the mission remains on the queue, as the robot has become busy in another thread

backend/api/Services/MissionSchedulingService.cs
```diff
@@ -452,32 +455,8 @@ private async Task StartMissionRun(MissionRun queuedMissionRun)
             return ongoingMissions;
         }
 
-        private async Task<bool> TheSystemIsAvailableToRunAMission(string robotId, string missionRunId)
+        private bool TheSystemIsAvailableToRunAMission(Robot robot, MissionRun missionRun)
         {
-            bool ongoingMission = await OngoingMission(robotId);
```
andchiind, line 457: Instead of checking the database again here (which can cause concurrency issues which are hard to keep track of), we instead rely on always being able to handle failing to schedule a mission due to the robot being busy.

backend/api/EventHandlers/MqttEventHandler.cs
andchiind: Basically, previously we were relying on the database callback to be up to date, but this tends not to be the case when calling database update. By setting the value here we make the thread execution deterministic.

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -118,6 +119,7 @@ private async void OnIsarStatus(object? sender, MqttReceivedArgs mqttArgs)
                     _logger.LogDebug("Semaphore acquired for updating robot current mission id");
 
                     await RobotService.UpdateCurrentMissionId(robot.Id, null);
+                    await MissionScheduling.StartNextMissionRunIfSystemIsAvailable(robot);
```
oysand, line 122: I think this should happen outside the semaphore

backend/api/EventHandlers/MqttEventHandler.cs
oysand: Was the updated robot not returned with the correct value?

backend/api/Services/MissionSchedulingService.cs
```diff
@@ -68,7 +68,7 @@ public async Task StartNextMissionRunIfSystemIsAvailable(Robot robot)
 
             if (missionRun == null) { return; }
 
-            if (!await TheSystemIsAvailableToRunAMission(robot.Id, missionRun.Id))
+            if (!TheSystemIsAvailableToRunAMission(robot, missionRun))
```
oysand, line 71: I think we should only call it with the id and read out the object again

backend/api/Services/MissionSchedulingService.cs
oysand, line 457: I think we should still use robotId, but if we do not care if there is an ongoing mission (that is waiting for the updates) we should not check it here

backend/api/EventHandlers/MqttEventHandler.cs
oysand, line 122: This should trigger the robot available event and not start directly from the mqtt event handler

oysand (changes requested):
I think we should still only pass the robot id. I suspect that the OnIsarRobotInfo might be the culprint that is setting the robot object to busy

backend/api/Services/MissionSchedulingService.cs
```diff
@@ -388,7 +391,7 @@ private async Task StartMissionRun(MissionRun queuedMissionRun)
             {
                 string errorMessage = $"Robot {robotId} has status {robot.Status} and is not available";
                 logger.LogError("{Message}", errorMessage);
-                throw new RobotNotAvailableException(errorMessage);
+                throw new RobotBusyException(errorMessage);
```
oysand, line 394: Remove reading in of the robot object above

backend/api/Services/MissionSchedulingService.cs
```diff
@@ -452,32 +455,8 @@ private async Task StartMissionRun(MissionRun queuedMissionRun)
...
-
-            var missionRun = await missionRunService.ReadById(missionRunId, readOnly: true);
-            if (missionRun is null)
-            {
-                string errorMessage = $"Mission run with Id {missionRunId} was not found in the database";
-                logger.LogError("{Message}", errorMessage);
-                throw new MissionRunNotFoundException(errorMessage);
-            }
-
             if (robot.MissionQueueFrozen && missionRun.MissionRunType != MissionRunType.Emergency)
```
oysand, line 460: Change to mission run type return home

backend/api/EventHandlers/MqttEventHandler.cs
oysand: Assign it to a seperate value to be passed to a mission starting event along with other prerequisites

oysand (changes requested):
See remaining comments

backend/api/Services/MissionSchedulingService.cs
andchiind: Resolved in conversation. We will allow this for now, but in the future this will be handled by moving the queue to ISAR

andchiind:
Relevant new issue: https://github.com/equinor/flotilla/issues/1891

backend/api/Services/MissionSchedulingService.cs
```diff
@@ -452,33 +424,9 @@ private async Task StartMissionRun(MissionRun queuedMissionRun)
...
-            var missionRun = await missionRunService.ReadById(missionRunId, readOnly: true);
-            if (missionRun is null)
-            {
-                string errorMessage = $"Mission run with Id {missionRunId} was not found in the database";
-                logger.LogError("{Message}", errorMessage);
-                throw new MissionRunNotFoundException(errorMessage);
-            }
-
-            if (robot.MissionQueueFrozen && missionRun.MissionRunType != MissionRunType.Emergency)
+            if (robot.MissionQueueFrozen && !(missionRun.IsEmergencyMission() || missionRun.IsReturnHomeMission()))
```
oysand, line 429: how is the ordering on this check? Would it go ((frozen && not emergency) or returnhomemission)?

andchiind: frozen && not (emergency or returnhome)

oysand (approved):
LGTM

PR #1890, opened 2024-12-17
Improve inspections view stability

mrica-equinor:

### DIFF
--- a/backend/api/Controllers/InspectionController.cs
+++ b/backend/api/Controllers/InspectionController.cs
@@ -1,5 +1,4 @@
-﻿using System.Globalization;
-using Api.Controllers.Models;
+﻿using Api.Controllers.Models;
 using Api.Database.Models;
 
 using Api.Services;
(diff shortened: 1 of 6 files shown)


backend/api/Controllers/InspectionController.cs
```diff
@@ -61,48 +60,19 @@ public async Task<ActionResult<TagInspectionMetadata>> Create([FromRoute] string
...
-            try
-            {
-                byte[] inspectionStream = await inspectionService.FetchInpectionImage(inspectionData.BlobName, inspectionData.BlobContainer, inspectionData.StorageAccount);
-                return File(inspectionStream, "image/png");
-            }
-            catch (Azure.RequestFailedException)
-            {
-                return NotFound($"Could not find inspection blob {inspectionData.BlobName} in container {inspectionData.BlobContainer} and storage account {inspectionData.StorageAccount}.");
-            }
+            return File(inspectionStream, "image/png");
```
eivindsjovold, line 75: Consider using Enums

backend/api/Controllers/InspectionController.cs
```diff
@@ -61,48 +60,19 @@ public async Task<ActionResult<TagInspectionMetadata>> Create([FromRoute] string
         /// </remarks>
         [HttpGet]
         [Authorize(Roles = Role.User)]
-        [Route("{installationCode}/{taskId}/taskId")]
+        [Route("{installationCode}/{isarInspectionId}/isarInspectionId")]
```
eivindsjovold, line 63: Should we reference isar by name, or should we try to implicitly not couple isar and flotilla?

backend/api/Services/InspectionService.cs
```diff
@@ -9,15 +9,20 @@
...
 {
     public interface IInspectionService
     {
         public Task<byte[]> FetchInpectionImage(string inpectionName, string installationCode, string storageAccount);
         public Task<Inspection> UpdateInspectionStatus(string isarTaskId, IsarTaskStatus isarTaskStatus);
         public Task<Inspection?> ReadByIsarTaskId(string id, bool readOnly = true);
+        public Task<Inspection?> ReadByIsarInspectionId(string id, bool readOnly = true);
         public Task<Inspection?> AddFinding(InspectionFindingQuery inspectionFindingsQuery, string isarTaskId);
         public Task<IDAInspectionDataResponse?> GetInspectionStorageInfo(string inspectionId);
+        public Task<byte[]?> GetInspectionImage(string installationCode, string isarInspectionId);
```
eivindsjovold, line 24: consider renaming to GetInspectionData

backend/api/Services/InspectionService.cs
```diff
@@ -142,11 +152,61 @@ private IQueryable<Inspection> GetInspections(bool readOnly = true)
...
+
+            var inspectionData = await GetInspectionStorageInfo(inspection.IsarInspectionId);
+
+            if (inspectionData == null)
+            {
+                logger.LogError($"Could not find inspection data for inspection with isar Id {inspection.IsarInspectionId}.");
+                return null;
+            }
+
+            if (!inspectionData.BlobContainer.ToLower(CultureInfo.CurrentCulture).Equals(installationCode.ToLower(CultureInfo.CurrentCulture), StringComparison.Ordinal))
```
eivindsjovold, line 194: Consider more readable check

backend/api/Services/InspectionService.cs
```diff
@@ -142,11 +152,61 @@ private IQueryable<Inspection> GetInspections(bool readOnly = true)
...
+            {
+                logger.LogError($"Could not find inspection data for inspection with isar Id {inspection.IsarInspectionId} because blob name {inspectionData.BlobName} does not match installation {installationCode}.");
```
eivindsjovold, line 196: Do we compare name of the blob, or blob location?

backend/api/Controllers/InspectionController.cs
tsundvoll: I think it is no need to make it more general than we need to, it is fine to reference ISAR.

In the opposite way I agree, ISAR should not need to know about Flotilla; but Flotilla needs to know about ISAR, since it makes calls to it anyways.

mrica-equinor: Since we have and inspectionId and an isarInspectionId, there is really no other way to reference the isar inspection Id, especially since it is generated by isar. We can have this discussion another time, but not relevant for this PR i think

backend/api/Services/InspectionService.cs
mrica-equinor: This is to make sure the blob container has the same name as the installation code, a bit reduntant but will help in case something goes wrong with the naming of the containers in the storage account. Agreed with Thomas it's worth keeping

mrica-equinor: The log was wrong but the check was fine

mrica-equinor:
Closed as duplicated of #1898

PR #1898, opened 2024-12-18
Simplify inspection controller

tsundvoll:
by moving more functionality to the inspection service and avoiding a call to the Flotilla database
by using the ISAR Inspection ID directly from the frontend to make the call to the IDA API

### DIFF
--- a/backend/api/Services/InspectionService.cs
+++ b/backend/api/Services/InspectionService.cs
@@ -14,11 +14,7 @@ namespace Api.Services
 {
     public interface IInspectionService
     {
-        public Task<byte[]?> FetchInpectionImage(
-            string inpectionName,
-            string installationCode,
-            string storageAccount
-        );
+        public Task<byte[]?> FetchInpectionImageFromIsarInspectionId(string isarInspectionId);
         public Task<Inspection> UpdateInspectionStatus(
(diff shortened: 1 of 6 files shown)


tsundvoll:
We need to implement the corresponding changes on the frontend before we take in this

backend/api/Services/InspectionService.cs
```diff
@@ -110,7 +113,7 @@ private IQueryable<Inspection> GetInspections(bool readOnly = true)
             return inspection;
         }
 
-        public async Task<IDAInspectionDataResponse?> GetInspectionStorageInfo(string inspectionId)
+        private async Task<IDAInspectionDataResponse?> GetInspectionStorageInfo(string inspectionId)
```
oysand, line 116: For it to work in prod we would need change CallApiForUserAsync to CallApiForAppAsync below

andchiind: We could in that case have an environment check here and use the correct function based on that. Then we can test it by changing the local environment to prod before merging it

oysand: I think it should be CallApiForApp for all environments

mrica-equinor (approved):
LGTM :)

PR #1916, opened 2024-12-23
Add constant string to use query key

olaals:

I think it is common to add a constant string to each useQuery https://tanstack.com/query/v5/docs/framework/react/guides/query-keys

I suspect that it will try to re-use the same cache if another useQuery also only has isarTaskId


andchiind:
See https://github.com/equinor/flotilla/pull/1918

PR #1919, opened 2025-01-02
Unify return to home and docking naming

### LINKED ISSUE #1793: "Drive to docking station" and "Return to home" missions drive to same place but sound like different places
These two missions are driving to either "Docking station" or "home". This makes it seem like home and docking station are different places but in the current implementation they are always the same. This can cause unnecessary confusion for a user.

Image

Image

If they are the same they should be referred to as the same from the users point of view. Suggest changing the name of the missions to either use "docking" or "home" and not both.

Clearer for user

N/A

andchiind:
Closes #1793 


oysand (approved):
LGTM

PR #1920, opened 2025-01-02
Make event handlers more deterministic

### LINKED ISSUE #1891: Make MissionEventHandler events pass the relevant robot state in the event arguments
Currently these events trigger mechanics based on user interactions or ISAR updates which updated Flotilla state, but they still read in the current robot state again when the event is handled. This leads to nondeterministic behaviour where event handlers can no longer handle the event as the state they are reading in is inconsistent with what was read in when triggering the event. I suggest that the events in MissionEventHandler should pass along the relevant state when the event was triggered, so we can be sure that the event is handled as intended when we call the event. For instance, when triggering OnRobotAvailable, we want a mission to be started, but if the state is read in again after the event is triggered, we can end up with inconcistent state which soft locks us from starting a mission. If the robot becomes available we want to guarantee that at least one event handler tries to run the mission. Trying to many times is not an issue if we handle the ISAR return values correctly.

Once this change is made, we can also make OnIsarStatus in MqttEventHandler.cs use the OnRobotAvailable event instead of calling MissionScheduling.StartNextMissionRunIfSystemIsAvailable.

It will make the system truly event-based as we can rely on event handlers to handle the event instead of the perceived current state of the system which may be inconcistent at the time the event is handled.

N/A

andchiind:
Closes #1891 

### DIFF
--- a/backend/api/EventHandlers/MissionEventHandler.cs
+++ b/backend/api/EventHandlers/MissionEventHandler.cs
@@ -109,24 +109,10 @@ namespace Api.EventHandlers
 
         private async void OnRobotAvailable(object? sender, RobotAvailableEventArgs e)
         {
-            _logger.LogInformation(
-                "Triggered RobotAvailable event for robot ID: {RobotId}",
-                e.RobotId
-            );
-            var robot = await RobotService.ReadById(e.RobotId, readOnly: true);
-            if (robot == null)
-            {
-                _logger.LogError(
-                    "Robot with ID: {RobotId} was not found in the database",
-                    e.RobotId
-                );
-                return;
-            }
-
             _startMissionSemaphore.WaitOne();
             try
             {
-                await MissionScheduling.StartNextMissionRunIfSystemIsAvailable(robot);
+                await MissionScheduling.StartNextMissionRunIfSystemIsAvailable(e.Robot);
             }
             catch (MissionRunNotFoundException)
             {
--- a/backend/api/EventHandlers/MissionEventHandler.cs
+++ b/backend/api/EventHandlers/MissionEventHandler.cs
@@ -126,23 +126,16 @@ namespace Api.EventHandlers
 
         private async void OnSendRobotToDockTriggered(object? sender, RobotEmergencyEventArgs e)
         {
+            var robot = e.Robot;
+
             _logger.LogInformation(
                 "Triggered EmergencyButtonPressed event for robot ID: {RobotId}",
-                e.RobotId
+                robot.Id
             );
-            var robot = await RobotService.ReadById(e.RobotId, readOnly: true);
-            if (robot == null)
-            {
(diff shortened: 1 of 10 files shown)


backend/api/EventHandlers/MissionEventHandler.cs
```diff
@@ -109,24 +107,16 @@ private async void OnMissionRunCreated(object? sender, MissionRunCreatedEventArg
 
         private async void OnRobotAvailable(object? sender, RobotAvailableEventArgs e)
         {
-            _logger.LogInformation(
-                "Triggered RobotAvailable event for robot ID: {RobotId}",
-                e.RobotId
-            );
-            var robot = await RobotService.ReadById(e.RobotId, readOnly: true);
-            if (robot == null)
+            if (e.Robot == null)
```
oysand, line 110: Can the robot be required to be a part of the event args so this can be known to never be null?

backend/api/EventHandlers/MissionEventHandler.cs
```diff
@@ -109,24 +107,16 @@ private async void OnMissionRunCreated(object? sender, MissionRunCreatedEventArg
...
-                );
+                _logger.LogError("OnRobotAvailable was triggered with null Robot value");
                 return;
             }
 
             _startMissionSemaphore.WaitOne();
             try
             {
-                await MissionScheduling.StartNextMissionRunIfSystemIsAvailable(robot);
+                await MissionScheduling.StartNextMissionRunIfSystemIsAvailable(e.Robot);
```
oysand, line 119: rather than passing the entire robot object, would it make sense to only pass the relevant information about the robot needed to start the next mission

backend/api/EventHandlers/MissionEventHandler.cs
andchiind: I will remove the null check. Seems like it will never be null

backend/api/EventHandlers/MissionEventHandler.cs
andchiind: This is out of scope for this PR

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -573,7 +573,7 @@ private async void OnIsarBatteryUpdate(object? sender, MqttReceivedArgs mqttArgs
                     robot.Name
                 );
                 EmergencyActionService.SendRobotToDock(
-                    new RobotEmergencyEventArgs(robot.Id, RobotFlotillaStatus.Recharging)
+                    new RobotEmergencyEventArgs(robot, RobotFlotillaStatus.Recharging)
```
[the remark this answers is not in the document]
andchiind: The whole object is used later on, so changing this is out of scope of this PR

oysand (approved):
LGTM

PR #1930, opened 2025-01-07
Do partial robot updates with SignalR

### LINKED ISSUE #1796: Report specific robot attribute updates with SignalR
Currently when a field in the robot model is updated, we send the whole robot object to the frontend with SignalR. The issue with this is that when we update different fields around the same time, one of the updates may be ignored in the second signalR message. By only sending the specific field on the robot which has been updated we avoid this issue. There may be a potential backend solution as well, but this might be harder to implement.

The robot values displayed on the frontend (status, battery level, pressure etc) will be more up to date. Currently we may need to wait for a new signalR message to get the previous battery level for instance.

N/A

andchiind:
Closes #1796 

### DIFF
--- a/backend/api/Controllers/Models/RobotAttributeResponse.cs
+++ b/backend/api/Controllers/Models/RobotAttributeResponse.cs
@@ -1,0 +1,37 @@
+﻿using System.Text.Json.Serialization;
+using Api.Database.Models;
+
+namespace Api.Controllers.Models
+{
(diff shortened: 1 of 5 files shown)


andchiind:
> This pull request has automatically been marked as stale as there has been no activity for 30 days.

How dare you

andchiind:
Closing this for now as it seems to not be the ideal approach to the problem.

PR #1967, opened 2025-01-17
Handle empty media stream config

Eddasol:
Dependent on https://github.com/equinor/isar/pull/698

### DIFF
--- a/backend/api/Services/IsarService.cs
+++ b/backend/api/Services/IsarService.cs
@@ -19,7 +19,7 @@ namespace Api.Services
 
         public Task<IsarMission> StartMoveArm(Robot robot, string armPosition);
 
-        public Task<MediaConfig> GetMediaStreamConfig(Robot robot);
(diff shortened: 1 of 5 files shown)


backend/api/Services/IsarService.cs
```diff
@@ -317,11 +317,13 @@ public async Task<MediaConfig> GetMediaStreamConfig(Robot robot)
                 logger.LogError("{Message}: {ErrorResponse}", message, errorResponse);
                 throw new ConfigException(message);
             }
-            if (response.Content is null)
```
andchiind, line 320: Is this equivalent to "if (response.StatusCode == HttpStatusCode.NoContent)"?

andchiind: The never one looks better, just wondering if it covers the same bases

Eddasol: No, this line doesn't seem to trigger regardless of content

andchiind (approved):
LGTM

PR #1974, opened 2025-01-23
Update inspections after isar mission accepted

mrica-equinor:

### DIFF
--- a/backend/api/Services/MissionRunService.cs
+++ b/backend/api/Services/MissionRunService.cs
@@ -321,6 +321,27 @@ namespace Api.Services
             DetachTracking(context, missionRun!);
         }
 
+        public async Task UpdateWithInspections(MissionRun missionRun)
+        {
+            if (missionRun.Robot is not null)
+            {
+                context.Entry(missionRun.Robot).State = EntityState.Unchanged;
+            }
+            if (missionRun.InspectionArea is not null)
+            {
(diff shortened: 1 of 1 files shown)


oysand (approved):
LGTM

backend/api/api.csproj
```diff
@@ -30,7 +30,9 @@
     <PackageReference Include="Swashbuckle.AspNetCore" Version="6.6.2" />
     <PackageReference Include="MQTTnet" Version="4.3.6.1152" />
     <PackageReference Include="MQTTnet.Extensions.ManagedClient" Version="4.3.6.1152" />
-    <PackageReference Include="System.Linq.Dynamic.Core" Version="1.5.1" />
+    <PackageReference Include="System.Linq.Dynamic.Core" Version="1.5.1" >
+        <NoWarn>NU1903</NoWarn>
+    </PackageReference>
```
Christdej, line 35: Remove this and merge it even if the pipeline fails

Christdej (changes requested):

Christdej (approved):
LGTM

backend/api/api.csproj
oysand: There was a test failing with this in. Good to have it temporarily so we can know that there are no tests failing, then remove the ignore so that the pipeline fails again

Christdej: So the test was failing but was not showing because this error stopped the pipeline before the test could fail?

PR #1997, opened 2025-01-30
Fix typos

Eddasol:

### DIFF
--- a/frontend/best_practices.md
+++ b/frontend/best_practices.md
@@ -1,13 +1,13 @@
-# Flotilla frontend development best practises
+# Flotilla frontend development best practices
 
--   [Flotilla frontend development best practises](#flotilla-frontend-development-best-practises)
+-   [Flotilla frontend development best practices](#flotilla-frontend-development-best-practices)
     -   [Setup](#setup)
         -   [Prettier](#prettier)
         -   [ESLint](#eslint)
     -   [Folder structure](#folder-structure)
     -   [Components](#react-components)
         -   [React arguments](#react-arguments)
-        -   [React functions](#react-state)
+        -   [React functions](#react-functions)
         -   [React state](#react-state)
         -   [Nesting](#nesting)
     -   [Contexts](#contexts)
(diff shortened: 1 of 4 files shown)


frontend/best_practices.md
```diff
@@ -1,13 +1,13 @@
...
--   [Flotilla frontend development best practises](#flotilla-frontend-development-best-practises)
+-   [Flotilla frontend development best practices](#flotilla-frontend-development-best-practices)
     -   [Setup](#setup)
         -   [Prettier](#prettier)
         -   [ESLint](#eslint)
     -   [Folder structure](#folder-structure)
     -   [Components](#react-components)
         -   [React arguments](#react-arguments)
-        -   [React functions](#react-state)
+        -   [React functions](#react-functions)
```
anetteu, line 10: Verify if this is correct

frontend/best_practices.md
```diff
@@ -35,41 +35,41 @@ npx eslint [path to src]
...
 
     const foo = (x:number, y: number) => x + y
 
 ## React components
 
-The flotilla frontend is programmed in Typescript using the React framework. In this framework all HTML components are defined using React components, which are pieces of code which define what the HTML should look like depending on the state of the system. We only utilise React Functions, which means that components are described in the form of functions which accept the component attributes are function arguments, and return the HTML at the end of the function. React classes should be avoided.
+The Flotilla frontend is programmed in Typescript using the React framework. In this framework, all HTML components are defined using React components, which are pieces of code that define what the HTML should look like depending on the state of the system. We only utilize React Functions, which means that components are described in the form of functions that accept the component attributes as function arguments and return the HTML at the end of the function. React classes should be avoided.
```
anetteu, line 66:
```suggestion
The Flotilla frontend is programmed in TypeScript using the React framework. In this framework, all HTML components are defined using React components, which are pieces of code that define what the HTML should look like depending on the state of the system. We only utilize React Functions, which means that components are described in the form of functions that accept the component attributes as function arguments and return the HTML at the end of the function. React classes should be avoided.
```

anetteu: For consistency with other changes

frontend/best_practices.md
```diff
@@ -91,7 +91,7 @@ export const RobotComponent = ({ mission, setColor } : {mission: MissionDefiniti
 
 ### React state
 
-React provides the useState hook which allows the user to define variables which persist between renders. Although it is tempting to use these for all variables, it is worth noting that unless the variable has some information which needs to be kept for another render in order to update another variable, it can be kept as a normal Typescript variable.
+React provides the useState hook which allows the user to define variables that persist between renders. Although it is tempting to use these for all variables, it is worth noting that unless the variable has some information which needs to be kept for another render in order to update another variable, it can be kept as a normal Typescript variable.
```
anetteu, line 94:
```suggestion
React provides the useState hook which allows the user to define variables that persist between renders. Although it is tempting to use these for all variables, it is worth noting that unless the variable has some information which needs to be kept for another render in order to update another variable, it can be kept as a normal TypeScript variable.
```

anetteu (approved):

PR #2007, opened 2025-02-03
Fix problem of duplicate robot poses

### LINKED ISSUE #1999: Bug if echo mission has two inspections  selected on mission planner
**Describe the bug**
When Thomas and I were at NL we tried to schedule a mission that had both normal picture and thermal picture selected and this would crash. 

**To Reproduce**
Create a mission in echo mission planner with two inspections on the same tag and shcedule it. 

Image

**Expected behavior**
This mission should run, two inspection tasks should be created in the mission, each one with one inspection.

**Screenshots**

andchiind:
Closes #1999 


Eddasol (approved):
LGTM

PR #2013, opened 2025-02-07
Simplify getAccessible functions

Christdej:

### DIFF
--- a/backend/api/Services/AreaService.cs
+++ b/backend/api/Services/AreaService.cs
@@ -76,7 +76,8 @@ namespace Api.Services
 
         public async Task<Area?> ReadById(string id, bool readOnly = true)
         {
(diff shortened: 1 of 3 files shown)


andchiind (commented):
Might this PR make the code slightly less safe, since if someone now accidentally calls GetAreas for instance, they might not be aware that they are sharing information the user should not have access to?

Christdej:
> Might this PR make the code slightly less safe, since if someone now accidentally calls GetAreas for instance, they might not be aware that they are sharing information the user should not have access to?

Potientially, but I think that we should not base the safety of the system om people remembering to use the correct function or add the correct .Where from a database call. So we should perhaps look into how we can fix that?

Christdej:
To be handled in https://github.com/equinor/flotilla/issues/2093

PR #2015, opened 2025-02-10
Remove Area

Christdej:
- [x] A test have been written


Christdej:
Will now give warnings on Return to Home missions if they fail becuase we dont have an Area, awaiting Return Home functionallity to be removed aswell.

backend/api/Controllers/MissionSchedulingController.cs
```diff
@@ -397,7 +339,6 @@ [.. missionInspectionAreaNames]
                     Name = missionDefinition.Name,
                     InspectionFrequency = scheduledMissionQuery.InspectionFrequency,
                     InstallationCode = scheduledMissionQuery.InstallationCode,
-                    InspectionArea = area.InspectionArea,
```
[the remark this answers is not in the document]
andchiind: It either uses a existing missionDefinition or creates a new mission definition here. Since this is the only way new mission definitions are created, and since we don't give the mission definition an inspection area, none of the mission definitions and mission runs will have inspection areas. And since the robots get inspection areas from the first mission they run, they also won't have inspection areas.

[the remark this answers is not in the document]
Christdej: Yeah, that is true. Im trying to find a good way of adding inspection area while still removing area. Any suggestions?

backend/api/Controllers/MissionSchedulingController.cs
```diff
@@ -436,19 +377,6 @@ [.. missionInspectionAreaNames]
                 await missionDefinitionService.Create(scheduledMissionDefinition);
             }
 
-            if (
-                missionRun.Robot.CurrentInspectionArea != null
-                && !await localizationService.RobotIsOnSameInspectionAreaAsMission(
-                    missionRun.Robot.Id,
-                    missionRun.InspectionArea!.Id
-                )
-            )
```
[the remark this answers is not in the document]
Christdej: Perhaps it should yeah. Also need to remove the `localizationService.RobotIsOnSameInspectionAreaAsMission` function perhaps.

backend/api/Controllers/MissionSchedulingController.cs
```diff
@@ -313,54 +312,6 @@ [FromBody] ScheduledMissionQuery scheduledMissionQuery
...
-                );
-                area = sortedAreas.Skip(1).First();
-            }
-            if (area == null)
-            {
-                logger.LogError(
-                    $"Mission {missionDefinition.Name} doesn't have any tags with valid area."
-                );
-                return NotFound($"No area found for mission '{missionDefinition.Name}'.");
-            }
```
andchiind, line 362: Could we perhaps keep this block of code in order to find which inspection area to assign the mission definition? We do not need to have an Area model in the database to do this, as it fetches them from an external API (STID). We do however need the inspection area to have an understanding of STID areas though.

andchiind: Originally we planned InspectionAreas to be a list or graph of Areas. Could we potentially store the references to STID areas as a list in InspectionAreas still, but without storing them as Area objects, for instance by only storing the area name?

Christdej:
Rest of work to be done documented in:
https://github.com/equinor/flotilla/issues/2024

backend/api/Controllers/MissionSchedulingController.cs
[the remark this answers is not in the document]
andchiind: I guess here we should check which polygon the tasks fit inside.

backend/api/Database/Context/InitDb.cs
```diff
@@ -375,7 +287,7 @@ private static List<MissionDefinition> GetMissionDefinitions()
 
         private static List<MissionTask> GetMissionTasks()
         {
-            var url = new Uri("https://stid.equinor.com/hua/tag?tagNo=ABCD");
+            var url = new Uri("https://test.equinor.com/hua/tag?tagNo=ABCD");
```
mrica-equinor, line 290: this url doesnt seem to point to anything, is it a problem?

Christdej: Its an example url / dummy url, I can rename it to more clearly show this

backend/api/Services/MissionSchedulingService.cs
```diff
@@ -103,22 +103,22 @@ public async Task StartNextMissionRunIfSystemIsAvailable(Robot robot)
...
-                return;
             }
 
             if (robot.CurrentInspectionArea == null)
             {
-                await robotService.UpdateCurrentInspectionArea(
-                    robot.Id,
-                    missionRun.InspectionArea!.Id
+                logger.LogError(
+                    "Robot {RobotName} with Id {RobotId} is not on an inspection area.",
```
mrica-equinor, line 111: This is old code that was used to set the current inspection area on the robot for the first time (localise the robot the first time). The code might have changed, but did you check that the inspection area is set somewhere else?

Christdej: The change we landed on (after some discussions) was that we set the currentInspectionArea through an endpoint, which will be part of the setup required when moving the robot to new inspectionArea. There is currently no other way the inspectionArea is set (or should not be atleast)

mrica-equinor: I see, if you've had a discussion and that was agreed then I'll approve and if long term it doesnt work we can think about this again

mrica-equinor (approved):
LGTM :)

Issue #1755, opened 2024-10-08
Improve fullscreen video stream resizing


sondreo:
Currently, when resizing the browser window while a video stream fullscreen is toggled, the video stream window does not dynamically scale to fit the updated size of the browser window.

The size of the `FullScreenCard` is [dynamically calculated](https://github.com/equinor/flotilla/blob/4078e5bc7b168140e08678a79415a0bc4df20194/frontend/src/components/Pages/MissionPage/VideoStream/FullScreenVideo.tsx#L51-L58) based on the browser window, as when finding the card width:
```typescript
const cardWidth = () => {
    const availableInnerHeight = window.innerHeight - 9 * 16
    const availableInnerWidth = window.innerWidth - 2 * 16
    const coverageFactor = 0.9
    const aspectRatio = 16 / 9
    return Math.round(
        Math.min(coverageFactor * availableInnerWidth, aspectRatio * coverageFactor * availableInnerHeight)
    )
}
```

The width is [calculated](https://github.com/equinor/flotilla/blob/4078e5bc7b168140e08678a79415a0bc4df20194/frontend/src/components/Pages/MissionPage/VideoStream/FullScreenVideo.tsx#L100) when creating the `FullScreenCard` component, however does not seem to be dynamically recalculated.

olaals:
@sondreo 

Does this issue depend on your PR? How can i test this? I have setup flotilla locally

PR #2019, opened 2025-02-12
Use default HTML5 fullscreen for video

### LINKED ISSUE #1914: Use default/standard HTML5 solution for full screen video
For the video stream solution it can simplify the code and prevent current and feature bugs by using standard full screen solutions when expanding the video preview. If the video reloads we need to ensure that the same stream is being edited, so that the user does not need to re-open the full screen window.

It will let the browser decide how to display the video.

N/A

andchiind:
Closes #1914, #1755

Can be tested by adding this code to line 88 in RobotPage.tsx:

useEffect(() => {
    navigator.mediaDevices.getUserMedia({video:true}).then((stream) => setVideoMediaStreams([stream.getVideoTracks()[0]]))
},[])


frontend/src/components/Pages/MissionPage/VideoStream/VideoStreamWindow.tsx
```diff
@@ -13,61 +11,41 @@ const VideoStreamContent = styled.div`
...
 
     const videoCards = videoStreams.map((videoStream, index) => (
         <VideoStreamCard
             key={index}
             videoStream={new MediaStream([videoStream])}
             videoStreamName={undefined}
-            toggleFullScreenMode={toggleFullScreenMode}
-            setFullScreenStream={updateFullScreenStream}
+            videoStreamId={'videostreamid-' + index}
+            setFullScreen={() => updateFullScreen('videostreamid-' + index)}
```
mrica-equinor, line 41: I think i would rather just pass on the updateFullScreen function and then inside the video stream card since we have the videoStreamId we specifically update it there, instead of so early in the code. 

In fact updateFullScreen could be moved to the other file too if I am not missing something, it isnt used anywhere in this file. Only setFullScreenElementId needs to be passed down

mrica-equinor (approved):
LGTM :)

Issue #1543, opened 2024-04-10
Refactor Inspection Findings Report


mrica-equinor:
**Describe the improvement you would like to see**
Refactor the Inspections findings report and think if it is possible to make it event based instead of a servie that is always running in the background.

**How will this change existing functionality?**
Current functionality shouldn't be affected.

**How will this improvement affect the current Threat Model?**
Leave blank for maintainers to fill out if you are uncertain about this.

PR #2025, opened 2025-02-13
Remove inspection findings functionality

mrica-equinor:


backend/api/EventHandlers/TeamsMessageEventHandler.cs
```diff
@@ -6,6 +6,8 @@
 using Api.Utilities;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
```
andchiind: Could we replace these with default JSON handlers while we're at it? Shouldn't be a big change :)

andchiind:
Could we add the csharpier commit to the previous commit?

backend/api/EventHandlers/TeamsMessageEventHandler.cs
mrica-equinor: I can, but then i have to rewrite the current notification system on teams when there are issues with the robot telemetery

mrica-equinor: Which is a bit of scope creep but makes sense to sdress it now

andchiind: nw, we can do it in another PR if we make an issue on it

backend/api/Database/Models/Inspection.cs
```diff
@@ -117,8 +117,6 @@ or InspectionStatus.Successful
 
         public DateTime? EndTime { get; private set; }
 
-        public List<InspectionFinding> InspectionFindings { get; set; }
```
andchiind, line 120: Remember to add a migration as well before merging

backend/api/EventHandlers/TeamsMessageEventHandler.cs
mrica-equinor: adressed

mrica-equinor:
Closes #1543

andchiind (approved):

PR #2028, opened 2025-02-17
Introduce inspection area polygon

Christdej:


Christdej:
Remaining work could be to remove Plant of Installation, documented here:
https://github.com/equinor/flotilla/issues/2032

backend/api/Services/InspectionAreaService.cs
```diff
@@ -127,6 +134,83 @@ public async Task<IEnumerable<InspectionArea>> ReadByInstallation(
...
+                var xj = polygon[j].X;
+                var yj = polygon[j].Y;
+
+                var intersect =
+                    ((yi > y) != (yj > y)) && (x < (xj - xi) * (y - yi) / (yj - yi) + xi);
+                if (intersect)
+                {
+                    inside = !inside;
+                }
+            }
```
tsundvoll, line 209: Should we not count how many intersections and check if n_intersections % 2 == 1?
https://medium.com/@girishajmera/exploring-algorithms-to-determine-points-inside-or-outside-a-polygon-038952946f87

tsundvoll: Or wait, I get it. It is just flipped every time there is an intersection. This works

Christdej: Yeah, just implemented the ray casting algorithm

backend/api/Services/InstallationService.cs
```diff
@@ -137,6 +144,33 @@ public async Task<Installation> Update(Installation installation)
             return installation;
         }
 
+        public async Task EnsureRobotIsOnSameInstallationAsMission(
```
tsundvoll, line 147: Can we name it AssertRobotIs...?

Ensure sounds to me like we are changing it to become right if it is not right already
```suggestion
        public async Task AssertRobotIsOnSameInstallationAsMission(
```

tsundvoll: `context.Database.EnsureCreated();`
from https://github.com/equinor/flotilla/blob/a0a603f0c4e4f07a78317b7b6a4000e2ca573344/backend/api/Configurations/CustomServiceConfigurations.cs#L44

Is one example where ensure makes sense. From its documentation is says:
```
Ensures that the database for the context exists.

• If the database exists and has any tables, then no action is taken. Nothing is done to ensure the database schema is compatible with the Entity Framework model.
• If the database exists but does not have any tables, then the Entity Framework model is used to create the database schema.
• If the database does not exist, then the database is created and the Entity Framework model is used to create the database schema.
```

In other words, if the condition is not met; actions will be taken. 

Since that is not the case in our method here, I think we should refrain from using "ensure" in the name

backend/api.test/Services/InspectionAreaService.cs
```diff
@@ -0,0 +1,107 @@
...
+            );
+            inspectionArea.AreaPolygonJson =
+                @"{
+                    ""zmin"": 0,
+                    ""zmax"": 10,
+                    ""positions"": [
+                        { ""x"": 0, ""y"": 0 },
+                        { ""x"": 0, ""y"": 10 },
+                        { ""x"": 10, ""y"": 10 },
+                        { ""x"": 10, ""y"": 0 }
```
tsundvoll, line 57: Here, we are just checking for a square polygon. Maybe sufficient for now, but an idea here could be to define a more complex polygon and use property-based testing to test it. Issue for another time probably.

Christdej: Yeah, though that was good enough for the test without having to think to much

backend/api/Services/InstallationService.cs
Christdej: I agree I just moved the function from one service to another, but I will rename it as it fits better to what it actually does

backend/api/Controllers/MissionSchedulingController.cs
```diff
@@ -361,6 +360,18 @@ [.. missionInspectionAreaNames]
...
+            if (
+                !inspectionAreaService.MissionTasksAreInsideInspectionAreaPolygon(
+                    missionTasks,
+                    area.InspectionArea
+                )
+            )
+            {
+                return Conflict(
+                    $"The tasks of the mission are not inside the inspection area of the robot"
+                );
```
Eddasol, line 372: Is conflict correct error type or is bad request more descriptive? Conflict indicates that the request didn't success at the moment due to some conflict with the current state, but this request will never succeed with the given input.

backend/api/Services/InspectionAreaService.cs
```diff
@@ -127,6 +134,83 @@ public async Task<IEnumerable<InspectionArea>> ReadByInstallation(
...
+                return true; // No polygon defined, so all tasks are inside
+            }
+
+            var inspectionAreaPolygon = JsonSerializer.Deserialize<InspectionAreaPolygon>(
+                inspectionArea.AreaPolygonJson
+            );
+
+            if (inspectionAreaPolygon == null)
+            {
+                return true; // Invalid polygon, so all tasks are inside
```
Eddasol, line 153: Should this trigger a warning before returning?

backend/api/Controllers/Models/CreateInspectionAreaQuery.cs
```diff
@@ -1,9 +1,12 @@
-﻿namespace Api.Controllers.Models
+﻿using Api.Services;
+
+namespace Api.Controllers.Models
```
Eddasol, line 3: Should perhaps the definition of InspectionAreaPolygon move to a models folder so we dont import services to models and models to services?

backend/api/Controllers/InspectionAreaController.cs
```diff
@@ -219,7 +257,7 @@ await inspectionAreaService.ReadByInstallationAndPlantAndName(
             catch (Exception e)
```
Eddasol, line 257: Suggest catching InvalidPolygonException before generic catch to improve error message

backend/api/Database/Models/InspectionArea.cs
```diff
@@ -19,5 +19,7 @@ public class InspectionArea
         [Required]
         [MaxLength(200)]
         public string Name { get; set; }
+
+        public string? AreaPolygonJson { get; set; } = "";
```
Eddasol, line 23: Why have it as a string here and not just use the InspectionAreaPolygon datatype?

backend/api/Services/InspectionAreaService.cs
```diff
@@ -261,4 +371,25 @@ public void DetachTracking(FlotillaDbContext context, InspectionArea inspectionA
...
+    {
+        [JsonPropertyName("zmin")]
+        public double ZMin { get; set; }
+
+        [JsonPropertyName("zmax")]
+        public double ZMax { get; set; }
+
+        [JsonPropertyName("positions")]
+        public List<XYPosition> Positions { get; set; } = [];
+    }
```
andchiind, line 385: This could be moved to the database models folder, and made [Owned]. That way inspectionAreas could include it, but it would not make a new table.

backend/api/Controllers/MissionSchedulingController.cs
Christdej: Yeah, I agree, Ill change it to BadRequest

backend/api/Services/InspectionAreaService.cs
Christdej: Would probably be best yeah =)

backend/api/Controllers/Models/CreateInspectionAreaQuery.cs
Christdej: Yeah, it should

backend/api/Services/InspectionAreaService.cs
andchiind: Actually, keeping it as a JSON could also be useful for simplifying adding new polygons. So I am OK with either approach.

backend/api/Services/InspectionAreaService.cs
```diff
@@ -127,6 +135,91 @@ public async Task<IEnumerable<InspectionArea>> ReadByInstallation(
...
+                inspectionArea.AreaPolygonJson
+            );
+
+            if (inspectionAreaPolygon == null)
+            {
+                logger.LogWarning(
+                    "Invalid polygon defined for inspection area {inspectionAreaName}",
+                    inspectionArea.Name
+                );
+                return true;
```
mrica-equinor, line 162: shouldn't this be false? is it for now while we don't have all polygons defined so we can still run missions?

backend/api/Services/InspectionAreaService.cs
```diff
@@ -127,6 +135,91 @@ public async Task<IEnumerable<InspectionArea>> ReadByInstallation(
...
+            {
+                var robotPosition = missionTask.RobotPose.Position;
+                if (
+                    !IsPositionInsidePolygon(
+                        inspectionAreaPositions,
+                        robotPosition,
+                        inspectionAreaZMin,
+                        inspectionAreaZMax
+                    )
+                )
```
mrica-equinor, line 178:
```suggestion
            foreach (var missionTask in missionTasks)
            {
                if (
                    !IsPositionInsidePolygon(
                        inspectionAreaPolygon.Positions,
                        missionTask.RobotPose.Position,
                        inspectionAreaPolygon.ZMin,
                        inspectionAreaPolygon.ZMax
                    )
                )
```

backend/api/Controllers/InspectionAreaController.cs
```diff
@@ -216,10 +255,15 @@ await inspectionAreaService.ReadByInstallationAndPlantAndName(
...
+            catch (InvalidPolygonException e)
+            {
+                logger.LogError(e, "Invalid polygon");
+                return BadRequest("Invalid polygon");
+            }
             catch (Exception e)
             {
                 logger.LogError(e, "Error while creating new inspection area");
-                throw;
+                return Conflict("Error while creating new inspection area");
```
Eddasol, line 266: 500 error since its an unexpected/unknown error?

Eddasol (approved):
LGTM :)

backend/api/Services/InspectionAreaService.cs
Christdej: Yeah, though was that we should allow if no polygon, but if there is a polygon and it is outside we should stop

backend/api/Controllers/InspectionAreaController.cs
Christdej: Yeah

backend/api/Services/InspectionAreaService.cs
Christdej: Fixed

Issue #2029, opened 2025-02-17
Add cronjob for automatically scheduling missions


Eddasol:
Add cronjob at midnight that checks when its necessary to schedule missions the coming day. This will schedule other cronjobs that schedules mission at desired times

PR #2033, opened 2025-02-19
Auto achedule missions

mrica-equinor:


mrica-equinor:
Closes #2033 [#2029](https://github.com/equinor/flotilla/issues/2029)

backend/api/Services/MissionDefinitionService.cs
```diff
@@ -26,6 +26,8 @@ public Task<List<MissionDefinition>> ReadByInspectionAreaId(
             bool readOnly = true
         );
 
+        public Task<List<MissionDefinition>?> ReadByHasInspectionFrequency(bool readOnly = true);
```
mrica-equinor, line 29:
```suggestion
        public Task<List<MissionDefinition>?> ReadByHasAutoScheduleFrequency(bool readOnly = true);
```

backend/api/HostedServices/InspectionFrequencyHostedService.cs
```diff
@@ -0,0 +1,206 @@
...
+using Api.Controllers.Models;
+using Api.Database.Models;
+using Api.Services;
+using Api.Services.MissionLoaders;
+using Api.Utilities;
+using Hangfire;
+
+namespace Api.HostedServices
+{
+    public class InspectionFrequencyHostedService : IHostedService, IDisposable
```
Eddasol, line 11: There are a lot of references to "InspectionFrequency" that should be changed

backend/api/HostedServices/InspectionFrequencyHostedService.cs
```diff
@@ -0,0 +1,206 @@
...
+            _scopeFactory.CreateScope().ServiceProvider.GetRequiredService<IRobotService>();
+
+        private IMissionLoader MissionLoader =>
+            _scopeFactory.CreateScope().ServiceProvider.GetRequiredService<IMissionLoader>();
+
+        public Task StartAsync(CancellationToken stoppingToken)
+        {
+            _logger.LogInformation("Inspection Frequency Hosted Service Running.");
+
+            var timeUntilMidnight = (DateTime.UtcNow.AddDays(1) - DateTime.UtcNow).TotalSeconds;
```
Eddasol, line 46: Want this just give number of seconds in a day? = 86400

backend/api/HostedServices/InspectionFrequencyHostedService.cs
```diff
@@ -0,0 +1,206 @@
...
+            _timer = new Timer(
+                DoWork,
+                null,
+                TimeSpan.FromSeconds(timeUntilMidnight),
+                TimeSpan.FromDays(1)
+            );
```
Eddasol, line 52: Have we thought trough what will happen if the pod dies? If we don't care when it will run we can just use 0 (or like 1minute) offset and it will run shortly after if the pod goes down

backend/api/HostedServices/InspectionFrequencyHostedService.cs
mrica-equinor: Great catch thank you! Forgot to select only the day part of the date
var timeUntilMidnight = (DateTime.UtcNow.Date.AddDays(1) - DateTime.UtcNow).TotalSeconds;

backend/api/HostedServices/InspectionFrequencyHostedService.cs
mrica-equinor: When the pod dies, it checks the time until midnight and then only does the work when it reaches the next midnight

mrica-equinor: Handled in the new commit

backend/api/Database/Models/AutoScheduleFrequency.cs
```diff
@@ -37,37 +37,27 @@ public void ValidateAutoScheduleFrequency()
...
-                    if (time <= nowLocalTimeOnly)
-                    {
-                        autoScheduleNext24Hours.Add(time - nowLocalTimeOnly);
-                    }
-                }
+                autoScheduleNext.AddRange(
+                    TimesOfDay
+                        .Where(time => time > nowLocalTimeOnly)
+                        .Select(time => time - nowLocalTimeOnly)
+                );
```
Eddasol, line 57: Have you thought about timezone difference when making this change?

Eddasol: Example: This runs at 00:00 UTC giving 02:00 for CET (summertime). If scheduling times contains timeOnlys in the range 00:00 to 01:59 those will never be added.

mrica-equinor: Yeah I think this will be handled in another PR, because we should never have Local time stored in the backend. So I will improve this in another PR to make sure everything in the backend is UTC so that this doesn't happen never. But it is tricky with keeping the local time in the frontend, will probs have to rewrite a lot of the frontend first, so I will not adress it now, since you want to test it this week. Otherwise testing of this needs to be moved to next week.

backend/api/HostedServices/InspectionFrequencyHostedService.cs
```diff
@@ -0,0 +1,211 @@
...
+
+            if (missionDefinitions == null)
+            {
+                _logger.LogInformation("No mission definitions with auto scheduling found.");
+                return;
+            }
+
+            var selectedMissionDefinitions = missionDefinitions.Where(m =>
+                m.AutoScheduleFrequency != null
+                && m.AutoScheduleFrequency.GetSchedulingTimesUntilMidnight() != null
```
Christdej, line 84: This might not work exactly how we expect because AutoScheduleFrequency is an owned property, and might not correctly be set to null once they have been set. Needs testing

mrica-equinor: Locally it has been working fine, it only fetches the ones with auto scheduling. This only is there because c# complains on the second line if we dont verify the first, the first will always be trie because it is iniciated as an empty list i think

Christdej: Nice

Issue #2042, opened 2025-02-21
Gray out "Send robots to dock" button if user does not have access to installation


mrica-equinor:
**Describe the improvement you would like to see**
We now have flotilla as a company app on insight and some people have been testing the page. I got feedback that the "Send robots to dock" button is clickable on the frontend. 

Obviously, if this user clicks the button, nothing will happen, but the button should be greyed out like the other elemtns in the frontend. 

<img width="443" alt="Image" src="https://github.com/user-attachments/assets/70c7075a-7135-4c0d-a4f1-db26ca587005" />

PR #2049, opened 2025-02-25
Redirect to asset page if no installation selected

mrica-equinor:

### DIFF
--- a/frontend/src/components/Pages/FrontPage/FrontPage.tsx
+++ b/frontend/src/components/Pages/FrontPage/FrontPage.tsx
@@ -4,6 +4,7 @@ import { InspectionOverviewSection } from 'components/Pages/InspectionPage/Inspe
 import { StopRobotDialog } from './MissionOverview/StopDialogs'
 import { tokens } from '@equinor/eds-tokens'
 import { MissionControlSection } from './MissionOverview/MissionControlSection'
+import { redirectIfNoInstallationSelected } from 'utils/RedirectIfNoInstallationSelected'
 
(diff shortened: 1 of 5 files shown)


mrica-equinor:
Closes #2042

frontend/src/utils/ValidateURL.tsx
```diff
@@ -0,0 +1,9 @@
+import { useInstallationContext } from 'components/Contexts/InstallationContext'
+import { config } from 'config'
+
+export function validateURL() {
+    const { installationCode, installationName } = useInstallationContext()
```
andchiind, line 5: Does using contexts inside non-component functions not cause an error/warning in the frontend?

andchiind (commented):
Looks good otherwise

frontend/src/utils/ValidateURL.tsx
andchiind: Could we rename the function so that it's clear that it will navigate away, and not just validate the inspection?

frontend/src/components/Pages/FrontPage/FrontPage.tsx
```diff
@@ -15,6 +16,8 @@ const StyledFrontPage = styled.div`
 `
 
 export const FrontPage = () => {
+    redirectIfNoInstallationSelected()
```
andchiind, line 19: If we wanted to be even more optimal we could useMemo, or a useEffect with empty conditionals here, to prevent this function from needing to be rerun on each render, but it's not too important.

PR #2112, opened 2025-03-20
Show map on missionDefinitionPage even if no robot

Eddasol:

### DIFF
--- a/backend/api/Services/MissionDefinitionService.cs
+++ b/backend/api/Services/MissionDefinitionService.cs
@@ -264,10 +264,6 @@ namespace Api.Services
                 .ThenInclude(area => area!.Installation)
                 .Include(missionDefinition => missionDefinition.Source)
                 .Include(missionDefinition => missionDefinition.LastSuccessfulRun)
-                .ThenInclude(missionRun => missionRun != null ? missionRun.Robot : null)
-                .Include(missionDefinition => missionDefinition.LastSuccessfulRun)
-                .ThenInclude(missionRun => missionRun != null ? missionRun.Tasks : null)!
-                .ThenInclude(missionTask => missionTask.Inspection)
                 .Include(missionDefinition => missionDefinition.InspectionArea)
                 .Where(
                     (m) =>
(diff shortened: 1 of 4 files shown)


Eddasol:
the 2 commits are different solution suggestion. Will be squashed before merging

backend/api/Services/MissionDefinitionService.cs
```diff
@@ -264,10 +264,6 @@ private IQueryable<MissionDefinition> GetMissionDefinitionsWithSubModels(
                 .ThenInclude(area => area!.Installation)
                 .Include(missionDefinition => missionDefinition.Source)
                 .Include(missionDefinition => missionDefinition.LastSuccessfulRun)
-                .ThenInclude(missionRun => missionRun != null ? missionRun.Robot : null)
-                .Include(missionDefinition => missionDefinition.LastSuccessfulRun)
-                .ThenInclude(missionRun => missionRun != null ? missionRun.Tasks : null)!
-                .ThenInclude(missionTask => missionTask.Inspection)
```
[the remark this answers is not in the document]
mrica-equinor: Think you need to include everything exceot lines 266 and 267 cause those were the ones introduced in the other PR. Probs other part os the code might need these imports as well (268, 269 ans 270)

Eddasol: Is there anything else that the mission definition page that actually use the lastSuccessful? If not then there shouldn't be an issue to remove.

It seems a bit overkill for a missionDefinition to return the lastSuccessfulRun and all its tasks and inspections. I think it makes sense to remove.

mrica-equinor (changes requested):
This bit of code needs to be changed before merging

image

backend/api/Services/MissionDefinitionService.cs
Eddasol, line 270: If removing the task include, the function ScheduleMissionRunFromMissionDefinitionLastSuccessfullRun needs to be updated so it doesn't fetch tasks like this anymore "missionDefinition.LastSuccessfulRun.Tasks"

mrica-equinor (approved):
LGTM :)

PR #2122, opened 2025-03-25
Rename TimesOfDay to reflect timezone

Eddasol:

### DIFF
--- a/backend/api/Database/Models/AutoScheduleFrequency.cs
+++ b/backend/api/Database/Models/AutoScheduleFrequency.cs
@@ -9,20 +9,20 @@ namespace Api.Database.Models
     public class AutoScheduleFrequency
     {
         [Required]
-        // In local time
-        public IList<TimeOnly> TimesOfDay { get; set; } = new List<TimeOnly>();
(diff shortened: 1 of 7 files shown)


backend/api/Database/Models/AutoScheduleFrequency.cs
```diff
@@ -9,20 +9,20 @@ namespace Api.Database.Models
     public class AutoScheduleFrequency
     {
         [Required]
-        // In local time
-        public IList<TimeOnly> TimesOfDay { get; set; } = new List<TimeOnly>();
+        // In Central European Time
+        public IList<TimeOnly> TimesOfDayCET { get; set; } = new List<TimeOnly>();
```
andchiind, line 13: Is this always CET and not CEST? If we include summer time then maybe we should also mention so in the comment above?

Eddasol: We are using Central European Time (which is equal to CEST in the summer), not Central European Summer Time

andchiind (approved):

PR #2200, opened 2025-05-15
Fix mission starting on pressure update

Eddasol:

### DIFF
--- a/backend/api/Database/Models/Robot.cs
+++ b/backend/api/Database/Models/Robot.cs
@@ -122,6 +122,12 @@ namespace Api.Database.Models
                 && Model.BatteryWarningThreshold > BatteryLevel
             )
                 return false;
+            if (
+                FlotillaStatus == RobotFlotillaStatus.Recharging
+                && Model.BatteryMissionStartThreshold != null
+                && Model.BatteryMissionStartThreshold > BatteryLevel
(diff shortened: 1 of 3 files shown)


backend/api/Database/Models/Robot.cs
```diff
@@ -122,6 +122,12 @@ public bool IsRobotReadyToStartMissions()
                 && Model.BatteryWarningThreshold > BatteryLevel
             )
                 return false;
+            if (
+                FlotillaStatus == RobotFlotillaStatus.Recharging
+                && Model.BatteryMissionStartThreshold != null
+                && Model.BatteryMissionStartThreshold > BatteryLevel
```
mrica-equinor, line 128: If you merge this code, then the robot will never start the second mission in the queue and we are back to where we were before, with missions fetting stuck in the queue

Eddasol: Note the adding of ```FlotillaStatus == RobotFlotillaStatus.Recharging```
That state should only occur if robot has returned to dock because of low battery

mrica-equinor: Ohh i see what you did there

mrica-equinor (approved):
LGTM :)

Issue #1987, opened 2025-01-28
Missions get stuck in ongoing


Eddasol:
**Describe the bug**
Sometimes a mission gets status ongoing, but its not actually running and its not possible to stop/skip the mission as isar isn't running that mission. When one mission gets stuck in ongoing it will occupy the ongoing spot belonging to the robot card, which hides the actual ongoing mission.

**To Reproduce**
* Queue a large amount of (short) missions
* Wait for them to run
* Once in a while one of the missions get stuck in ongoing

**Expected behavior**
Only the ongoing mission should have status ongoing. There should max be one ongoing mission per robot

**Screenshots**
Not that time spent is 235minutes. Other missions have been both started and completed while this mission stayed ongoing. 
<img width="861" alt="Image" src="https://github.com/user-attachments/assets/9d191e9b-fc9e-4854-b768-786614a70cb3" />

andchiind:
Has this been reproduced recently?

oysand:
Also happens when restarting ISAR during a mission #1943

Eddasol:
When this bug occurs one robot can end up with 2 ongoing missions. Seems that the mission that completed didn't get its state updated from ongoing to one of the completed states, even though it did complete. The mission isn't actually ongoing, but seems ongoing as it has an outdated state.

Eddasol:
<img width="830" alt="Image" src="https://github.com/user-attachments/assets/fee35f02-2a79-4d09-adb0-13febbc5aa46" />

PR #2203, opened 2025-05-15
Avoid rescheduling of missions stuck in ongoing

Eddasol:
Currently, if a mission gets stuck in ongoing (without actually being ongoing) it will be rescheduled if a recharging event occurs. This PR checks if a robot has more than one ongoing mission (which should never happen) and fail all missions that aren't the currentMission.

Mitigates some of the consequences of https://github.com/equinor/flotilla/issues/1987 but doesn't solve the issue


andchiind (approved):
We could also make the stop mission endpoint in ISAR return the ID of the mission it stopped, but this approach also works.

PR #2204, opened 2025-05-16
Remove unused returns, else-statements and brackets

Eddasol:

### DIFF
--- a/backend/api/EventHandlers/MissionEventHandler.cs
+++ b/backend/api/EventHandlers/MissionEventHandler.cs
@@ -76,7 +76,10 @@ namespace Api.EventHandlers
             }
             catch (MissionRunNotFoundException)
             {
-                return;
+                _logger.LogWarning(
+                    "Mission run not found for robot ID: {RobotId} when exceuting OnMissionRunCreated",
+                    missionRun.Robot.Id
+                );
             }
             finally
             {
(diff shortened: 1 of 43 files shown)


backend/api/EventHandlers/MissionEventHandler.cs
```diff
@@ -196,7 +196,6 @@ private async void OnSendRobotToDockTriggered(object? sender, RobotEmergencyEven
             catch (Exception ex)
             {
                 _logger.LogError(ex, "Failed to send robot {RobotId} to dock", robot.Id);
-                return;
```
andchiind, line 199: This could be nice to keep in case someone adds more code below later, but it's not too important

backend/api/EventHandlers/MissionEventHandler.cs
```diff
@@ -242,10 +241,7 @@ RobotEmergencyEventArgs e
             {
                 await MissionScheduling.StartNextMissionRunIfSystemIsAvailable(robot);
             }
-            catch (MissionRunNotFoundException)
-            {
-                return;
-            }
+            catch (MissionRunNotFoundException) { }
             finally
```
andchiind, line 251: The same is true here

andchiind (approved):
LGTM, nice cleanup

backend/api/EventHandlers/MissionEventHandler.cs
Eddasol: Can reintroduce this one if that is preferred. Don't have a strong opinion on this specific one

PR #2237, opened 2025-06-10
Require inspection area for mission definitions

Eddasol:


backend/api/Controllers/Models/MissionDefinitionResponse.cs
```diff
@@ -6,37 +6,26 @@ namespace Api.Controllers.Models
...
-        [JsonPropertyName("autoScheduleFrequency")]
         public AutoScheduleFrequency? AutoScheduleFrequency { get; set; }
 
-        [JsonPropertyName("lastSuccessfulRun")]
         public virtual MissionRun? LastSuccessfulRun { get; set; }
 
-        [JsonPropertyName("inspectionArea")]
         public InspectionAreaResponse? InspectionArea { get; set; }
 
-        [JsonPropertyName("isDeprecated")]
```
andchiind, line 33: I'm a bit confused by this change. Is the reasoning that the specified names were equivalent to what was anyways generated based on the JSON encoder config?

Eddasol: This is something we use on messages from isar, sara and Echo. MissionDefinitionResponse is the only other place the JsonPropertyName is used. Removing it cause it is unnecessary, and to increase inconsistency in how we implement models

backend/api/Database/Models/MissionDefinition.cs
```diff
@@ -30,7 +30,7 @@ public class MissionDefinition : SortableRecord
 
         public virtual MissionRun? LastSuccessfulRun { get; set; }
 
-        public InspectionArea? InspectionArea { get; set; }
+        public InspectionArea InspectionArea { get; set; }
```
Christdej, line 33: Why not [Required] if it is required?

backend/api/Controllers/Models/MissionDefinitionResponse.cs
Eddasol: We do use it in the frontend buts it not necessary to specify as Json if the json matches the existing name. It will be the same as what we do with MissionRunResponse

Christdej (approved):
LGTM, might cause problems if the insepctionArea is empty in database when running migrations

PR #2243, opened 2025-06-12
Reuse missionRun when interupted and resumed

Eddasol:

### DIFF
--- a/backend/api/Services/MissionSchedulingService.cs
+++ b/backend/api/Services/MissionSchedulingService.cs
@@ -14,7 +14,7 @@ namespace Api.Services
 
         public Task FreezeMissionRunQueueForRobot(string robotId);
 
-        public Task StopCurrentMissionRun(string robotId, string? stopReason = null);
+        public Task MoveCurrentMissionRunBackToQueue(string robotId, string? stopReason = null);
 
         public Task AbortAllScheduledNormalMissions(string robotId, string? abortReason = null);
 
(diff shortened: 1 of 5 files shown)


backend/api/Services/MissionSchedulingService.cs
```diff
@@ -463,56 +466,35 @@ IEnumerable<string> interruptedMissionRunIds
...
+                    "Mission run created",
+                    missionRun?.InspectionArea?.Installation,
+                    missionRun != null ? new MissionRunResponse(missionRun) : null
+                );
             }
         }
 
         private async Task StartMissionRun(MissionRun queuedMissionRun, Robot robot)
         {
+            // Reset status reason in case this is a restarted mission run
```
oysand, line 488: Could it make sense to have this be a separate function to describe the behavior enough so that a comment isn't necessary?

Eddasol: When entering the function we don't know if its a restarted mission or not. One option is to create a IsRestartedMissionRun that checks if any of the tasks are not notStarted and then reset status reason

Eddasol:
> Will the IsarMissionId still be consistent if we restart a paused mission? Just checking

It will be overwritten with the new isarMissionId. As this variable is only used to know which status messages from ISAR belongs to which Flotilla mission, the "old" isarMissionId should no longer be of interest as it no longer runs

PR #2260, opened 2025-06-24
Receive analysis result on broker

### LINKED ISSUE #2258: Subscribe to sara/clo_result_available
When the constant level oiler workflow is done, SARA will publish a message about inspection_id, oil_level and path to visualized image. Have Flotilla subscribe to this message (and maybe visualize the visualized image somehow as well)

N/A, we already have communication between the SARA and Flotilla

andchiind:
Closes #2258 

### DIFF
--- a/backend/api/EventHandlers/MqttEventHandler.cs
+++ b/backend/api/EventHandlers/MqttEventHandler.cs
@@ -806,6 +806,8 @@ namespace Api.EventHandlers
 
             inspection.AnalysisResult = analysisResult;
 
+            await InspectionService.UpdateInspectionAnalysisResults(inspection.Id, analysisResult);
+
             _ = SignalRService.SendMessageAsync(
                 "Analysis Result Ready",
                 installation,
--- a/backend/api/Database/Context/InitDb.cs
+++ b/backend/api/Database/Context/InitDb.cs
@@ -529,6 +529,7 @@ namespace Api.Database.Context
             {
                 task.Inspection = inspections[0];
             }
+            missionRuns[0].Tasks = tasks;
             context.AddRange(tasks);
             context.AddRange(missionDefinitions);
             context.AddRange(missionRuns);
--- a/backend/api/Services/InspectionService.cs
+++ b/backend/api/Services/InspectionService.cs
@@ -2,7 +2,6 @@
 using System.Globalization;
 using System.Net;
 using System.Text.Json;
-using Api.Controllers.Models;
 using Api.Database.Context;
 using Api.Database.Models;
 using Api.Services.Models;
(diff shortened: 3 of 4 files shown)


backend/api/Database/Models/AnalysisResult.cs
```diff
@@ -0,0 +1,42 @@
...
+        public string? Unit { get; set; }
+
+        [JsonPropertyName("class")]
+        public string? Class { get; set; }
+
+        [JsonPropertyName("confidence")]
+        public float? Confidence { get; set; }
+
+        [JsonPropertyName("warning")]
+        public string? Warning { get; set; }
```
Eddasol, line 31: Is it necessary to have a string. Boolean would be more intuitive to understand and error message could be incorporated into DisplayText

backend/api/Database/Models/AnalysisResult.cs
```diff
@@ -0,0 +1,42 @@
+using System.ComponentModel.DataAnnotations;
+using System.Text.Json.Serialization;
+
+namespace Api.Database.Models
+{
+    public class AnalysisResult
+    {
+        [JsonPropertyName("inspectionId")]
```
Eddasol, line 8: JsonProperty shouldn't be necessary in AnalysisResult

andchiind: It is necessary if we want to avoid making another DTO. Since it is already in the same format we want it in the frontend we can return it directly in SignalR. If we don't have this JsonPropertyName we can't be sure that it capitalizes words correctly.

backend/api/MQTT/MessageModels/SaraAnalysisResult.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Api.Mqtt.MessageModels
+{
+#nullable disable
+    public class SaraAnalysisResultMessage : MqttMessage
+    {
+        [JsonPropertyName("inspection_id")]
+        public string InspectionId { get; set; }
+
+        [JsonPropertyName("analysisType")]
```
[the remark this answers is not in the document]
andchiind: To make it consistent with models in SARA

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -756,5 +758,61 @@ private async void OnSaraInspectionResultUpdate(object? sender, MqttReceivedArgs
                 inspectionResultMessage
             );
         }
+
+        private async void OnSaraAnalysisResultMessage(object? sender, MqttReceivedArgs mqttArgs)
+        {
+            var saraAnalysisResult = (SaraAnalysisResultMessage)mqttArgs.Message;
+
+            var analysisResult = new AnalysisResult
```
mrica-equinor, line 766: SaraAnalysisResultMessage and AnalysisResult are the same, do we need both?

mrica-equinor (approved):
LGTM :) just look at the comment i made

backend/api/EventHandlers/MqttEventHandler.cs
andchiind: I think so since we might want to change one but not the other in the future without doing another migration. I think they also have different JSON encodings

PR #2382, opened 2025-09-04
Don't auto-schedule already scheduled missions

### LINKED ISSUE #2322: Dont have several of "the same mission" in queue
Image

It should not be several of the same mission in the que. As mission are queued up while robot is down, it can be added several of the same mission to the que.

In Dev we also have hundreds of pending missions, that should perhaps be cleaned up automatically.

Suggested fix:
For auto missions: Don't schedule a mission again if it is already in the queue.
For a user adding a mission: Give a notice (pop-up) to the user with the option of "scheduling anyways" if this is intended.

andchiind:
Closes #2322 

### DIFF
--- a/backend/api/Services/AutoScheduleService.cs
+++ b/backend/api/Services/AutoScheduleService.cs
@@ -226,6 +226,10 @@ namespace Api.Services
                 return;
             }
 
+            var pendingMissionRuns = await missionRunService.ReadMissionRunQueue(robot.Id);
+            if (pendingMissionRuns.Any((m) => m.MissionId == missionDefinition.Id))
+                return; // No need to schedule already scheduled mission
+
             logger.LogInformation(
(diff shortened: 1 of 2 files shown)


backend/api/Services/AutoScheduleService.cs
```diff
@@ -226,6 +226,10 @@ public async Task AutoScheduleMissionRun(string missionDefinitionId, TimeOnly ti
                 return;
             }
 
+            var pendingMissionRuns = await missionRunService.ReadMissionRunQueue(robot.Id);
+            if (pendingMissionRuns.Any((m) => m.MissionId == missionDefinition.Id))
+                return; // No need to schedule already scheduled mission
```
Eddasol, line 231: Add logging before return so its easier to understand behavior

andchiind: Done 👍

Eddasol (approved):
Note comment before merging :)

PR #2385, opened 2025-09-09
Stop fetching installations from echo

andchiind:


frontend/src/components/Contexts/RobotContext.tsx
```diff
@@ -22,10 +23,20 @@ interface Props {
 
 interface IRobotContext {
     enabledRobots: Robot[]
+    installationCode: string
+    installationName: string
+    installationInspectionAreas: InspectionArea[]
+    activeInstallations: Installation[]
+    switchInstallation: (selectedName: string) => void
```
mrica-equinor, line 30: I am a bit unsure of having these in the robot context, just because they are not related to the robot... they are related to the instalation the robot is on. Why not have the robot keep the instalation code and then we split the other ones to the instalation context which only cares about instalations for enabled robots? What was the reasoing here?

andchiind: Because this avoids a circular dependency. The robot context needs to know what the selected installation is, but the installation context would not know which installations are available without first knowing which robots are active. I could split it into a different context which uses the installations that this new big context fetches from the backend, but then the new context would have no database interaction and would be a bit pointless in my opinion. I could rename the robot context to something else now though if that would be more precise?

mrica-equinor: I see, what would you rename it to tho? RobotInstallationContext? I guess it is more explicit about what the context has and then it wouldnt be so weird why we have inspection areas in the robot context

andchiind: Maybe just "AssetContext"? Or just keep it robotcontext since we only care about installations within the context of robots?

mrica-equinor: AssetContext sounds good, then its like an overview. Still dont think robotcontext will be very clear 😅

PR #2506, opened 2025-12-03
Store mission filter in URL

andchiind:
image

image

There is a pre-existing bug where if you are on the front-page and navigate to the mission history page then the path gets overwritten, but the solution I have added still works. You can also see that it works when you refresh the page if you use the /history path. I think the front page URL issue already has a separate issue for it.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -4,7 +4,6 @@ import { AssetSelectionPage } from 'pages/AssetSelectionPage/AssetSelectionPage'
 import { FlotillaSite } from 'pages/FlotillaSite'
 import { LanguageProvider } from 'components/Contexts/LanguageContext'
 import { MissionControlProvider } from 'components/Contexts/MissionControlContext'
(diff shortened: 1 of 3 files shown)


frontend/src/components/NavigationMenu/NavigationMenuPages.tsx
```diff
@@ -33,7 +34,9 @@ export const PredefinedMissionsPage = () => {
 export const MissionHistoryPage = () => {
     return (
         <DefaultPage pageName="history">
-            <MissionHistoryView refreshInterval={1000} />
+            <MissionFilterProvider>
+                <MissionHistoryView refreshInterval={1000} />
+            </MissionFilterProvider>
```
mrica-equinor: Could probs move MissionFilterProvider inside MissionHistoryView and then its always there instead of importing it in this file and the frontpage file

andchiind: Good point. I'll do that

PR #2515, opened 2025-12-10
Don't hardcode mission status

### LINKED ISSUE #2510: Isar can accept two mission if slow stopping
**Describe the bug**
If adding two missions to queue while robot is returning home, it will accept both if stopping is a bit slow. Should only accept one. Result is that flotilla gets 2 ongoing missions but only one of them will run and complete. The second one will get stuck in ongoing.

**To Reproduce**
1. Add a sleep at the beginning of stop function in isar-robot
2. Add two missions in queue while state is returning home

**Expected behavior**
Should give conflict for one of the missions. Should one accept one

**Screenshots**
Resulting logs
Image

andchiind:
Closes #2510 

Wait for confirmation from ISAR that the mission has started. The main "breaking change" is that missions won't show as in progress until we get a mission status event from ISAR. To accomodate this we now treat NotStarted differently from not having received any status from ISAR. This goes in hand with this ISAR PR https://github.com/equinor/isar/pull/1034.

### DIFF
--- a/backend/api/EventHandlers/MqttEventHandler.cs
+++ b/backend/api/EventHandlers/MqttEventHandler.cs
@@ -424,7 +424,8 @@ namespace Api.EventHandlers
 
             var flotillaMissionRun = await MissionRunService.ReadById(
                 isarMission.MissionId,
-                readOnly: true
+                readOnly: true,
+                includeDeprecated: true
             );
             if (flotillaMissionRun is null)
             {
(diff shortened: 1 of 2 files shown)


backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -710,31 +710,6 @@ private async void OnIsarStartup(object? sender, MqttReceivedArgs mqttArgs)
                 robot.Name,
                 robot.IsarId
             );
-            try
-            {
-                await MissionScheduling.MoveCurrentMissionRunBackToQueue(robot.Id);
```
Eddasol: Do we have an alternative way to handle "re-scheduling" of missions when isar restarts?

andchiind: No, the mission would be considered failed, and ISAR should stop the mission. If we want to reschedule it, then ISAR can send a "mission aborted" MQTT message.

backend/api/Services/MissionSchedulingService.cs
```diff
@@ -272,6 +253,11 @@ await missionRunService.UpdateMissionRunProperty(
                 logger.LogWarning("{Message}", $"No mission was running for robot {robot.Id}");
             }
 
+            var ongoingMissionRunIds = ongoingMissionRuns
+                .Where(missionRun => missionRun.Id == robot.CurrentMissionId)
+                .Select(missionRun => missionRun.Id)
+                .ToList();
+
```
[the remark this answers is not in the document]
andchiind: In that situation it would be updated to not failed once it starts again, but yeah that might be a good idea anyways.

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -706,35 +705,10 @@ private async void OnIsarStartup(object? sender, MqttReceivedArgs mqttArgs)
             }
 
             _logger.LogInformation(
-                "Received ISAR restart event for robot {robotName} with ISAR id {isarId}. Will restart ongoing missionRun, if any, by moving it back to the queue.",
+                "Received ISAR restart event for robot {robotName} with ISAR id {isarId}.",
```
Eddasol: The entire startUpMessage can probebly removed as it doesn't do anything anymore. Suggest making a new issue to handle this

PR #2530, opened 2025-12-17
Only reschedule failed tasks when aborting

andchiind:
This won't work consistently without this PR: https://github.com/equinor/isar/pull/1044


andchiind:
> I cannot get this to work while testing locally. The mission just becomes cancelled when the battery runs low, and the remaining tasks dont run. I also don't see the usefulness to retry failed tasks whenever the battery runs low?

Currently the whole mission gets rescheduled even though part of the mission completed before being aborted. This avoids rerunning the same tasks over again.

PR #2612, opened 2026-02-27
Round off co2 percentage to one decimal

Eddasol:
- [ ] A test has been written


frontend/src/components/Contexts/InspectionsContext.tsx
```diff
@@ -5,20 +5,30 @@ import { useQuery } from '@tanstack/react-query'
 import { queryClient } from '../../App'
 import { useBackendApi } from 'api/UseBackendApi'
 
+interface IImageData {
+    data: string | undefined
```
andchiind: This is a byte string, not a string, isn't it? Here is the output on the backend 

Image

Eddasol: Checked a bit and it seems as if data its the ObjectURL which is a string

PR #2661, opened 2026-04-13
Fix old missions in queue on isar restart

### LINKED ISSUE #2660: CurrentMissionId doesn't become null after mission
**Describe the bug**
When a mission completes the CurrentMissionId gets set to null by calling `await RobotService.UpdateCurrentMissionId(robot.Id, null);` on line 553 in MqttEventHandler.cs.

When I then call the "/robots/{id}" endpoint the currentMissionId is not null.

**To Reproduce**
1. Run mission
2. Wait for mission to complete
3. Call "/robots/{id}" endpoint and note currentMissionId

**Expected behavior**
CurrentMissionId should be null after a mission

**Screenshots**

Eddasol:
- [ ] A test has been written

### DIFF
--- a/backend/api/EventHandlers/MqttEventHandler.cs
+++ b/backend/api/EventHandlers/MqttEventHandler.cs
@@ -391,6 +391,7 @@ namespace Api.EventHandlers
                     isarAbortedMission.Reason
                 );
             }
+            catch (NoUnfinishedTasksInMissionException) { }
             catch (RobotNotFoundException)
             {
                 _logger.LogWarning(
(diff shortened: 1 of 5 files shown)


backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -550,6 +552,7 @@ private async void OnIsarMissionUpdate(IsarMissionMessage isarMission)
             if (robot.CurrentMissionId == flotillaMissionRun.Id)
             {
                 await RobotService.UpdateCurrentMissionId(robot.Id, null);
+                robot.CurrentMissionId = null;
```
andchiind, line 555: Is this change being committed to the database anywhere?

Eddasol: Its being comited to the database in the function on the line above. This is only to ensure that the value is updated incase the its used more within this function

nicholasdalhaug (approved):

PR #2678, opened 2026-04-21
Handle robot recharging with isar mission

andchiind:
See https://github.com/equinor/isar/pull/1120

image

- [ ] A test has been written


frontend/src/components/Displays/RobotDisplays/RobotStatusIcon.tsx
```diff
@@ -49,6 +49,7 @@ export const RobotStatusChip = ({ status, itemSize }: StatusProps) => {
             iconColor = tokens.colors.text.static_icons__default.hex
             break
         }
+        case RobotStatus.GoingToRechargingWithMission:
```
oysand, line 52: This seem to be the status below, but with information about the mission status. Could it make sense to infer it from the current mission and its status? That way whether it has a mission or not can be determined by the Robot.CurrentMissionId?

oysand (approved):
LGTM

frontend/src/components/Displays/RobotDisplays/RobotStatusIcon.tsx
andchiind: With how it currently works we don't always want to display the current mission, but that could be done in a different refactor.

PR #2694, opened 2026-05-05
Replace echo missions with just using mission definitions

### LINKED ISSUE #2693: Store mission definition tasks in the source object directly instead of linking to echo
Currently the mission definitions mostly just link to echo missions. They basically just act as links to echo and we almost always fetch from echo to make sure the new missions we run are up to date with echo. The only exception is custom missions. I propose that we replace the custom mission string JSON and the link to echo with storing the tasks directly in the flotilla database, and remove the echo dependency in the mission planning. We can then just update mission definitions directly in order to update missions (something we have already been doing to some extent since we overwrite echo mission names already and append zoom levels).

We already have a mission planner ready, which simply needs to call flotilla endpoints instead of echo endpoints to work. The frontend for updating missions does not care what the source is https://github.com/equinor/pointcloud_processing/issues/666 .

The benefits of doing this is many.
- Firstly we remove the problem of constantly coordinating the information in echo mission planner with the mission info we store in flotilla. There would be one correct data model and we wouldn't need to translate between them
- Secondly we will be able to update the model with the information we need without needing to rely on a 3rd party
- Thirdly we will improve access control, as we can, for instance, ensure that only those that can run missions are allowed to create them. This is not a requirement and we can be more lenient, but currently we don't have this control
- Fourthly we will be able to coordinate the data models in Flotilla, ISAR and SARA by using OpenAPI, making maintaining data models easier
- Fifthly we will be able to set analysis type directly in the mission tasks, removing complicated logic and hacky solutions in SARA https://github.com/equinor/sara/issues/329
- Sixthly it is basically required in order to implement variable zoom values properly (https://github.com/equinor/flotilla/issues/2624 and https://github.com/equinor/pointilla_maps/issues/21)
- Seventhly we will be able to test the mission editor in a dev environment, or even locally. Currently we are only able to test in production environment, which is far from ideal (this should probably be the first one, but I can't be bothered to change the numbering now)
- Eightly we will be able to create integration tests involving the full mission pipeline. Previously we would either have to access a prod mission database environment in such a test, or simply mock it and hope the model wouldn't change later
- Ninethly we no longer need to maintain custom mission separately from echo missions. There would only be normal missions, which does not care how the mission was first edited
- Tenthly the echo mission planner has a lot of strange quirks which are hard to work around. This change will simplify code in the pointilla mission planner backend

I see no reason to put this off too long, since it will benefit a lot of other tasks and is not that big of a task at this point. This should not be an epic in my opinion, since it can be done in a single PR which is not too big. The only dependency is that we need to update the pointilla mission planner backend, but I developed this backend in such a way that it should be easy to switch out. Besides this the pointilla mission planner needs to request the flotilla role scope, but that is a single line change.

We will no longer store references to the echo mission planner database in Source.cs, which is linked in MissionDefinition.cs.

The tasks stored in the database can be more or less identical to what ISAR expects to receive. It could be more or less identical to the MissionTask owned table, except we shouldn't store inspection results there. Although this shouldn't be the case in MissionTask anyways in my opinion. It is probably fine to make the new model also be owned. In echo it wasn't but it's personal preference.

It will remove excess 3rd party dependencies which is positive in removing attack vectors.

It also generally gives us full control of access control to the input to Flotilla. Previously that was not the case.

andchiind:
Things to do after merging this PR:

- Update the IsarMissionDefinition.cs constructor for IsarTaskDefinition so that we send the analysistype to ISAR. At the same time ISAR should have a PR where we make use of this value
- Start migrating echo missions to mission definitions. The missions will not be available to run until we do this. The analysistype defaults to NoAnalysis, but until SARA makes use of this value that doesn't matter

- [x] A test has been written


backend/api.test/Controllers/InspectionAreaControllerTests.cs
```diff
@@ -10,6 +10,7 @@
 using Api.Services;
 using Api.Test.Database;
 using Microsoft.Extensions.DependencyInjection;
+using Org.BouncyCastle.Asn1.X509;
```
Christdej: What is this?

backend/api/Database/Models/MissionTaskDefinition.cs
```diff
@@ -0,0 +1,94 @@
...
+    public enum AnalysisType
+    {
+        Fencilla,
+        CLOE,
+        ThermalReading,
+        CO2,
+        NoAnalyis,
+    }
+
+    [Owned]
```
Christdej, line 16: Imo, dont have owned here, doesnt give much I think, but can make annotations like "Required" or lists hard to work with

backend/api.test/Controllers/InspectionAreaControllerTests.cs
andchiind: That is me missclicking with "cmd+."

backend/api/Database/Models/MissionTaskDefinition.cs
andchiind: Are you mainly worried that we won't be able to enforce the required values? Thankfully none of the values here are in lists, since I agree that that is annoying to work with.

Christdej: I am mainly worried that if we want to add required fields or lists in the future this will be more complicated than it needs to be because we added a [Owned]. Also when you add a [Owned] required list (like done on MissionDefinition), I believe there is no difference to not using [Owned], a new table will be created

andchiind: I agree, we can go with non-owned task definitions

backend/api/Controllers/Models/MissionDefinitionResponse.cs
```diff
@@ -1,85 +1,43 @@
...
                     : null;
-            InspectionArea = new InspectionAreaResponse(missionDefinition.InspectionArea);
             LastSuccessfulRun = missionDefinition.LastSuccessfulRun;
+            InspectionArea = new InspectionAreaResponse(missionDefinition.InspectionArea);
             IsDeprecated = missionDefinition.IsDeprecated;
-            SourceId = missionDefinition.Source.SourceId;
         }
     }
-
-    public class MissionDefinitionWithTasksResponse(
```
andchiind, line 53: No point differentiating any more, since we only made a model without tasks since it took a long time to fetch them.

backend/api/Controllers/Models/PlantInfo.cs
```diff
@@ -1,25 +0,0 @@
-﻿#nullable disable
-using Api.Database.Models;
-
-namespace Api.Controllers.Models
-{
-    public class PlantInfo
```
andchiind, line 6: This was unused

backend/api/Database/Models/MissionTask.cs
```diff
@@ -82,22 +48,17 @@ public MissionTask(MissionTask copy)
...
 
-        [MaxLength(200)]
-        public Uri? TagLink { get; set; }
-
         [Required]
         public Pose RobotPose { get; set; }
 
-        public int? PoseId { get; set; }
+        [Required]
+        public AnalysisType AnalysisType { get; set; }
```
Christdej: Possible that we will have a problem here with [required] as not all MissionTasks have a AnalysisType in the database

backend/api/Database/Models/TaskDefinition.cs
```diff
@@ -0,0 +1,80 @@
...
+#pragma warning disable CS8618
+namespace Api.Database.Models
+{
+    public enum AnalysisType
+    {
+        Fencilla,
+        CLOE,
+        ThermalReading,
+        CO2,
+        NoAnalyis,
```
Christdej:
```suggestion
        NoAnalysis,
```

backend/api/Database/Models/MissionDefinition.cs
```diff
@@ -11,7 +11,7 @@ public class MissionDefinition : SortableRecord
         public string Id { get; set; }
 
         [Required]
```
Christdej, line 15: Can remove [Required] here, does nothing I think, if we want to keep required, perhaps initialize list to an empty list

backend/api/Database/Context/FlotillaDbContext.cs
```diff
@@ -26,7 +26,6 @@ public FlotillaDbContext(DbContextOptions options)
         public DbSet<Installation> Installations => Set<Installation>();
         public DbSet<InspectionArea> InspectionAreas => Set<InspectionArea>();
         public DbSet<ExclusionArea> ExclusionAreas => Set<ExclusionArea>();
-        public DbSet<Source> Sources => Set<Source>();
         public DbSet<AccessRole> AccessRoles => Set<AccessRole>();
         public DbSet<UserInfo> UserInfos => Set<UserInfo>();
         public DbSet<TagInspectionMetadata> TagInspectionMetadata => Set<TagInspectionMetadata>();
```
Christdej, line 30: Add
public DbSet<TaskDefinition> TaskDefinitions => Set<TaskDefinition>();
to have the plural naming we have on the other DbSets

backend/api/Database/Models/TaskDefinition.cs
Christdej: Consider moving "NoAnalysis" to the top, for it to be the default value, then in constructors, if we forget to set AnalysisType, it will default to NoAnalysis and not Fencilla

backend/api/Controllers/MissionDefinitionController.cs
```diff
@@ -90,36 +93,133 @@ public async Task<
...
+            {
+                logger.LogError(e, "{ErrorMessage}", e.Message);
+                return BadRequest(e.Message);
+            }
+            catch (HttpRequestException e)
+            {
+                logger.LogError(e, "Error retrieving missions from Mission Loader");
+                return StatusCode(StatusCodes.Status502BadGateway);
+            }
+            catch (MissionLoaderUnavailableException e)
```
Christdej: Probably can no longer throw this exception type?

backend/api/Controllers/MissionDefinitionController.cs
```diff
@@ -90,36 +93,133 @@ public async Task<
...
+                    InspectionArea = inspectionAreaForMission,
+                };
+                await missionDefinitionService.Create(newMissionDefinition);
+                return Ok(newMissionDefinition);
+            }
+            catch (MultipleInspectionAreasFoundException e)
+            {
+                return BadRequest(e.Message);
+            }
+            catch (SourceException e)
```
Christdej: Probably no longer can throw this exception?

backend/api/Database/Models/MissionTask.cs
```diff
@@ -82,22 +48,17 @@ public MissionTask(MissionTask copy)
         [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
         public string Id { get; set; }
 
-        [Required]
-        public int TaskOrder { get; set; }
```
Christdej, line 86: Why have we removed TaskOrder? No the ordering of the tasks are up EF load order

backend/api/Controllers/Models/MissionDefinitionResponse.cs
```diff
@@ -1,85 +1,43 @@
...
-            Id = missionDefinition.Id;
-            Name = missionDefinition.Name;
-            InstallationCode = missionDefinition.InstallationCode;
-            Comment = missionDefinition.Comment;
-            InspectionFrequency = missionDefinition.InspectionFrequency;
+            Id = missionDefinition.Id ?? string.Empty;
+            Tasks = missionDefinition.Tasks;
+            Name = missionDefinition.Name ?? string.Empty;
+            InstallationCode = missionDefinition.InstallationCode ?? string.Empty;
+            Comment = missionDefinition.Comment ?? string.Empty;
```
Christdej: Does "?? string.Empty" do anything here? They might as well be null rather then ""?

frontend/src/models/MissionDefinition.ts
```diff
@@ -1,6 +1,34 @@
...
+    objectHeight: number
+}
+
+interface MissionTaskDefinition {
+    id: string
+    tagId: string
+    description?: string
+    robotPose: Pose
+    targetPosition: Position
+    isarZoomDescription?: ZoomDescription
```
Christdej:
```suggestion
    zoomDescription?: ZoomDescription
```

Christdej (changes requested):

backend/api/Database/Models/MissionDefinition.cs
andchiind: Are you sure it does nothing? I thought it was what determined the nullability in the migration?

backend/api/Database/Context/FlotillaDbContext.cs
andchiind: Deleted this in the latest version

backend/api/Controllers/MissionDefinitionController.cs
andchiind: Good point

backend/api/Database/Models/MissionTask.cs
andchiind: The aim was to have the ordering be set by the actual ordering of the list, but I will experiment and research whether this is guaranteed to be consistent

backend/api/Database/Models/MissionTask.cs
andchiind: Good point! Will remove it, then for now ISAR can interpret null values as no analysis

backend/api/Database/Models/MissionTask.cs
andchiind: Alas, you win

backend/api/Database/Models/MissionDefinition.cs
Christdej: Ignore my comment, I am not sure why I added this comment

frontend/src/components/Dialogs/MissionEditDialog.tsx
Christdej: Installation and navigate never used

frontend/src/models/Inspection.ts
```diff
@@ -1,14 +1,17 @@
...
     id: string
     isarInspectionId: string
     isCompleted: boolean
-    inspectionType: InspectionType
+    inspectionType: SensorType
     analysisResult?: AnalysisResult
     inspectionTarget: Position
     videoDuration?: number
     inspectionUrl?: string
+    analysisType: AnalysisType
```
Christdej: Still singular analysisType, while backend used list, problemo

frontend/src/models/MissionDefinition.ts
```diff
@@ -1,6 +1,34 @@
...
+import { Pose } from './Pose'
+import { Position } from './Position'
+import { SensorType } from './Inspection'
+
+export enum AnalysisType {
+    Fencilla = 'Fencilla',
+    CLOE = 'CLOE',
+    ThermalReading = 'ThermalReading',
+    CO2 = 'CO2',
+    NoAnalyis = 'NoAnalysis',
```
Christdej:
```suggestion
    NoAnalysis = 'NoAnalysis',
```

backend/api/Database/Models/TaskDefinition.cs
```diff
@@ -0,0 +1,77 @@
...
+            RobotPose = taskQuery.RobotPose;
+            ZoomDescription = taskQuery.ZoomDescription;
+            AnalysisTypes = taskQuery.AnalysisTypes;
+            SensorType = taskQuery.SensorType;
+            TargetPosition = taskQuery.TargetPosition;
+            VideoDuration = taskQuery.VideoDuration;
+        }
+
+        [Key]
+        public int Index { get; set; }
```
Christdej, line 34: TaskDefinition primary key is [Key] public int Index with UseIdentityByDefaultColumn, but application code explicitly sets Index on Create (index, 0-based) and on Update (index + 1, 1-based). This both (a) risks PK collisions across mission definitions because the PK is not composite with MissionDefinitionId and (b) is inconsistent: a freshly-created definition gets tasks 0..n while an updated one gets 1..n+1. Use a separate Order property for user-visible ordering and let Index be DB-generated, or make Index non-key and define a composite key (MissionDefinitionId, Order). — backend/api/Database/Models/TaskDefinition.cs:3542, backend/api/Controllers/MissionDefinitionController.cs:1678, backend/api/Controllers/MissionDefinitionController.cs:1752

backend/api/Services/MissionDefinitionService.cs
```diff
@@ -142,13 +135,6 @@ public async Task<List<MissionDefinition>> ReadByInspectionAreaId(
                 .ToListAsync();
         }
 
-        public async Task<MissionDefinition?> ReadBySourceId(string sourceId, bool readOnly = true)
-        {
-            return await GetMissionDefinitionsWithSubModels(readOnly: readOnly)
-                .Where(m => m.IsDeprecated == false)
-                .FirstOrDefaultAsync(m => m.Source.SourceId.Equals(sourceId));
-        }
```
Christdej, line 150: Have removed IsDeprecated filter, is this intentional? Do we want to show deprecated missions?

Christdej (changes requested):

backend/api/Services/MissionDefinitionService.cs
andchiind: We want to never show them, and they are currently filtered out. I am actually considering removing it altogether. It was originally intended to replace actual deletion, but since then people have re-added actual deletion and then IsDeprecated stopped being used. But good point.

backend/api/Database/Models/TaskDefinition.cs
andchiind: I have now made it a composite key with the parent ID. I don't really see (b) as an issue but I can change it if you want.

Christdej (approved):
LGTM

aeshub:
### Added: wire `analysis_types` through to ISAR

Two commits on top of this PR to finish the wiring from the new `AnalysisTypes` model into the ISAR start-mission contract:

- **`Add analysis_types to ISAR mission definition`** (4465eb7d) — adds the `analysis_types` field to the `IsarInspectionDefinition` JSON struct so the per-inspection analysis list is included in the body of the POST to ISAR's `/schedule/start-mission`. Flotilla enum values are mapped to the keys SARA understands (`Fencilla` → `"fencilla"`, `CLOE` → `"cloe"`, `ThermalReading` → `"thermal-reading"`, `CO2` → `"co2"`). Empty / null input serialises as `null`, which makes SARA fall back to its file-extension default. Pairs with equinor/isar#1133  and equinor/sara#363.

- **`Set AnalysisTypes on MissionTask in ToMissionRunTask`** (155a2d6f) — drive-by: `MissionTask.AnalysisTypes` was added in this PR but never populated by `TaskDefinition.ToMissionRunTask()` (only the nested `Inspection.AnalysisTypes` was set). That made the field silently empty for every scheduled mission and zeroed it out on the round-trip through `ToMissionTaskDefinition`. Now populated from the same source.

Two new unit tests in `IsarMissionDefinitionTests.cs` cover the snake_case wire format and the drive-by.

PR #2703, opened 2026-05-08
Add workflow to validate migrations against temp db

### LINKED ISSUE #2533: Add check to ensure that migration on PR isn't outdated
If a PR (B) with a migration is rebased on a commit (A) that also has migrations, the migrations (for B) needs to be regenerated. If not this will create problems for the next PR that includes migrations.

There should be a way to warn such that the chance of taking in outdated migrations are reduced.

One suggestion is to have a workflow that autogenerates the migrations for the PR and checks if it matches the migrations that is included in the PR. In some cases its necessary to manually change a migration so such a check should only be a warning and not a blocker.

Alternativ is to check if timestamp of included migration is newer that previous migration

nicholasdalhaug:


Eddasol (approved):
LGTM

PR #1137, opened 2023-11-08
Refactoring frontend

mrica-equinor:

PR #2741, opened 2026-05-27
Add acoustic measurement inspection type

aeshub:
## Summary

Adds `AcousticMeasurement` as a new inspection sensor type, plumbing acoustic measurement parameters (frequency range, SNR threshold, detection type) end-to-end from `MissionDefinition` through `TaskDefinition`, `Inspection`, and the ISAR payload.

- Adds `SensorType.AcousticMeasurement` and `RobotCapabilitiesEnum.take_acoustic_measurement`
- Adds `AcousticInspectionMetadata` owned model persisted on both `Inspection` and `TaskDefinition` (mirrors how `VideoDuration` is handled)
- Adds `IsarAcousticInspection` wire DTO (snake_case) emitted to ISAR only when present
- Exposes the metadata via the `TaskQuery` request model and corresponding frontend `Inspection` / `MissionTaskDefinition` interfaces
- Adds EF Core migration adding 4 nullable columns on `Inspections` and `TaskDefinition`
- Adds English and Norwegian translation keys so the Mission History filter dropdown displays a proper label

## Blocked on

- Requires [ISAR #1137](https://github.com/equinor/isar/pull/1137) to merge and a subsequent ISAR release before this can be deployed — ISAR must understand the `acoustic` field on the wire payload.

## Notes

- Form/parameter validation is intentionally left to `pointilla_maps` (the mission editor) and ISAR as the backstop, mirroring how `VideoDuration` is treated today.
- No frontend UI for editing/displaying acoustic parameters is included here; the TS models are added so consumers (pointilla_maps, task overview) can be extended in follow-ups.

Related: equinor/robotics-infrastructure#924

### DIFF
--- a/backend/api/Database/Models/Inspection.cs
+++ b/backend/api/Database/Models/Inspection.cs
@@ -1,6 +1,8 @@
 ﻿using System.ComponentModel.DataAnnotations;
 using System.ComponentModel.DataAnnotations.Schema;
+using System.Text.Json.Serialization;
 using Api.Services.Models;
+using Microsoft.EntityFrameworkCore;
 #pragma warning disable CS8618
 namespace Api.Database.Models
 {
@@ -18,7 +20,8 @@ namespace Api.Database.Models
             Position target,
             IList<AnalysisType> analysisTypes,
             float? videoDuration,
-            string? taskDescription = null
+            string? taskDescription = null,
+            AcousticInspectionMetadata? acousticInspectionMetadata = null
         )
         {
             InspectionType = sensorType;
@@ -26,6 +29,7 @@ namespace Api.Database.Models
             VideoDuration = videoDuration;
             AnalysisTypes = analysisTypes;
             TaskDescription = taskDescription;
+            AcousticInspectionMetadata = acousticInspectionMetadata;
         }
 
         // Creates a blank deepcopy of the provided inspection
@@ -39,6 +43,9 @@ namespace Api.Database.Models
             InspectionTarget = new Position(copy.InspectionTarget);
             TaskDescription = copy.TaskDescription;
             AnalysisTypes = copy.AnalysisTypes;
+            AcousticInspectionMetadata = copy.AcousticInspectionMetadata is not null
+                ? new AcousticInspectionMetadata(copy.AcousticInspectionMetadata)
+                : null;
         }
 
         [Key]
@@ -62,6 +69,8 @@ namespace Api.Database.Models
         public AnalysisResult AnalysisResult { get; set; }
 
         public float? VideoDuration { get; set; }
+
+        public AcousticInspectionMetadata? AcousticInspectionMetadata { get; set; }
 
         [MaxLength(250)]
         public string? InspectionUrl { get; set; }
@@ -90,6 +99,9 @@ namespace Api.Database.Models
                     RobotCapabilitiesEnum.take_co2_measurement
                 ),
                 SensorType.Audio => capabilities.Contains(RobotCapabilitiesEnum.record_audio),
+                SensorType.AcousticMeasurement => capabilities.Contains(
+                    RobotCapabilitiesEnum.take_acoustic_measurement
+                ),
                 _ => false,
             };
         }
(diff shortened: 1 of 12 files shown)


aeshub:
### Missing cross-field validation: `SensorType` vs `AcousticInspectionMetadata`

`TaskQuery` / `TaskDefinition` / `Inspection` currently allow any combination of `SensorType` and `AcousticInspectionMetadata`. Two states are reachable that shouldn't be:

**Case A — `SensorType: AcousticMeasurement` with `AcousticInspectionMetadata: null`**

- Persisted on `TaskDefinition` and `Inspections`.
- `IsarInspectionDefinition` ctor sets `Acoustic = null` and `[JsonIgnore(WhenWritingNull)]` omits the field, so ISAR receives `{ "type": "AcousticMeasurement" }` with no `acoustic` block.
- ISAR's `to_inspection_task` raises `ValueError("No acoustic parameters given to AcousticMeasurement inspection task")`. Mission run fails at start.
- The broken `TaskDefinition` row remains in the DB and fails identically on every re-run.

**Case B — non-acoustic `SensorType` with non-null `AcousticInspectionMetadata`**

- Orphan metadata is persisted on rows whose `SensorType` is e.g. `Image`.
- The wire payload contains an `acoustic` block regardless of the type. ISAR's spec for non-acoustic types has `needs_acoustic_params=False`, so the field is silently dropped (Pydantic `extra="ignore"`).
- No runtime failure, but the DB row holds contradictory data indefinitely. Worse than Case A from a debugging standpoint.

### Suggested fix

Add an `IValidatableObject` invariant on `TaskQuery` so misconfigured requests are rejected at the controller boundary (the `[ApiController]` attribute turns invalid `ModelState` into automatic 400 with `ValidationProblemDetails`):

```csharp
public class TaskQuery : IValidatableObject
{
    // ... existing props ...

    public IEnumerable<ValidationResult> Validate(ValidationContext _)
    {
        var isAcoustic = SensorType == SensorType.AcousticMeasurement;
        var hasAcousticMeta = AcousticInspectionMetadata is not null;

        if (isAcoustic && !hasAcousticMeta)
            yield return new ValidationResult(
                "AcousticInspectionMetadata is required when SensorType is AcousticMeasurement.",
                new[] { nameof(AcousticInspectionMetadata) });

        if (!isAcoustic && hasAcousticMeta)
            yield return new ValidationResult(
                "AcousticInspectionMetadata must be null unless SensorType is AcousticMeasurement.",
                new[] { nameof(AcousticInspectionMetadata) });
    }
}
```

Effect:
- Misconfigured requests get `400 Bad Request` with a field-targeted error message before anything is persisted.
- pointilla_maps gets immediate, actionable feedback at save time instead of an opaque "mission won't start" later.
- ~12 LOC, no schema or runtime changes.

If you want to be more conservative, dropping the second rule still catches Case A (the loud-but-late-binding failure) while leaving Case B untouched. My preference would be to enforce both, since there is no realistic legitimate use case for non-acoustic tasks carrying acoustic params, and silent data drops are harder to diagnose later than 400s.

backend/api/Migrations/20260608130010_AddAcousticInspectionMetadata.cs
Christdej, line 1: Remake migrations as they are older then the ones on main, can cause migration problemes down the line

backend/api/Database/Models/Inspection.cs
```diff
@@ -103,5 +115,64 @@ public enum SensorType
...
+        float frequencyTo,
+        float snrValueThreshold,
+        AcousticDetectionType detectionType
+    )
+    {
+        [Required]
+        public float FrequencyFrom { get; set; } = frequencyFrom;
+
+        [Required]
+        public float FrequencyTo { get; set; } = frequencyTo;
```
Christdej, line 143: Could add validation range here, so missions are rejected at creation rather then scheduling, as ISAR will fail if range is outside 0 - 100_000

Christdej (changes requested):
Squash some commits

Christdej (approved):
LGTM

PR #2759, opened 2026-06-02
Fix CLOE 1 month toggle to fetch 30 days of data

### LINKED ISSUE #2756: CLOE data view "1 month" toggle does not actually show one month of data
## Summary

On the CLOE data view (`/<installationCode>/cloe-view`), the time-range toggle offers **7 days** and **1 month**. The "1 month" option does not reliably show a month of data — in many cases it shows the same data as "7 days", or otherwise an arbitrary subset that has nothing to do with the selected time range.

## Where to see it

`frontend/src/pages/DataViewPage/CloeDataViewPage.tsx` — the page rendered at route `cloe-view` (registered in `frontend/src/pages/FlotillaSite.tsx`). The "Measured oil level" line plot at the bottom is the affected component.

## Steps to reproduce

1. Open the CLOE data view for an installation that has CLOE / "Avlesning" missions running roughly daily or more often (so there are clearly more than 10 successful runs in the last month).
2. Observe the "Measured oil level" plot under "7 days".
3. Switch to "1 month".
4. Open the browser network tab.

## Expected

- Switching to "1 month" shows roughly 30 days of measurements per tag.
- Switching the toggle triggers a request that fetches the data for the selected window (or otherwise guarantees that data covering the window is available on the client).

## Actual

- "1 month" shows the same set of points as "7 days" whenever the robot has run the "Avlesning" mission frequently enough.
- Toggling between "7 days" and "1 month" fires no new network request; the plot is recomputed purely from data already on the client.
- For installations with fewer runs the two views may differ, but the "1 month" view still cannot show more than a fixed number of missions, regardless of how far back they reach.

## Root cause

The CLOE page fetches mission runs with `backendApi.getMissionRuns({...})` (`CloeDataViewPage.tsx:138-153`) and passes only:

- `installationCode`
- `orderBy: 'EndTime desc, Name'`
- `statuses: [Successful, PartiallySuccessful]`
- `nameSearch: 'Avlesning'`

It does **not** pass any time-range parameter (`MinEndTime` / `MaxEndTime`) and does **not** pass `pageSize`.

The backend endpoint `GET missions/runs` (`backend/api/Controllers/MissionRunController.cs:31`) is paginated with a default page size of **10** (`backend/api/Controllers/Models/QueryStringParameters.cs:13`). So the request returns at most the 10 most recent successful "Avlesning" runs for the installation, full stop.

The "1 month" / "7 days" toggle is then applied entirely client-side (`CloeDataViewPage.tsx:184-194`) by filtering those at-most-10 missions against a `Date.now() - days * 86_400_000` cutoff. Consequences:

- If the robot runs "Avlesning" more often than ~once every 3 days, all 10 returned runs already fall inside the last 7 days, and "1 month" looks identical to "7 days".
- The user can never see more than the last 10 mission runs in this view, regardless of the selected time range.
- Switching the toggle does not refetch (`timeRange` is not in the `useEffect` dependency list at `CloeDataViewPage.tsx:165-169`), so the user sees no loading state and no new request — the UI silently lies about what window it is showing.

## Impact

The "1 month" trend view on the CLOE page is effectively non-functional for any installation with regular CLOE missions. Users cannot see longer-term oil-level trends, which is one of the main reasons this view exists.

## Notes

- Backend already supports the necessary query parameters (`MinEndTime`, `MaxEndTime`, `PageSize`) — see `MissionRunQueryStringParameters` and the path-building in `frontend/src/api/BackendApi.tsx:41-80`. The frontend simply does not use them on this page.
- `MaxPageSize` on the backend is 1000 (`QueryStringParameters.cs:6`).

aeshub:
Fixes #2756.

Pass `MinEndTime` (30 days ago, Unix epoch seconds) and `PageSize=200` to `GET missions/runs` from the CLOE data view, so the page actually has 30 days of "Avlesning" runs available. The existing client-side 7d/1mo toggle now narrows real 30-day data, instead of filtering at-most-10 records.

- Toggle is still instant (no extra request on switch).
- Magic number constants extracted: `CLOE_FETCH_WINDOW_DAYS`, `CLOE_PAGE_SIZE`, `MS_PER_SECOND`, `MS_PER_DAY`.

### DIFF
--- a/frontend/src/pages/DataViewPage/CloeDataViewPage.tsx
+++ b/frontend/src/pages/DataViewPage/CloeDataViewPage.tsx
@@ -39,6 +39,11 @@ enum TimeRange {
     OneMonth = '1month',
 }
 
+const CLOE_FETCH_WINDOW_DAYS = 30
+const CLOE_PAGE_SIZE = 200
+const MS_PER_SECOND = 1000
+const MS_PER_DAY = 24 * 60 * 60 * MS_PER_SECOND
+
 enum CloeMissionNames {
     Avlesning = 'Avlesning',
 }
@@ -141,12 +146,15 @@ export const CloeDataViewPage = () => {
     const backendApi = useBackendApi()
 
     const fetchMissions = (): Promise<Mission[]> => {
+        const minEndTime = Math.floor((Date.now() - CLOE_FETCH_WINDOW_DAYS * MS_PER_DAY) / MS_PER_SECOND)
         return backendApi
             .getMissionRuns({
                 installationCode: installation.installationCode,
                 orderBy: 'EndTime desc, Name',
                 statuses: [MissionStatus.Successful, MissionStatus.PartiallySuccessful],
                 nameSearch: CloeMissionNames.Avlesning,
+                minEndTime: minEndTime,
+                pageSize: CLOE_PAGE_SIZE,
             })
             .then((missionRuns) => {
                 const missions = missionRuns.content


oysand (approved):
LGTM

PR #2770, opened 2026-06-04
Add option to connect to blob storage via connection string

olaals:
- [ ] A test has been written

Needed to connect to a mocked blob storage locally (azurite container) locally. Else it is a non-breaking change as it will be default continue to use DefaultAzureCredential as it did before.


Christdej (approved):
LGTM

PR #2781, opened 2026-06-11
Fix task markers on map

Eddasol:
- [ ] A test has been written

Prior to this PR the current task always had the task number corresponding to the last task in the map. This is since orderTasksByDrawOrder is ordered by draw order and not task order.

This PR reverts some of the functionality in this commit. This means that the order used in missionPage assumes that taskOrder is correct and 1 indexed.
https://github.com/equinor/flotilla/commit/0a23a2eb08d697a5bfdb3c505310c3a8cef17b54#diff-63bdc92977a2ff8ed9dc6d30832df46f5445c97aaca4e5f01efe5d19e2479986

### DIFF
--- a/frontend/src/models/Task.ts
+++ b/frontend/src/models/Task.ts
@@ -4,6 +4,7 @@ import { Pose } from './Pose'
 
 export interface Task {
     id: string
+    taskOrder: number
     tagId?: string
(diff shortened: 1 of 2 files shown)


marianards (approved):
LGTM

PR #2784, opened 2026-06-12
Add option for disabling tls mqtt for local dev

### LINKED ISSUE #2750: Add env var for disabling tls on mqtt (local development)
Make it possible to disable tls on mqtt locally. Makes setting up local development flows easier by not managing certs etc locally. 

The env var should make it so that it defaults to requiring TLS if not set. It must explicitly be disabled. 

For example: 

MQTT_ALLOW_INSECURE_LOCAL_CONNECTIONS=true

Make it easier to setup local development flows.

There is a higher chance of misconfiguration, but with a good env var name that displays that it should not be set in any cloud env, and making sure the default behaviour is to require TLS it is sufficiently mitigated.

Our mqtt is not exposed to public internet meaning that the risk if it was misconfigured is not an immediate attack vector.

olaals:
- [ ] A test has been written

Makes it possible to opt-in to not use TLS with mqtt via env vars (defaults to requiring TLS). 

Useful for local dev as it makes the setup easier.

### DIFF
--- a/backend/api/MQTT/MqttService.cs
+++ b/backend/api/MQTT/MqttService.cs
@@ -59,17 +59,24 @@ namespace Api.Mqtt
             _maxRetryAttempts = mqttConfig.GetValue<int>("MaxRetryAttempts");
             _shouldFailOnMaxRetries = mqttConfig.GetValue<bool>("ShouldFailOnMaxRetries");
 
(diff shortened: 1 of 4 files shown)


Christdej (approved):
LGTM

PR #2809, opened 2026-06-26
Let frontend interact with SARA directly

### LINKED ISSUE #2569: Have the frontend making the API call to sara directly
Currently the frontend makes a call to the backend to retrieve the images from the sara anon storage account and visualized account. Then this image is downloaded by the backend and fowarded as a file to the frontend where it is cached.

It has been discussed that an improvement to this code would be to have the frontend directly call the sara api and the download of the image happens on the frontend. 

This task can be split in two as well:

1. Have the download of the blob happen in the frontend
2. Have the frontend call sara instead of calling the backend which calls sara

Functionality is the same, it's just a better code approach.

andchiind:
To be merged with this PR: https://github.com/equinor/sara/pull/421

This PR lets the frontend get its inspection data from SARA directly, removing the need to intermediate data models and storage in Flotilla backend. It is also a security improvement, as SARA is able to authenticate users with their own credentials, as opposed to relying on app-to-app auth.

Watch out that there are two new frontend environment variables that need to be set when testing locally. I have already made a PR for these in the infrastructure repo.

- [ ] A test has been written


frontend/src/api/SaraApi.tsx
```diff
@@ -0,0 +1,42 @@
...
+            .GET<any>(path)
+            .then((response) => {
+                if (!response.content.analyses || response.content.analyses.length < 1) throw Error('No analysis found')
+                return response.content.analyses[response.content.analyses.length - 1].visualizedSAS
+            })
+            .catch(handleError('GET', path))
+        return content
+    }
+
+    async getRawDataUrl(inspectionId: string): Promise<string> {
```
Christdej: Is this used anywhere?

andchiind: Removed it now

Christdej (approved):

PR #2811, opened 2026-06-29
Add video and acoustic inspection playback

### LINKED ISSUE #2807: Video inspection results cannot be viewed in the mission run UI
## Problem

Flotilla's mission run pages let users review inspection results. Today, image and CO2 (gas) measurement results render and can be inspected. However, inspection results that are **videos** — produced by the `Video`, `ThermalVideo`, and `AcousticMeasurement` sensor types — cannot be viewed or played anywhere in the UI.

As more robots and inspection types produce video output (including acoustic leak detection, which records an MP4 clip), an operator or engineer reviewing a completed mission has no way to access this data. The video result is missing, broken, or shown as an unusable placeholder.

## Current behavior

- **Image** inspection results: viewable in the results gallery and the large viewer dialog.
- **Video** inspection results (`Video`, `ThermalVideo`, `AcousticMeasurement`): not viewable or playable.
- `AcousticMeasurement` results additionally carry metadata (e.g. frequency range, SNR threshold, detection type) that is never surfaced to the user.

## Expected behavior

- A user reviewing a mission run can identify which results are videos and play them, with standard playback controls (play/pause, seek, volume, fullscreen).
- Video results appear in the results gallery with a clear "this is a video" affordance.
- For `AcousticMeasurement` results, the associated acoustic metadata is shown alongside the video.
- Opening and browsing a mission page with video results stays responsive — page load should not noticeably degrade.

## Scope

- Applies to the inspection-visualization review flow (not the separate analysis flow).
- `Audio`-only results remain out of scope.

## Acceptance criteria

- [ ] `Video` and `ThermalVideo` results can be played from a mission run page.
- [ ] `AcousticMeasurement` results can be played and show their associated metadata.
- [ ] Video results are visually distinguishable from image results in the gallery.
- [ ] Loading a mission page that contains video results stays responsive.

aeshub:
Closes #2807.

Lets users view MP4 video inspection results (Video, ThermalVideo, AcousticMeasurement) in the mission-run UI, alongside the existing image/CO2 visualisations.

- Frontend: add `react-player` and a video player in the inspection report; small cards show a static placeholder, clips are fetched only when the large viewer opens.
- Acoustic measurements show metadata (frequency range, SNR threshold, detection type) in the dialog footer.
- Backend: resolve media content type from stored blob type, then extension, then sensor type; serve it on the inspection media endpoint (HTTP route unchanged).
- Rename inspection `image`→`media` internally (method/action/logs); analysis path untouched.
- Add unit tests for the content-type resolver.

Verified: frontend prettier/lint/build/ts-unused-exports clean; backend builds with -warnaserror, CSharpier clean, resolver tests 7/7. Full dotnet test (Testcontainers) and integration tests not run locally.


frontend/src/pages/InspectionReportPage/InspectionVideoPlayer.tsx
```diff
@@ -0,0 +1,55 @@
...
+            <Typography color={tokens.colors.text.static_icons__tertiary.hex}>{TranslateText('Video')}</Typography>
+        </StyledVideoPlaceholder>
+    )
+}
+
+export const VideoPlayer = ({ src }: { src: string }) => (
+    <StyledVideoWrapper>
+        <StyledVideoPlayer src={src} controls playsInline />
+    </StyledVideoWrapper>
+)
```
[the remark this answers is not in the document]
aeshub: react-player v3 renamed `url` to `src` (we're on 3.4.0 — `PlayerProps.src`, no `url`). Playback is confirmed working, so keeping `src`.

frontend/src/pages/InspectionReportPage/InspectionReportImage.tsx
```diff
@@ -116,14 +117,27 @@ const InspectionValueWithPlaceholder = ({ task, isLargeImage }: { task: Task; is
...
 
+const InspectionVideoWithPlaceholder = ({ task, isLargeImage }: { task: Task; isLargeImage: boolean }) => {
+    const { useImageData } = useInspectionsContext()
+    const { data, isPending, isError } = useImageData(task.inspection.isarInspectionId)
+    if (isError || !data) {
+        return <TextAsImage isLargeImage={isLargeImage} text={'No inspection could be found'} />
+    } else if (isPending) {
+        return <PendingResultPlaceholder isLargeImage={isLargeImage} />
+    } else return isLargeImage ? <VideoPlayer src={data} /> : <VideoPlaceholder />
+}
```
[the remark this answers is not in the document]
aeshub: Good catch. Fixed in 15c6c6eb: small video cards now call a new lightweight `inspection/image/{id}/exists` endpoint (HEAD-style blob existence check, no download) and show the placeholder until the clip lands. The full blob is fetched only when the dialog opens.

backend/api/Controllers/InspectionController.cs
```diff
@@ -69,22 +72,22 @@ await inspectionService.FetchInspectionImageFromIsarInspectionId(
...
             {
                 logger.LogWarning(
-                    "Could not find inspection image with ISAR Inspection ID {IsarInspectionId}",
+                    "Could not find inspection media with ISAR Inspection ID {IsarInspectionId}",
                     isarInspectionId
                 );
                 return NotFound(
-                    $"Could not find inspection image with ISAR Inspection ID{isarInspectionId}"
+                    $"Could not find inspection media with ISAR Inspection ID{isarInspectionId}"
                 );
```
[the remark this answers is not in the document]
aeshub: Fixed in 15c6c6eb.

backend/api/Controllers/InspectionController.cs
```diff
@@ -69,22 +72,22 @@ await inspectionService.FetchInspectionImageFromIsarInspectionId(
...
                 logger.LogError(
                     e,
-                    "Could not find inspection image with ISAR Inspection ID {IsarInspectionId}",
+                    "Could not find inspection media with ISAR Inspection ID {IsarInspectionId}",
                     isarInspectionId
                 );
                 return NotFound(
-                    $"Could not find inspection image with ISAR Inspection ID{isarInspectionId}."
+                    $"Could not find inspection media with ISAR Inspection ID{isarInspectionId}."
                 );
```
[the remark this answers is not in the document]
aeshub: Fixed in 15c6c6eb.

frontend/src/components/Contexts/InspectionsContext.tsx
```diff
@@ -27,6 +33,7 @@ interface Props {
 
 const defaultInspectionsContext = {
     useImageData: () => ({ data: undefined, isPending: false, isError: true }),
```
marianards:
```suggestion
    useMediaData: () => ({ data: undefined, isPending: false, isError: true }),
```

now useImageData makes no sense anymore since we also download videos, let's call it useMediaData instead

frontend/src/components/Contexts/InspectionsContext.tsx
```diff
@@ -100,6 +110,18 @@ export const InspectionsProvider: FC<Props> = ({ children }) => {
         return { data: result.data, isPending: result.isPending, isError: result.isError }
     }
 
+    const useMediaExists = (inspectionId: string): IExistsData => {
+        const result = useQuery({
+            queryKey: ['inspectionMediaExists', inspectionId],
+            queryFn: async () => backendApi.getInspectionMediaExists(inspectionId),
+            retry: 1,
+            staleTime: 10 * 60 * 1000, // If data is received, stale time is 10 min before making new API call
+            enabled: inspectionId !== undefined,
```
marianards, line 119:
```suggestion
```

reduntant as inspectionId cannot be undefined so suggest removing it but it is following the previous patters, can also change the others or keep this one as is

backend/api/Controllers/InspectionController.cs
```diff
@@ -69,26 +72,46 @@ await inspectionService.FetchInspectionImageFromIsarInspectionId(
...
         }
 
+        /// <summary>
+        /// Check whether inspection media is available for the given isarInspectionId
+        /// </summary>
+        /// <remarks>
+        /// Lightweight existence check so the frontend can show a placeholder until the
+        /// media has arrived, without downloading the blob.
+        /// </remarks>
+        [HttpGet("image/{isarInspectionId}/exists")]
```
marianards:
```suggestion
        [HttpGet("media/{isarInspectionId}/exists")]
```

backend/api/Services/InspectionService.cs
```diff
@@ -48,15 +51,51 @@ IBlobService blobService
...
-            return await blobService.DownloadBlob(
+            var blob = await blobService.DownloadBlob(
                 inspectionData.BlobName,
                 inspectionData.BlobContainer,
                 inspectionData.StorageAccount
             );
+            if (blob is null)
+            {
+                return null;
+            }
```
marianards, line 68:
```suggestion
            if (blob is null) return null;
```

This is just preference you don't need to change it if you dont want to

marianards:
DownloadContentAsync loads the full blob into server memory before sending. This was fine for images, but we had planned to eventually move to SAS tokens because of it. For videos this could be significant, and it also prevents the browser from doing range requests (which it needs for seeking). Worth considering a short-lived SAS URL redirect instead, I think flotilla needs Storage Blob Delegator RBAC role on, i also know that @andchiind has made a PR recently on this https://github.com/equinor/flotilla/pull/2809 so you should check with him too, if the  Storage Blob Delegator is the correct way to do this, then it should be few code changes

frontend/src/components/Contexts/InspectionsContext.tsx
aeshub: If it follows previous patterns I would keep it. Otherwise I would remove all of them but want to avoid additional stuff in here.

frontend/src/components/Contexts/InspectionsContext.tsx
aeshub: I avoided this rename on purpose but I think it was an oversight from me. Thought it was a bigger issue than it was. Will fix it.

aeshub: Done in 744b2c13 — renamed `useImageData` → `useMediaData` across the inspections context and the report image components. Left the `IImageData` return type as-is since `useAnalysisData` shares it and analysis really is an image.

backend/api/Controllers/InspectionController.cs
aeshub: Done in 449980b5. The download route was still `inspection/image/{id}`, so I renamed both endpoints — `inspection/media/{id}` and `inspection/media/{id}/exists` — for consistency, and updated the frontend `BackendApi` calls. Scanned the frontend and confirmed no other references to the old path remain.

backend/api/Services/InspectionService.cs
aeshub: Tried it: CSharpier 1.2.5 won't keep the one-liner — it reformats `if (blob is null) return null;` to a braceless two-line `if (blob is null)\n    return null;`, which would also be the only braceless `if` in the backend. Keeping the braced block so `csharpier check` stays green and the style stays consistent. Thanks though!

marianards (approved):
LGTM :)

PR #2824, opened 2026-07-09
Update with new SARA endpoints

### LINKED ISSUE #2710: Use sara directly instead of using missions
We now have a CLOE data view. This view reads all missions and has to filter them to find relevant tags and analysis results. We do not want to load so much data into the frontend and do that much filtering. Rather, we should have an endpoint in the sara backend where we can get the relevant data. This would isolate sara a bit more from flotilla. Also, we would remove much of the filtering in the frontend in the CLOE data view.

### LINKED ISSUE #2806: Make endpoint for tasks
We are downloading all mission runs and then doing filtering in the frontend. I believe that is in general a bad way of doing it. I suggest making an endpoint with paginated filtering on tasks on the backend. That will reduced the complexity of the frontend. 

As it is now, we have to do this by filtering on mission runs, since tasks do not have installation knowledge, then flatten to only tasks. There are som bi changes needed if we want to query on the tasks table directly. 
Also, it seems to be difficult to filter on the list of enums, AnalysisTypes. This will have to be solved. 

Once this endpoint exists, the frontend for data overviews will be better. However, we should also consider if the responsibilities of data viewing is better suited for sara than flotilla.

### LINKED ISSUE #2819: Analysis value/warning not displayed after AnalysisResults removal
### Summary

Analysis value / warning / confidence are no longer displayed anywhere in the
frontend that reads `task.inspection.analysisResult`. The field is always
`undefined`, so the affected components silently render nothing.

### Root cause

Migration `20260626095909_RemoveAnalysisResults` dropped the `AnalysisResults`
table. `backend/api/Database/Models/Inspection.cs` now exposes only
`AnalysisTypes` (no `AnalysisResult` property), and there is no Inspection
response DTO that re-adds one. The API therefore never sends `analysisResult`.

Nothing populates it client-side either: there are no assignments to
`analysisResult` anywhere in the frontend, and the `analysisResultReady` SignalR
handler (`InspectionsContext.tsx`) uses only `inspectionId` to refetch the SARA
image — it discards the value/warning/confidence in the message.

However `frontend/src/models/Inspection.ts:9` still declares
`analysisResult?: AnalysisResult`, so the following components read an
always-`undefined` field. Optional chaining hides the failure:

- `DataViewPage/DataView.tsx` (102, 152, 231)
- `DataViewPage/DataViewTable.tsx` (49, 79, 82, 83)
- `MissionPage/TaskOverview/TaskTable.tsx` (34, 67, 93, 95)
- `MissionPage/TaskOverview/TaskAnalysisDisplay.tsx` (21, 22)
- `MissionPage/AnalysisResultView.tsx` (46, 68–85) — image works via `useAnalysisData`; warning/value/confidence are blank
- `MissionHistory/HistoricMissionCard.tsx` (44)

This regresses, among others, the CLOE table-row warning colouring from #2755.

### Suggested fix

A live numeric path already exists: `backendApi.getValue(inspectionId)` wrapped
by `useValueData` (`InspectionsContext.tsx:117`), currently consumed only by
`InspectionReportImage.tsx`. Remediation:

1. Migrate numeric reads from `analysisResult?.value` to `useValueData`, and
   image reads to `useAnalysisData`, as `InspectionReportImage.tsx` already does.
2. Remove the dead `analysisResult` field from `models/Inspection.ts` once no
   consumers remain.
3. Warning / confidence currently have no live source — needs a backend
   decision on whether to expose them (e.g. persist, or a dedicated endpoint).

### Notes

The combined inspection/analysis card (`ImageOverview.tsx`, #2791) already
renders the analysis image via `analysisTypes` and is not affected.

andchiind:
This PR updates Flotilla to reflect recent changes in SARA.

The biggest change is that the SARA data view relies solely on SARA endpoints. At the moment, in main, the data view does not work as intended. To facilitate this change, a lot of components were rewritten to use the InspectionData model, as opposed to inferring info from tasks.

- [ ] A test has been written

### DIFF
--- a/frontend/src/api/BackendApi.tsx
+++ b/frontend/src/api/BackendApi.tsx
@@ -254,15 +254,6 @@ export class BackendApi {
         return result.content
     }
 
-    async getValue(inspectionId: string): Promise<number> {
-        const path: string = 'inspection/value/' + inspectionId
-
-        return this.api
-            .GET<number>(path)
-            .then((response) => response.content)
-            .catch(handleError('GET', path))
-    }
-
     async skipAutoScheduledMission(missionId: string, timeOfDay: string): Promise<void> {
         const path: string = `missions/definitions/${missionId}/skip-auto-mission`
         const body = { timeOfDay: timeOfDay }
--- a/frontend/src/api/SaraApi.tsx
+++ b/frontend/src/api/SaraApi.tsx
@@ -1,31 +1,57 @@
-import { InspectionRecord } from 'models/InspectionRecord'
+import { InspectionData, InspectionRecord, inspectionRecordToInspectionData } from 'models/InspectionRecord'
 import { BackendAPICaller } from './ApiCaller'
 import { handleError } from './ApiError'
+import { AnalysisType } from 'models/MissionDefinition'
+
+interface PaginatedInspectionRecords {
+    items: InspectionRecord[]
+    pageNumber: number
+    pageSize: number
+    totalCount: number
+    totalPages: number
+    hasNext: boolean
+    hasPrevious: boolean
+}
 
 export class SaraApi {
     constructor(private readonly api: BackendAPICaller) {}
 
-    async getAnonymizedDataUrl(inspectionId: string): Promise<string> {
+    async getSaraDataByInspectionId(inspectionId: string): Promise<InspectionData> {
         const path: string = 'api/inspection-record/inspection-id/' + inspectionId
 
         const content = this.api
             .GET<InspectionRecord>(path)
             .then((response) => {
                 if (!response.content.analyses || response.content.analyses.length < 1) throw Error('No analysis found')
-                return response.content.analyses[response.content.analyses.length - 1].anonymizedSAS
+                return inspectionRecordToInspectionData(response.content)
             })
             .catch(handleError('GET', path))
         return content
     }
 
-    async getAnalysedDataUrl(inspectionId: string): Promise<string> {
-        const path: string = 'api/inspection-record/inspection-id/' + inspectionId
+    async getSaraData(
+        inspectionIds?: string[] | null,
+        installationCode?: string | null,
+        tagId?: string | null,
+        analysisType?: AnalysisType | null,
+        minDate?: Date | null,
+        maxDate?: Date | null
+    ): Promise<InspectionData[]> {
+        let path: string = 'api/inspection-record?'
+
+        if (inspectionIds) path = path + inspectionIds.map((i) => 'InspectionIds=' + i).join('&')
+        if (installationCode) path = path + 'InstallationCode=' + installationCode + '&'
+        if (tagId) path = path + 'Tag=' + tagId + '&'
+        if (analysisType) path = path + 'AnalysisTypes=' + [analysisType] + '&'
+        if (minDate) path = path + 'MinCreationTime=' + minDate.toISOString() + '&'
+        if (maxDate) path = path + 'MaxCreationTime=' + maxDate
 
         const content = this.api
-            .GET<InspectionRecord>(path)
+            .GET<PaginatedInspectionRecords>(path)
             .then((response) => {
-                if (!response.content.analyses || response.content.analyses.length < 1) throw Error('No analysis found')
-                return response.content.analyses[response.content.analyses.length - 1].visualizedSAS
+                if (!response.content) throw Error('No inspection records found')
+
+                return response.content.items.map((r) => inspectionRecordToInspectionData(r))
             })
             .catch(handleError('GET', path))
         return content
--- a/frontend/src/components/Contexts/InspectionsContext.tsx
+++ b/frontend/src/components/Contexts/InspectionsContext.tsx
@@ -1,25 +1,32 @@
 import { createContext, FC, useContext, useEffect, useRef } from 'react'
 import { SignalREventLabels, useSignalRContext } from './SignalRContext'
-import { SaraAnalysisResultReady, SaraInspectionVisualizationReady } from 'models/Inspection'
 import { useQuery } from '@tanstack/react-query'
 import { queryClient } from '../../App'
 import { useBackendApi } from 'api/UseBackendApi'
 import { useSaraApi } from 'api/UseSaraApi'
+import { FlotillaAnalysisResultMessage, InspectionData } from 'models/InspectionRecord'
+import { AnalysisType } from 'models/MissionDefinition'
 
-interface IImageData {
-    data: string | undefined
+interface IInspectionData {
+    data: InspectionData | undefined
     isPending: boolean
     isError: boolean
 }
-interface IValueData {
-    data: number | undefined
+interface IInspectionListData {
+    data: InspectionData[] | undefined
     isPending: boolean
     isError: boolean
 }
 interface IInspectionsContext {
-    useMediaData: (inspectionId: string) => IImageData
-    useAnalysisData: (inspectionId: string) => IImageData
-    useValueData: (inspectionId: string) => IValueData
+    useSaraData: (inspectionId: string) => IInspectionData
+    useSaraListData: (
+        inspectionIds?: string[] | null,
+        installationCode?: string | null,
+        tagId?: string | null,
+        analysisType?: AnalysisType | null,
+        minDate?: Date | null,
+        maxDate?: Date | null
+    ) => IInspectionListData
 }
 
 interface Props {
@@ -27,9 +34,8 @@ interface Props {
 }
 
 const defaultInspectionsContext = {
-    useMediaData: () => ({ data: undefined, isPending: false, isError: true }),
-    useAnalysisData: () => ({ data: undefined, isPending: false, isError: true }),
-    useValueData: () => ({ data: undefined, isPending: false, isError: true }),
+    useSaraData: () => ({ data: undefined, isPending: false, isError: true }),
+    useSaraListData: () => ({ data: undefined, isPending: false, isError: true }),
 }
 
 const InspectionsContext = createContext<IInspectionsContext>(defaultInspectionsContext)
@@ -50,15 +56,14 @@ export const InspectionsProvider: FC<Props> = ({ children }) => {
     useEffect(() => {
         if (connectionReady) {
             registerEvent(SignalREventLabels.inspectionVisualizationReady, (username: string, message: string) => {
-                const inspectionVisualizationData: SaraInspectionVisualizationReady = JSON.parse(message)
-                const inspectionId = inspectionVisualizationData.inspectionId
+                const inspectionId: string = JSON.parse(message)
                 queryClient.invalidateQueries({
                     queryKey: ['fetchInspectionData', inspectionId],
                 })
                 queryClient.fetchQuery({
                     queryKey: ['fetchInspectionData', inspectionId],
                     queryFn: async () => {
-                        return await saraApiRef.current.getAnonymizedDataUrl(inspectionId)
+                        return await saraApiRef.current.getSaraDataByInspectionId(inspectionId)
                     },
                     retry: 1,
                     staleTime: 10 * 60 * 1000,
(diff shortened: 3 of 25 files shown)


frontend/src/api/BackendApi.tsx
```diff
@@ -254,15 +254,6 @@ export class BackendApi {
...
 
-    async getValue(inspectionId: string): Promise<number> {
-        const path: string = 'inspection/value/' + inspectionId
-
-        return this.api
-            .GET<number>(path)
-            .then((response) => response.content)
-            .catch(handleError('GET', path))
-    }
-
```
andchiind, line 265: This technically makes that endpoint unused, but I didn't remove it in this PR since it might still be used for debugging in swagger

frontend/src/pages/DataViewPage/DataView.tsx
```diff
@@ -53,255 +56,199 @@ interface DataViewProps {
...
+        null
+    )
+
+    if (isPending) {
+        return <PendingResultPlaceholder isLargeImage={true} />
+    } else if (isError || !data) {
+        return <TextAsImage isLargeImage={true} text={'No inspection could be found'} />
+    }
+
+    // Fetches only the latest inspection data per tag
```
andchiind, line 233: This is what was being done before the refactor as well, though if we want to update it, to use a running average for example, then that is much easier to do now.

frontend/src/pages/MissionPage/MissionPage.tsx
```diff
@@ -120,25 +122,42 @@ export const MissionPage = ({
...
         : false
 
+    const { data, isPending, isError } = useSaraListData(
+        selectedMission?.tasks.map((t) => t.inspection.isarInspectionId),
+        null,
+        null,
+        null,
+        null,
+        null
+    )
```
andchiind, line 147: An improvement could maybe to use arguments with default values.

frontend/src/api/BackendApi.tsx
nicholasdalhaug: If we were to test it, would we not just use sara to read the value? Then you can delete the endpoint.

frontend/src/api/SaraApi.tsx
```diff
@@ -1,31 +1,57 @@
...
+        if (inspectionIds) path = path + inspectionIds.map((i) => 'InspectionIds=' + i + '&').join('&')
+        if (installationCode) path = path + 'InstallationCode=' + installationCode + '&'
+        if (tagId) path = path + 'Tag=' + tagId + '&'
+        if (analysisType) path = path + 'AnalysisTypes=' + [analysisType] + '&'
+        if (minDate) path = path + 'MinCreationTime=' + minDate.toUTCString() + '&'
+        if (maxDate) path = path + 'MaxCreationTime=' + maxDate
 
         const content = this.api
-            .GET<InspectionRecord>(path)
+            .GET<PaginatedInspectionRecords>(path)
```
nicholasdalhaug, line 50: This does not look paginated, especially since the page is not in the path. Is it paginated?

frontend/src/components/Displays/TaskDisplay.tsx
```diff
@@ -1,13 +1,11 @@
 import { Typography } from '@equinor/eds-core-react'
-import { MissionTaskDefinition } from 'models/MissionDefinition'
-import { Task } from 'models/Task'
 
-export const TagIdDisplay = ({ task, index }: { task: Task | MissionTaskDefinition; index: number }) => {
-    if (!task.tagId) return <Typography key={index + 'tagId'}>{'N/A'}</Typography>
-    else return <Typography key={index + 'tagId'}>{task.tagId!}</Typography>
+export const TagIdDisplay = ({ tagId, index }: { tagId: string | undefined; index: number }) => {
+    if (!tagId) return <Typography key={index + 'tagId'}>{'N/A'}</Typography>
+    else return <Typography key={index + 'tagId'}>{tagId!}</Typography>
```
nicholasdalhaug, line 5: Nice!

frontend/src/components/Displays/TimeseriesLinePlot.tsx
```diff
@@ -42,31 +41,27 @@ const PALETTE_COLORS = [
...
-    onPointClick?: (point: { id: string; time: Date; value: number }) => void
-    selectedPoint: TimeseriesLinePlotSelectedPoint | undefined
+    onPointClick?: (point: TimeseriesLinePlotDataPoint) => void
+    selectedInspectionId: string | undefined
 }
 
-export const TimeseriesLinePlot = ({ data, yLabel, ymin, ymax, onPointClick, selectedPoint }: Props) => {
-    const dataEntries = useMemo(() => Object.entries(data), [data])
+export const TimeseriesLinePlot = ({ data, yLabel, ymin, ymax, onPointClick, selectedInspectionId }: Props) => {
+    const dataEntries = Object.values(data)
```
nicholasdalhaug: Removing the useMemo makes the options object change every render. Maybe we should not do that? I am not sure if it makes a difference, it just seems like a bad pattern.

nicholasdalhaug: Also, since they are not the whole entries now, we shoul rename the variable. I suggest `dataValues`.

frontend/src/models/InspectionRecord.ts
```diff
@@ -25,3 +42,63 @@ export interface InspectionRecord {
...
+
+const videoFileEndings = ['mp4', 'mpg', 'mpeg', 'm4v']
+
+const sasURLToFileType = (sasURL: string): FileType => {
+    const reg: RegExp = /\.([a-zA-Z0-9]+)(\?skoid=)/
+    const fileEnding = sasURL.match(reg)?.[1]
+    if (!fileEnding) return FileType.VALUE
+    if (imageFileEndings.includes(fileEnding?.toLowerCase())) return FileType.IMAGE
+    if (videoFileEndings.includes(fileEnding.toLowerCase())) return FileType.VIDEO
+    return FileType.VALUE
```
nicholasdalhaug, line 80: Should this maybe throw an error instead? It cannot be a value if it has a fileending ,right?

frontend/src/api/SaraApi.tsx
```diff
@@ -1,31 +1,57 @@
...
+        inspectionIds?: string[] | null,
+        installationCode?: string | null,
+        tagId?: string | null,
+        analysisType?: AnalysisType | null,
+        minDate?: Date | null,
+        maxDate?: Date | null
+    ): Promise<InspectionData[]> {
+        let path: string = 'api/inspection-record?'
+
+        if (inspectionIds) path = path + inspectionIds.map((i) => 'InspectionIds=' + i + '&').join('&')
```
Christdej: Might be better to use URLSearchParams and repeat append('InspectionIds', id) for each

frontend/src/api/SaraApi.tsx
```diff
@@ -1,31 +1,57 @@
...
+        if (installationCode) path = path + 'InstallationCode=' + installationCode + '&'
+        if (tagId) path = path + 'Tag=' + tagId + '&'
+        if (analysisType) path = path + 'AnalysisTypes=' + [analysisType] + '&'
+        if (minDate) path = path + 'MinCreationTime=' + minDate.toUTCString() + '&'
+        if (maxDate) path = path + 'MaxCreationTime=' + maxDate
```
Christdej: Use .toISOString() instead

frontend/src/components/Contexts/InspectionsContext.tsx
```diff
@@ -114,26 +110,38 @@ export const InspectionsProvider: FC<Props> = ({ children }) => {
...
+                    tagId,
+                    analysisType,
+                    minDate,
+                    maxDate
+                )
             },
             retry: 1,
             staleTime: 10 * 60 * 1000, // If data is received, stale time is 10 min before making new API call
-            enabled: inspectionId !== undefined,
+            enabled: installationCode !== undefined,
```
Christdej: Remember that undefined != null

frontend/src/pages/MissionPage/TaskOverview/TaskTable.tsx
```diff
@@ -58,47 +60,42 @@ export const MissionDefinitionTaskTable = ({ tasks }: MissionDefinitionTaskTable
...
-        const rowStyle =
-            task.status === TaskStatus.InProgress || task.status === TaskStatus.Paused
-                ? { background: tokens.colors.infographic.primary__mist_blue.hex }
-                : task.inspection.analysisResult?.warning
-                  ? { background: tokens.colors.interactive.danger__highlight.hex }
-                  : {}
-        const markerColors = getColorsFromTaskStatus(task.status)
+const TaskTableRow = ({ task, index }: { task: Task; index: number }) => {
+    const { useSaraData } = useInspectionsContext()
+    const { data, isPending } = useSaraData(task.inspection.isarInspectionId)
```
Christdej: This will do a HTTP call per task, even though we already have the full list via useSaraListData on the parent.

frontend/src/pages/InspectionReportPage/InspectionView.tsx
```diff
@@ -19,52 +17,132 @@ import { useContext, useState } from 'react'
...
+                <StyledDialogContent>
+                    <StyledDialogInspectionView>
+                        <TextAsImage isLargeImage={true} text="No inspection could be found" />
+                    </StyledDialogInspectionView>
+                </StyledDialogContent>
+            </StyledDialog>
+        )
+    }
+
+    document.addEventListener('keydown', (event) => {
```
Christdej, line 52: Since this is inside a render body, every re-render adds another global listener that is never removed. Arrow-key handling will fire N times and hold stale state, move to useEffect with cleanup. This goes for the other addEventListeners aswell

Christdej (dismissed):

frontend/src/pages/DataViewPage/ThermalReadingViewPage.tsx
```diff
@@ -14,7 +15,7 @@ export const ThermalReadingViewPage = () => {
 
     return (
         <DataView
-            analysisType={AnalysisTypes.ThermalReading}
+            analysisType={AnalysisType.ThermalReading}
```
nicholasdalhaug, line 18: Nice!

frontend/src/pages/InspectionReportPage/ImageOverview.tsx
nicholasdalhaug, line 1: I am not sure I have seen the InspectionReportPage before. Not sure what it is.

nicholasdalhaug (changes requested):
Looks very good. I am not sure I understood all details since it was rather large and a lot to keep in mind. Maybe next time the PR can be half the size :) I would like to discuss this when you have the time.

frontend/src/components/Contexts/InspectionsContext.tsx
andchiind: installationCode shouldn't be null, but I can add that check anyways

frontend/src/api/SaraApi.tsx
andchiind: In the data view we don't paginate it, but the results are paginated, we just use the first page. This PR just recreates the existing data view functionality, it doesn't change it.

frontend/src/api/BackendApi.tsx
andchiind: I would have to check with @aeshub when he is back, since it does some more post-processing. For now I would like to keep it

frontend/src/models/InspectionRecord.ts
andchiind: It can have a value if it is a fileEnding that we aren't able to display. The output of a workflow could be a JSON, while at the same time the "result" value can be set.

frontend/src/pages/InspectionReportPage/ImageOverview.tsx
andchiind: It's not something new in this PR

frontend/src/pages/InspectionReportPage/InspectionView.tsx
andchiind: This code isn't new in this PR, I just moved it. Good comments but I will consider them out of scope of this PR.

frontend/src/api/SaraApi.tsx
andchiind: Wouldn't this also involve changing the local URL? Though I might be misunderstanding

frontend/src/api/SaraApi.tsx
andchiind: I can make an issue to do so in a future PR maybe. Though there is already an issue relating to using more than just the latest inspection data per tag.

frontend/src/api/SaraApi.tsx
andchiind: I think I might keep it as is for now, since this is more similar to how it has been done elsewhere in the code, but that could be a good idea for a refactor.

frontend/src/pages/MissionPage/TaskOverview/TaskTable.tsx
andchiind: Fixed

nicholasdalhaug (approved):

PR #2839, opened 2026-07-24
Remove sqlite in memory and use postgresql testcontainer locally

nicholasdalhaug:


nicholasdalhaug:
I do not think that this requires a migration.

andchiind (approved):
Good stuff!

PR #2846, opened 2026-07-29
Fix mission scheduling race condition

olaals:
## Why

Concurrent callers of `StartNextMissionRunIfSystemIsAvailable` (the scheduling controller, the auto scheduler, and the MQTT event handler) could pick and start the **same** queued mission run at the same time, causing duplicate/aborted starts.

## What

- Add a per-robot async lock (`MissionSchedulingLock`, registered as a singleton) held around the whole `StartNextMissionRunIfSystemIsAvailable` flow. Internal recursive calls now go through a private `...Internal` method so the lock is taken exactly once per top-level attempt.
- Immediately move a started run out of the `Queued` state (to `Ongoing`) while the lock is still held, so a concurrent scheduling attempt cannot re-read and start the same run before ISAR reports status over MQTT. The MQTT status update remains the source of truth and overwrites this.

## Notes

- Test setup updated to pass a `MissionSchedulingLock` instance.

Happens in local orchestration on a scheduled mission (happens several times in a row), duplicated mission even when only one mission was scheduled:

image

### DIFF
--- a/backend/api.test/TestSetupHelpers.cs
+++ b/backend/api.test/TestSetupHelpers.cs
@@ -190,7 +190,8 @@ public static class TestSetupHelpers
             signalRService,
             _errorHandlingService,
             _inspectionAreaService,
-            _exclusionAreaService
+            _exclusionAreaService,
+            new MissionSchedulingLock()
         );
         var _missionDefinitionService = new MissionDefinitionService(
             context,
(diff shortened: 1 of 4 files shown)


olaals:
Closing: the observed duplicate-mission symptom is caused by a frontend StrictMode double-POST, fixed in #2848. The backend scheduling lock proposed here is not needed to address that bug.

PR #2847, opened 2026-07-29
Warm up MSAL token pipeline at startup

olaals:
## Why

Microsoft.Identity.Web builds its `MergedOptions` and the underlying MSAL confidential-client application **lazily on the first token request**. That first initialization is not safe against a second concurrent caller: while one thread populates the merged options, another concurrent ISAR call can observe a half-initialized configuration and fail with `"No ClientId was specified."`.

## What

- Add `MsalWarmupHostedService` (registered as a hosted service) that acquires a single application (client-credentials) access token for the ISAR downstream API **before the web server starts accepting requests**, forcing the options and confidential client to be fully built once.
- Reads the scope from `Isar:Scopes:0`; if none is configured it logs a warning and skips.
- Uses a 30s timeout and is fully non-fatal: a failed warm-up (e.g. no network or invalid secret in local dev) is logged and startup continues, with the token acquired lazily on first real use.

## Notes

Observed with local orchestration on fresh start, happens every time: 

image

### DIFF
--- a/backend/api/HostedServices/MsalWarmupHostedService.cs
+++ b/backend/api/HostedServices/MsalWarmupHostedService.cs
@@ -1,0 +1,72 @@
+using Microsoft.Identity.Web;
+
+namespace Api.HostedServices
+{
+    /// <summary>
+    /// Warms up the Microsoft.Identity.Web token-acquisition pipeline at startup by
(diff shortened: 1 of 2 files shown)


olaals:
Solved by fixing the duplicate mission scheduling

PR #2848, opened 2026-07-29
Fix duplicate mission scheduling from a single click

olaals:
# Problem
## Step 1 — The app is wrapped in `<StrictMode>`

`frontend/src/index.tsx:41-46`
```tsx
ReactDOM.createRoot(rootElement).render(
    <StrictMode>
        ...
    </StrictMode>
)
```
In development builds, StrictMode deliberately **mounts → unmounts → remounts** each component and **runs effects twice** to surface impure/unsafe effects. (In production this doubling does not happen — which is why the bug shows up under Tilt but is easy to miss.)

## Step 2 — User clicks "Schedule", which opens the verification dialog

`frontend/src/pages/InspectionPage/ScheduleMissionDialogs.tsx:82-86`
```tsx
const onScheduleButtonPress = (missions: MissionDefinition[]) => () => {
    if (!selectedRobot) return
    setMissionsToSchedule(missions)
    setIsInspectionAreaVerificationDialogOpen(true)   // just opens the dialog
}
```
Notice the click handler itself does **not** schedule anything. It only sets state to render the dialog component. The actual scheduling was delegated *into* that dialog.

## Step 3 — The dialog mounts and decides it can schedule directly

`frontend/src/pages/InspectionPage/ScheduleMissionDialogs.tsx:197-198` renders:
```tsx
<ScheduleMissionWithInspectionAreaVerification
    scheduleMissions={scheduleMissions}
    ...
/>
```

Inside that component, it computes whether any user confirmation is actually needed:

`ScheduleMissionWithInspectionAreaVerification.tsx:51-53`
```tsx
const resolvedDialog = getDialogToOpen()                       // undefined = no conflict
const dialogToOpen = resolvedDialog ?? DialogTypes.unknown
const shouldScheduleDirectly = !!selectedRobot && resolvedDialog === undefined
```
When the robot is already in the right inspection area (the common case), `getDialogToOpen()` returns `undefined`, so `shouldScheduleDirectly === true` — meaning "no dialog needed, just schedule."

## Step 4 — The effect fires the request (this is the bug)

`ScheduleMissionWithInspectionAreaVerification.tsx:55-57`
```tsx
useEffect(() => {
    if (shouldScheduleDirectly) scheduleMissions()
}, [shouldScheduleDirectly])
```
This is a **side effect (an HTTP POST) executed from render lifecycle**, with no idempotency guard (no `useRef`/"already fired" flag). Under StrictMode's mount → unmount → remount:

1. **First mount** → effect runs → `scheduleMissions()` → **POST 1**
2. StrictMode unmounts (the cleanup here is empty — nothing cancels the in-flight request)
3. **Remount** → `shouldScheduleDirectly` is still `true` → effect runs again → `scheduleMissions()` → **POST 2**

Two POSTs from one click. This matches the observed log: two simultaneous requests, connIds `0HNNDIMGLUVV0` + `0HNNDIMGLUVV2` at 19:20:51.

## Step 5 — Each POST becomes its own MissionRun

`scheduleMissions` loops and calls the backend per mission:

`ScheduleMissionDialogs.tsx:89-94`
```tsx
const scheduleMissions = () => {
    setIsInspectionAreaVerificationDialogOpen(false)
    if (!selectedRobot || !missionsToSchedule) return
    missionsToSchedule.forEach((mission) => {
        backendApi.scheduleMissionDefinition(mission.id, selectedRobot.id)...
    })
}
```
On the backend, each `POST /missions/schedule/...` creates exactly one **Queued** `MissionRun` and then starts the oldest queued run (`MissionSchedulingController.cs:200-233`). So two POSTs deterministically produce **two MissionRuns** — the "two missions from one click" symptom.

## Why

Scheduling a mission (or restarting one) from the UI creates **two** `MissionRun`s from a single user action. Backend logs show two simultaneous `POST /missions/schedule/...` requests per click.

## Root cause

`ScheduleMissionWithInspectionAreaVerification` fired the schedule request from a `useEffect` that ran on mount. Under React `<StrictMode>` (active in dev builds), effects are intentionally invoked twice, so the request was sent twice.

## Fix

Move the scheduling request out of the effect and into the click handler:

- `getInspectionAreaDialogType` (new pure helper) decides at the call site whether a verification dialog is needed.
- If no verification is needed, the request fires directly (exactly once), no dialog.
- `ScheduleMissionWithInspectionAreaVerification` is now purely presentational — it takes a `dialogType` and `robot`, shows the message, and no longer schedules anything itself.
- Both call sites (`ScheduleMissionDialogs`, `MissionRestartButton`) resolve the live robot from context and own the scheduling call.

This removes the root cause and is immune to StrictMode double-invocation, rather than guarding the effect with a ref.

## Verification

- `npm run build` (tsc + vite) — clean
- `npm run lint` — 0 errors on changed files
- `npm run prettier_check` — clean

### DIFF
--- a/frontend/src/components/Displays/InspectionAreaVerificationDialogs/ScheduleMissionWithInspectionAreaVerification.tsx
+++ b/frontend/src/components/Displays/InspectionAreaVerificationDialogs/ScheduleMissionWithInspectionAreaVerification.tsx
@@ -1,79 +1,46 @@
-import { useEffect } from 'react'
 import {
     ConflictingMissionInspectionAreasDialog,
     ConflictingRobotInspectionAreaDialog,
 } from './ConflictingInspectionAreaDialog'
 import { UnknownInspectionAreaDialog } from './UnknownInspectionAreaDialog'
-import { useAssetContext } from 'components/Contexts/AssetContext'
 import { InspectionArea } from 'models/InspectionArea'
+import { RobotWithoutTelemetry } from 'models/Robot'
+import { getUniqueInspectionAreas, InspectionAreaDialogType } from './getInspectionAreaDialogType'
 
 interface IProps {
-    scheduleMissions: () => void
+    dialogType: InspectionAreaDialogType
     closeDialog: () => void
-    robotId: string
+    robot: RobotWithoutTelemetry | undefined
     missionInspectionAreas: InspectionArea[]
 }
 
-enum DialogTypes {
-    unknownNewInspectionArea,
-    conflictingMissionInspectionAreas,
-    conflictingRobotInspectionArea,
-    unknown,
-}
-
 export const ScheduleMissionWithInspectionAreaVerification = ({
-    robotId,
-    missionInspectionAreas,
-    scheduleMissions,
+    dialogType,
     closeDialog,
+    robot,
+    missionInspectionAreas,
 }: IProps) => {
-    const { enabledRobots } = useAssetContext()
-
-    const unikMissionInspectionAreas = missionInspectionAreas.filter(
-        (inspectionArea, index, self) => self.findIndex((i) => i.id === inspectionArea.id) === index
+    const uniqueMissionInspectionAreaNames = getUniqueInspectionAreas(missionInspectionAreas).map(
+        (area) => area?.inspectionAreaName ?? ''
     )
-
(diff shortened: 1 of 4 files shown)


andchiind (approved):
OK to merge. Code is a little obfuscated, but so was the code before. Please make an issue to refactor ScheduleMissionDialogs.tsx and child components before merging

PR #2863, opened 2026-08-05
Select the identity provider by configuration

oysand:
Adds `Authentication:Provider`, defaulting to `EntraId`. Setting it to `Oidc` points both inbound token validation and outbound downstream-API token acquisition at any conformant OpenID Connect issuer — a Keycloak realm for local development and for the armada integration tests, or an operator's own issuer for a deployment outside Azure.

This is not an authentication bypass: issuer, audience, signature, lifetime and role validation stay enabled under either value, only the issuer differs. What is gated is transport security — a plain HTTP authority is accepted only in `Local` and `IntegrationTest`, and throws at startup anywhere else.

`IntegrationTest` has no appsettings file here on purpose. The name exists only to permit that plain-HTTP authority; the configuration behind it is passed as environment variables from armada, next to the Keycloak realm that defines the clients and scopes, so the two cannot drift apart across repositories.

Two implementation notes worth a reviewer's attention:

- The inbound override is split across `Configure` and `PostConfigure`. `JwtBearerPostConfigureOptions` rejects an HTTP authority and runs before any post-configuration we can register, while Microsoft.Identity.Web installs its `AadIssuerValidator` during post-configuration.
- Outbound, `IDownstreamApi` resolves every bearer token through `IAuthorizationHeaderProvider`, so replacing that one service covers ISAR, SARA and Pointilla. The token endpoint is read from the discovery document rather than assumed.
- **Application tokens only.** On-behalf-of is not implemented for a generic issuer, and the user overloads throw `NotSupportedException` rather than quietly substituting an app token — which would send a service principal where the caller asked for the signed-in user. This affects `PointillaService`, the one downstream caller using `CallApiForUserAsync`, and only under `Provider=Oidc`. Note that `CreateAuthorizationHeaderAsync` is the *unified* entry point and must still serve the app flow: it refuses only when `RequestAppToken` is not set.

**Commit 1 is unrelated and separable**: `MockSignalRService` has a pre-existing race that fails on clean `main` (3 of 3 isolated runs). It is included because it stood between this branch and a green CI run; say the word and I will split it out.

Verified: 106/106 tests, `dotnet build -warnaserror` and csharpier clean, each commit builds individually. The Flotilla → ISAR hop is proven in the full integration suite (equinor/armada#87, 4/4 green).

Merge order: equinor/isar#1158 → isar-robot bump → **this PR** + equinor/sara#456 → images published → equinor/armada#87.

Supersedes #2856, which stays open as a fallback until this set is approved.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: broker broker-aspire compose
+.PHONY: broker broker-aspire compose keycloak
 
 broker-aspire:
 	docker compose up broker otel-collector aspire-dashboard --build
@@ -8,3 +8,7 @@ compose:
 
 broker:
 	docker compose up broker --build
+
+# Local OpenID Connect issuer, an alternative to Entra ID. See backend/README.md.
+keycloak:
+	docker compose --profile keycloak up keycloak
--- a/backend/README.md
+++ b/backend/README.md
@@ -72,6 +72,44 @@ dotnet ef migrations remove
 - **Development**: after merging a PR that touches `backend/api/Migrations`, manually run the ["Run database migrations (Development)"](https://github.com/equinor/flotilla/actions/workflows/run_development_migrations.yml) workflow.
 - **Staging / Production**: applied automatically by the [deploy_to_staging](https://github.com/equinor/flotilla/blob/main/.github/workflows/deploy_to_staging.yml) and [promote_to_production](https://github.com/equinor/flotilla/blob/main/.github/workflows/promote_to_production.yml) workflows.
 
+## Authentication
+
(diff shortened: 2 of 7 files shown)


oysand:
The Flotilla → ISAR hop is now verified end to end (equinor/armada#87, full suite green). From the run:

```
Resolved token endpoint http://keycloak:8080/realms/robotics/protocol/openid-connect/token
  from http://keycloak:8080/realms/robotics/.well-known/openid-configuration
Acquiring token for scope 'isar-api' from http://keycloak:8080/realms/robotics/protocol/openid-connect/token
```

followed by 9 `Api.Services.IsarService` calls. Both Keycloak-specific fixes to `GenericOidcAuthorizationHeaderProvider` — discovery-based token endpoint, and sending `client_secret` — work against a real downstream service. That was the only gap listed under Verification; this PR is now fully verified.

Christdej (approved):
lgtm

PR #2864, opened 2026-08-06
Do not schedule missions without tasks from controller

### LINKED ISSUE #2838: Add error message when queuing empty mission
If you add a predefined mission to the queue that has  **no tasks** some error message or dialog box should inform you that the mission is empty. 

The way it is now, you can add to the queue an empty mission with no tasks, it will buffer in the Mission Queue and only when you refresh the page it disappears from the queue and appears as an Aborted mission in the mission history. And the user is not informed that this is because the mission has no tasks.

Image

Alternatively, not be allowed to queue an empty mission.

nicholasdalhaug:


nicholasdalhaug:
The frontend now shows this:
image

Eddasol (approved):
LGTM :)

PR #2891, opened 2026-08-18
Fix frontend thermal data view

nicholasdalhaug:
The problem was mainly that the filetype was read wrong. For some reason the sasURLs have parameters after them. We now remove those. Also, we can assume that it is a valid sasURL or throw an error otherwise. 

Fixed 2 other small problems.

### DIFF
--- a/backend/api/Database/Context/InitDb.cs
+++ b/backend/api/Database/Context/InitDb.cs
@@ -198,7 +198,7 @@ namespace Api.Database.Context
             var task1 = new TaskDefinition
             {
                 Index = 1,
-                TagId = "dummy tag id 1",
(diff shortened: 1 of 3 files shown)


andchiind (approved):

Eddasol (approved):
Tested and it works well :)

Previous i got this "TaskAnalysisDisplay.tsx:22 React does not recognize the `hasFinding` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `hasfinding` instead. If you accidentally passed it from a parent component, remove it from the DOM element." in the console and no photo. With this PR I get a photo and no error message

PR #2894, opened 2026-08-19
Remove result button when no result

### LINKED ISSUE #2887: Do not show Analysis column on mission runs without results
Now we see that there is a column for analysis even though we have no analysis (except anonymize) on these images. 

Image

Solution: Change frontend to not show "Results" button when `result: null` in the inspection results data.

nicholasdalhaug:


andchiind (approved):
LGTM. If you've tested that it works then feel free to merge

PR #2904, opened 2026-08-21
Add feedback on results

nicholasdalhaug:


nicholasdalhaug:
Looks like this:
image

andchiind (approved):

nicholasdalhaug:
You might call it a breaking change since the frontend will complain if the sara endpoint does not exist.

PR #2906, opened 2026-08-24
Feedback from users on Flotilla

### LINKED ISSUE #2886: Feedback button
Add a possibility to add feedback in flotilla frontend

markusmoen:
### Summary
Adds an in-app feedback flow that lets authenticated users submit bug reports, feature requests, or other feedback directly to the Flotilla team.

### Changes
Adds a localized feedback dialog accessible from the header
Prefills user identity from MSAL and validates form input
Adds the POST /submit-feedback backend endpoint
Sends feedback to a dedicated Microsoft Teams webhook
Refactors Teams notifications to support multiple destinations
Adds backend controller and Teams notification service tests
### Configuration
Requires changes in [infrastructure](https://github.com/equinor/robotics-infrastructure/blob/main/k8s_kustomize/robotics/base/backend/backend.yaml) and adding secrets to key vaults. 
> [!IMPORTANT]
>  The name of the webhook url for system alerts has also been changed

- [x] A test has been written

Closes #2886


backend/api/Utilities/TeamsAdaptiveCard.json
```diff
@@ -6,13 +6,13 @@
       "content": {
         "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
         "type": "AdaptiveCard",
-        "version": "1.6",
```
aeshub, line 9: What is the reason for having to perform this downgrade?

backend/api/Services/TeamsNotificationService.cs
```diff
@@ -0,0 +1,89 @@
...
+                return;
+
+            logger.LogError(
+                "Failed to send Teams message to {Destination}. Status code: {StatusCode}, Reason: {ReasonPhrase}",
+                destination,
+                response.StatusCode,
+                response.ReasonPhrase
+            );
+            response.EnsureSuccessStatusCode();
+        }
```
aeshub: The setup currently will log this error three times. First here, then again in the controller and again when the controller throws the unhandled exception. Suggestion is to log only once, but to catch the error here and add any information you would like to the error.

So here you can do a try/catch around response.EnsureSuccessStatusCode and then catch the specific status codes you are interested in (if any) or just have to try/catch at all and no logging and just let the error trickle up to the controller which will catch it. You can also to a catch HttpException and add the destination, status code and reason_phrase to the message that way.

The controller should have a catch all at the bottom as you do now and then you can add specific considerations for different exceptions in the catch if you prefer. Then each catch should return an HTTP status code, e.g. NoContent(), OK(), Forbidden() or others but don't throw the error to the user. Instead log the error in the controller and give the user either a 500 or 502 etc.

backend/api/Controllers/FeedbackController.cs
```diff
@@ -0,0 +1,47 @@
...
+                await teamsNotificationService.SendTeamsMessageAsync(
+                    feedbackQuery.ToString(),
+                    TeamsDestination.Feedback
+                );
+                return NoContent();
+            }
+            catch (Exception e)
+            {
+                logger.LogError(e, "Error during POST of feedback");
+                throw;
```
aeshub: See other comment but avoid this throw. Instead do a return with status code and just log the error in the program. That way the error details are not revealed to the user but still logged by the program. The pattern is as follows:

```c#
    try
    {
        DoSomething();
    }
    catch(Exception e)
    {
         LogException(e);
         return StatusCode(500);
    }
```

backend/api/Program.cs
```diff
@@ -94,6 +94,8 @@
 builder.Services.AddScoped<ITaskDurationService, TaskDurationService>();
 builder.Services.AddScoped<ILastMissionRunService, LastMissionRunService>();
 
+builder.Services.AddHttpClient<ITeamsNotificationService, TeamsNotificationService>();
```
aeshub: **`AddHttpClient<T>` registers a transient typed client, but `TeamsMessageEventHandler` is a singleton**

`Program.cs:97` registers `TeamsNotificationService` as a typed client:

```csharp
builder.Services.AddHttpClient<ITeamsNotificationService, TeamsNotificationService>();
```

Typed clients are **transient** by default. `TeamsMessageEventHandler` is registered via `AddHostedService` (`Program.cs:113`), so it resolves one instance and holds it — along with its `HttpClient` and `HttpMessageHandler` — for the entire process lifetime. That defeats the factory's handler rotation, so the connection never picks up DNS changes. Relevant here because Teams webhook endpoints sit behind rotating Azure front ends; the symptom would be intermittent delivery failures weeks apart, which is unpleasant to diagnose.

We already have the correct pattern for this in `GenericOidcAuthorizationHeaderProvider.cs:19` — a singleton that injects `IHttpClientFactory` and calls `CreateClient()` per request.

Suggested fix:

```diff
-builder.Services.AddHttpClient<ITeamsNotificationService, TeamsNotificationService>();
+builder.Services.AddHttpClient();
+builder.Services.AddSingleton<ITeamsNotificationService, TeamsNotificationService>();
```

```diff
 public class TeamsNotificationService(
-    HttpClient httpClient,
+    IHttpClientFactory httpClientFactory,
     IConfiguration configuration,
```

```diff
-        using var response = await httpClient.SendAsync(request, cancellationToken);
+        using var client = httpClientFactory.CreateClient();
+        using var response = await client.SendAsync(request, cancellationToken);
```

Both consumers keep injecting `ITeamsNotificationService` unchanged, and `MockTeamsNotificationService` is already registered `AddSingleton` in `TestWebApplicationFactory.cs:65`, so tests are unaffected apart from a one-line stub factory in `TeamsNotificationServiceTests.CreateService`.

Bonus: as a singleton you can also cache the parsed `AdaptiveCardTemplate` in a field instead of doing `File.ReadAllText` on every message.

backend/api/Controllers/Models/FeedbackQuery.cs
```diff
@@ -0,0 +1,52 @@
...
+        [Required]
+        [MaxLength(3000)]
+        public string Description { get; set; }
+        public required FeedbackType Type { get; set; }
+
+        [Required]
+        public DateTime Timestamp { get; set; }
+
+        [Required]
+        public string Url { get; set; }
```
aeshub, line 30: I suggest also adding a max length to the URL. 200 for example should be fine. Also run the url through sanitization if there is no issue with that.

aeshub (changes requested):
Have a look at the comments provided and reach out if you have questions.

backend/api/Utilities/TeamsAdaptiveCard.json
markusmoen: Short answer is 1.6 did not work and microsoft uses 1.2 in their examples [1](https://learn.microsoft.com/en-us/connectors/teams/?tabs=text1%2Cdotnet#microsoft-teams-webhook), [2](https://learn.microsoft.com/en-us/microsoftteams/platform/webhooks-and-connectors/how-to/connectors-using?tabs=cURL%2Ctext1)

1.6 gave this result 
image

Digging a bit deeper I found [this](https://learn.microsoft.com/en-us/adaptive-cards/resources/partners), suggesting 1.5 should work. Tested locally and 1.5 works, so still requires a downgrade but not that far

backend/api/Controllers/Models/FeedbackQuery.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        public override string ToString()
+        {
+            return string.Join(
+                Environment.NewLine,
+                $"Type: {Type}",
+                $"Title: {Title}",
+                $"Description: {Description}",
+                $"Email: {Email}",
+                $"Short name: {ShortName}",
```
andchiind, line 40: Do I understand correctly that this is the user info? If so, could we not include them in the ToString, just in case the text is logged? GDPR does not let us store user data without the ability to delete it upon request, so since the logs are stored we should avoid logging user info such as emails and short names.

backend/api/Controllers/FeedbackController.cs
```diff
@@ -0,0 +1,44 @@
...
+    ) : ControllerBase
+    {
+        /// <summary>
+        /// Submit feedback to the Flotilla team
+        /// </summary>
+        /// <remarks>
+        /// <para> This query submits feedback to the Flotilla team </para>
+        /// </remarks>
+        [HttpPost]
+        [Authorize(Roles = Role.Any)] //TODO: which roles to use
```
andchiind: It can be good to remove the TODOs before merging, although the most important thing is that they're not in the final commit.

backend/api/Controllers/Models/FeedbackQuery.cs
andchiind: Hmm nevermind I see that you only use this function when sending the Teams message. It's probably fine then as long as we're careful

backend/api/Controllers/FeedbackController.cs
markusmoen: Makes sense, we might want to do the same for some of the other controllers at some point

aeshub (approved):
LGTM

Feel free to add an issue if you saw other locations with the same throw pattern in controllers. You can add the good first issue label to that one.

PR #2907, opened 2026-08-25
Make data view see earlier data but not in the table

nicholasdalhaug:


frontend/src/pages/DataViewPage/DataView.tsx
```diff
@@ -77,29 +77,28 @@ const DataViewContent = ({
...
             if (!Object.hasOwn(plotData, tagId)) plotData[tagId] = []
             plotData[tagId].push({
                 time: sampleTimestamp,
                 value: parseFloat(inspection.value),
                 inspectionId: inspection.inspectionId,
             })
-            inspectionLookup.set(inspection.inspectionId, inspection)
         })
         return plotData
     }, [inspectionData, selectedInspectionId])
```
aeshub:
```suggestion
    }, [inspectionData, selectedInspection])
```
Should it be this instead now that the body no longer reads selectedInspectionId? It should still work as it is but the moment some other input is added it will be wrong.

nicholasdalhaug: Yes. Good catch!

frontend/src/pages/DataViewPage/DataViewTable.tsx
```diff
@@ -31,6 +32,18 @@ export const DataViewTable = ({
...
+    // Assumes it is already sorted
+    const uniqueTagInspectionData = useMemo(() => {
+        return inspectionData.reduce((acc: InspectionData[], current) => {
+            const existingIndex = acc.findIndex((item) => item.tag === current.tag)
+            if (existingIndex === -1) {
+                acc.push(current)
+            }
+            return acc
+        }, [])
+    }, [inspectionData])
```
aeshub:
```suggestion
  const uniqueTagInspectionData = useMemo(() => {
      const seenTags = new Set<string>()
      return inspectionData.filter((inspection) => {
          if (seenTags.has(inspection.tag)) return false
          seenTags.add(inspection.tag)
          return true
      })
  }, [inspectionData])
```
Suggested change. I find it slightly more readable and it should be linear cost instead of quadratic although that doesn't matter that much here but if we use the same solution for larger sets of data later it might.

nicholasdalhaug: We should not edit an external variable inside filter, but a forEach could work.

frontend/src/pages/DataViewPage/DataViewTable.tsx
```diff
@@ -31,6 +32,20 @@ export const DataViewTable = ({
 }) => {
     const { TranslateText } = useLanguageContext()
 
+    // The table shall only show one line per tag
+    // Assumes it is already sorted
+    const uniqueTagInspectionData = useMemo(() => {
```
aeshub: ### The `#` column no longer matches the map markers

The dedup lives inside `DataViewTable`, but the map at `DataView.tsx:126` still receives the full `inspectionData`. Both label entries with `index + 1` over whatever array they're given:

- table `#` chip — `DataViewTable.tsx:66`
- map tooltip — `PointillaMapMarkers.tsx:55`

On `main` these iterate the same array, so marker `3` is row `3`. Now the table iterates the deduped array and the map the full one — with 5 tags × 20 samples the table shows 1–5 while the map shows 1–100, so row 2 is marker 21. The `#` column goes from redundant to misleading.

Related: `getInspectionMarkers` creates one `circleMarker` per record with no dedup or clustering, so every historical sample for a tag stacks on the same coordinate (compounding `0.8`-opacity fills, plus N permanent tooltips printing different numbers on one pixel). Pre-existing, but deduping the map input fixes it for free. `InspectionOverviewSection` at `DataView.tsx:189` has the same issue — it renders every historical image per tag.

### Suggested fix

Move the `uniqueTagInspectionData` block as-is from `DataViewTable` up into `DataViewContent`, so all three consumers share one array:

```tsx
// DataView.tsx — unchanged logic, just hoisted
const uniqueTagInspectionData = useMemo(() => {
    const uniqueTags = new Set<string>()
    const uniqueInspectionData: InspectionData[] = []
    inspectionData.forEach((inspection) => {
        if (!uniqueTags.has(inspection.tag)) {
            uniqueTags.add(inspection.tag)
            uniqueInspectionData.push(inspection)
        }
    })
    return uniqueInspectionData
}, [inspectionData])
```

then pass it to all three:

```tsx
<DataViewTable inspectionData={uniqueTagInspectionData} ... />
<InspectionsPlantMap ... inspections={uniqueTagInspectionData} />
<InspectionOverviewSection inspectionData={uniqueTagInspectionData} />
```

`linePlotData` should keep using the full `inspectionData` — it needs the whole series. This also puts the dedup next to the "assumes it is already sorted" assumption rather than hiding it in a leaf component.

After the move, `DataViewTable.tsx` no longer needs its `useMemo` import (`InspectionData` is still used by the props type).

nicholasdalhaug: I agree. Done

andchiind (approved):

PR #2908, opened 2026-08-25
Let the broker take its credentials at runtime

oysand:
## Why

The broker's identity is baked into the image: one committed `passwd_file` of real hashed passwords, and one committed server certificate whose CN forces the network alias `broker`.

Two consequences:

- **Every environment shares one set of credentials.** Dev, staging and production authenticate with the same passwords against the same certificate.
- **The armada integration tests need production secrets.** They are handed the plaintext MQTT passwords and the broker's TLS private key from `FlotillaTestsKv` purely so they can talk to a broker built from this image.

## What

`entrypoint.sh` assembles the password file and the certificates into a writable `mosquitto/config/runtime/` on startup, from the environment:

| Variable | Secret | Meaning |
| --- | --- | --- |
| `MQTT_PASSWORDS` | yes | Comma separated `user:password` pairs, hashed into a fresh `passwd_file` with `mosquitto_passwd` |
| `TLS_SERVER_KEY` | yes | PEM or the bare base64 body, as the key vault holds it today |
| `TLS_SERVER_CERT` | no | Server certificate as PEM |
| `TLS_CA_CERT` | no | CA certificate as PEM |

**Everything falls back to the committed material when unset**, so existing deployments are unaffected and can be migrated one environment at a time.

`scripts/generate-mqtt-credentials.sh` mints a per-environment CA, server certificate and a random password per ACL user, and prints which outputs are secret.

`access_control` stays committed deliberately — it is policy, not a secret, and it is what the tests assert on.

Side effect: the broker is no longer pinned to the hostname `broker`, since the certificate is issued for whatever `generate-mqtt-credentials.sh` is given.

The local stacks use this rather than the fallback, so the new path is exercised every day rather than only in CI. `tilt/preflight.py` generates the same `broker/.local-credentials` and `broker/.env` that `setup.sh` does — reusing them until the certificate is close to expiry — and `tilt/docker-compose.broker.yml` feeds them in with `MQTT_REQUIRE_ENV_CREDENTIALS=true`, so a broker that did not receive them fails to start instead of quietly running on the credentials committed here. `tilt/flotilla.tilt` exports `mqtt_password()` and `mqtt_ca_cert()` for callers that run their own MQTT clients against this broker, and no longer resolves `flotilla-mqtt-password` from Key Vault. The robotics local orchestration consumes those exports in equinor/robotics#59.

## Verified

Built the image and exercised every path:

- Fully generated CA, certificate and passwords: TLS handshake succeeds with hostname verification against the generated CA.
- ACLs still enforced — `Denied PUBLISH` for `sara` → `isar/x/status`, and `disconnected, not authorised` on subscribe to `isar/#`.
- Wrong password → `Connection Refused: not authorised`.
- Legacy path: bare-body `TLS_SERVER_KEY`, no `MQTT_PASSWORDS`, no certs → falls back to the committed material and starts.
- `MQTT_ALLOW_INSECURE_LOCAL_CONNECTIONS=true` with no TLS material at all → starts, auth still enforced.
- Full armada integration suite (4 tests, `-n auto`) against this broker with generated credentials and no `.env` — 4 passed. Re-run after the rebase onto current `main`, against locally built images: 4 passed.

And for the Tilt route:

- `uv run --script tilt/preflight.py` generates on the first run and reports reuse on the second; `broker/.env` is `0600` and its format matches `setup.sh` byte for byte.
- The broker starts on `mosquitto.conf` reporting all four credentials from the environment, with **no fallback warning**, both from this repo's Tiltfile and from the robotics stack's two-file compose merge — the case where Compose resolves paths against the *other* repo's directory.
- `tilt up` here: the backend connects as `flotilla` with the generated password and `Mqtt__AllowInsecureLocalConnections=false`.
- End to end in the robotics stack (equinor/robotics#59): all four clients — `flotilla`, `sara` and two `isar` — authenticated over TLS, a full inspection mission completed, 3156 publishes with zero denials.
- ISAR verifies the broker against the generated CA passed as `ISAR_MQTT_CA_CERT`, and fails with `CERTIFICATE_VERIFY_FAILED` without it.

## Follow-up

Once every environment has its own credentials in its own key vault, the committed `passwd_file` and `certs/` can be deleted and the fallback removed. Tracked in `robotics-infrastructure`. Local development no longer uses them either after this PR, so the remaining consumers are the deployed environments alone.

  - There is no test harness for the broker image; verified manually as listed above, and end to end by the armada integration suite.
  - No issue exists; this is part of the work to remove external secrets from the integration tests.
  - Per-environment rollout is `robotics-infrastructure` work, noted under Follow-up.

## Merge order

This is one of five PRs that remove the need for any external secret to run the armada
integration tests. They must merge in this order:

1. equinor/flotilla#2908 — broker takes its credentials at runtime
   equinor/isar#1166 — MQTT CA certificate configurable
   *then wait for the `:dev` images to publish*
1b. equinor/robotics#59 — point the local stack at the generated credentials (needs 1; blocks nothing)
2. equinor/armada#100 — generate MQTT credentials per test run
3. equinor/flotilla#2918, equinor/isar-robot#402, equinor/sara#488 — stop passing the unused secret
4. armada follow-up — drop the now-unused secret declaration, retire the key vault

Out of order it breaks: armada's reusable workflow currently declares
`INTEGRATION_TEST_AZURE_CLIENT_SECRET` as `required: true` and every caller uses `@main`, so a
caller that stops passing it before step 2 fails immediately. Conversely step 2 needs the new
broker and the configurable ISAR CA already published, because it feeds the broker its
credentials through the environment.

Follow-up rollout is tracked in equinor/robotics-infrastructure#1169.


oysand:
Pushed `368910fe`, from a review of this branch. The fallback was the weak point: every credential
degraded to the material committed here, and only `MQTT_PASSWORDS` said so. An environment could
believe it had rotated when it had not.

- Each credential now logs its source, and `MQTT_REQUIRE_ENV_CREDENTIALS=true` makes any fallback
  a startup failure. All three deployed environments will set it (equinor/robotics-infrastructure#1169),
  so they cannot quietly run on the shared credentials. It is rejected together with
  `MQTT_ALLOW_INSECURE_LOCAL_CONNECTIONS`, which would be a contradiction.
- `MQTT_PASSWORDS` is checked against `access_control`. A user missing a password simply could not
  authenticate, and mosquitto does not explain why.
- A malformed `user:password` pair now aborts startup. The loop no longer runs in a pipeline
  subshell, where its `exit 1` could not stop the script.
- `generate-mqtt-credentials.sh` takes several hostnames. SARA connects on
  `broker.robotics.svc.cluster.local` while ISAR uses `broker`, so the single-SAN version would
  have locked SARA out at the first rotation. It also stops assigning `HOSTNAME`, prints the
  `az keyvault secret set` commands, and generates passwords that cannot contain the
  `MQTT_PASSWORDS` separator.
- `setup.sh` generates throwaway local credentials instead of prompting for the key vault's broker
  key, so a laptop no longer holds a key that a deployed environment uses too. `broker/.env` keeps
  single-line PEM bodies because Compose's `env_file` cannot hold a multi-line value.

Verified against `eclipse-mosquitto:2.0.22`: each failure path produces its intended message, and a
generated credential set completes a TLS handshake with `mosquitto_pub --cafile`, both as full PEM
and as the single-line body `setup.sh` writes.

aeshub (approved):

PR #2912, opened 2026-08-27
Remove unused stop ID feature

andchiind:
This was anyways unused, and it is annoying to make use of anyways due to the structure in the frontend. Deleting it also simplifies the logic in ISAR quite a bit. We can re-add it if we see that it is useful, but since nobody has noticed that it wasn't being used until now then I doubt that. It was more useful back when return home missions were treated as normal missions, because back then it was easy to accidentally stop a return home when it just started.

See https://github.com/equinor/isar/pull/1168 for info on the ISAR change.

- [ ] A test has been written


Christdej (approved):
LGTM

PR #2919, opened 2026-08-27
Change task table with only valid interactions and showing value

nicholasdalhaug:


nicholasdalhaug:
Depends on https://github.com/equinor/sara/pull/498

nicholasdalhaug:
The task table now looks like this:
image

Note that fencilla will still have a button even without a detection. That should be fixed in an other PR.

frontend/src/components/Displays/TaskDisplay.tsx
```diff
@@ -1,4 +1,26 @@
...
+export const AnalysisValueDisplay = ({
+    value,
+    unit,
+    analysisType,
+}: {
+    value: string
+    unit?: string
+    analysisType?: string
+}) => {
+    if (saraAnalysisTypeToEnum(analysisType) === AnalysisType.CLOE) {
```
andchiind, line 14: In the future it would be nice to do this conversion in SARA instead, but we can leave that for another PR

andchiind, line 14: It seems a little odd to hardcode this. Was there really a problem with SARA providing a percentage before?

frontend/src/pages/DataViewPage/DataView.tsx
```diff
@@ -89,9 +90,14 @@ const DataViewContent = ({
             if (inspection.value == null || inspection.value === '' || !sampleTimestamp) return
             if (selectedInspection && tagId !== selectedInspection.tag) return
             if (!Object.hasOwn(plotData, tagId)) plotData[tagId] = []
+
+            const value: number =
+                saraAnalysisTypeToEnum(inspection.analysisType) === AnalysisType.CLOE
```
andchiind, line 95: This change where SARA provides a ratio instead of a percentage seems to add quite a bit of hardcoding on the flotilla side

frontend/src/pages/MissionPage/TaskOverview/TaskTable.tsx
```diff
@@ -43,7 +49,7 @@ export const TaskTable = ({ tasksAndData }: TaskTableProps) => {
                 {tasksAndData &&
                     tasksAndData.map((taskAndData, index) => (
                         <TaskTableRow
-                            key={taskAndData.task.id + 'tasktable'}
+                            key={taskAndData.task.id}
```
andchiind, line 52: What was the point of deleting the string?

nicholasdalhaug: To remove the confusion of why it was there.

frontend/src/pages/DataViewPage/DataView.tsx
nicholasdalhaug: It is in general best to let the api return data and not how data should be visualized. Yes, it adds a small bit of code in the frontend to visualize data, but that is an important job of the frontend, not the backend.

frontend/src/components/Displays/TaskDisplay.tsx
nicholasdalhaug: This allows us to also add functionality for how to display fencilla differently than a string "false". 
It is now closer to sara returning data and not how to visualize the data. Cleaner split of responsibility.

andchiind (approved):

PR #2923, opened 2026-08-28
[do not merge] bisect: is the deploy startup_failure repo-specific?

Christdej:
Temporary. Points flotilla's dev build job at the same armada variant (`bisect/b-env-nonempty`, which adds only a job-level `environment: Development` on top of 7aec39a) that reliably produces `startup_failure` on isar-anymal.

If flotilla's Deploy to Development also fails, the trigger is the `environment` key itself. If it passes, the trigger is something specific to the isar repositories, which would explain why eight other callers are unaffected on `@main`.

Tracking: equinor/armada#112. Will close without merging.

### DIFF
--- a/.github/workflows/deploy_to_development.yml
+++ b/.github/workflows/deploy_to_development.yml
@@ -34,7 +34,7 @@ jobs:
     strategy:
       matrix:
         component: [broker, backend, frontend]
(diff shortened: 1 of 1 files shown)


Christdej:
Closing unmerged — a pull request does not trigger `Deploy to Development`, which runs on push to main, so this cannot produce the comparison it was opened for. Pushing an experiment to flotilla's main is not worth the risk. Findings recorded in equinor/armada#112.

PR #2925, opened 2026-08-28
Add formulation for fence detections

nicholasdalhaug:


Eddasol (approved):
LGTM

Would prefer finding instead of breach as it looks for more than just holes

PR #2934, opened 2026-08-31
Standardize controller exception responses

### LINKED ISSUE #2920: Standardize exception handling in controllers with explicit status codes
Currently the controllers use a mix of
```csharp
catch (Exception e)
{
    logger.LogError(e, "Error during GET of ...");
    throw;
}
```
and 
```csharp
catch (Exception e)
{
    logger.LogError(e, "Error during POST of feedback");
    return StatusCode(StatusCodes.Status500InternalServerError);
}
```

The second pattern is preferred since it logs only once, is predictable for the frontend and does not leak any internal information to the client. 

Replace throw with an explicit return StatusCode(...) across all controllers. Consider wether specific exception types should be caught before the generic catch (Exception) to give another statuscode than 500. 

ripleyforge-chase:
Closes #2920

## Problem

Several controllers catch Exception, log it, and then rethrow it. This allows unexpected exceptions to escape the controller after they have already been logged, producing inconsistent API behavior and potentially passing internal failure details farther through the exception pipeline.

The repository already uses explicit, bodyless 500 responses in other controllers. This change applies that convention consistently.

## Changes

- Replace 24 generic controller rethrows with explicit Status500InternalServerError responses.
- Preserve existing specific exception handling.
- Return bodyless 404 responses for known missing installation or plant dependencies on create paths.
- Return bodyless 409 responses when an inspection or exclusion area already exists.
- Change the missing-installation case in RobotService.CreateFromQuery from DbUpdateException to the existing InstallationNotFoundException.
- Add HTTP-level regression coverage using the existing xUnit, WebApplicationFactory, and Testcontainers test style.

## Status-code decisions

Only exceptions with stable domain semantics receive a non-500 response:

- InstallationNotFoundException and PlantNotFoundException: 404
- InspectionAreaExistsException and ExclusionAreaExistsException: 409
- Existing InvalidPolygonException handling remains 400
- Unknown and infrastructure exceptions remain an opaque 500

Framework exceptions are not mapped generically because their meaning depends on where they originated.

## Threat model

This reduces the Information Disclosure risk at the API boundary. Unexpected exceptions are logged server-side and clients receive bodyless error responses without exception messages, stack traces, database details, or dependency information.

The more specific 404 and 409 responses are limited to named domain exceptions on Admin-authorized create endpoints. Their responses are bodyless, so they do not introduce additional internal details or a new unauthenticated resource oracle.

The change does not add trust boundaries, data flows, external connectivity, or storage, and it does not change authentication or authorization behavior. Server-side error logging is preserved for operational detection.

## Verification

- Added a regression test proving an unexpected service exception returns an opaque HTTP 500 response instead of escaping the request.
- Added a regression test proving robot creation against an unknown installation returns HTTP 404.
- Focused regression tests: 2 passed, 0 warnings.
- Full backend suite: 115 passed, 0 warnings.
- CSharpier formatting applied using the repository-pinned version.
- Strict secret scan passed for all changed files.
- Independent correctness and security review found no actionable issues.

## Ready for review checklist

- [x] A test has been written
  - [ ] This change does not need a new test
  - [x] There is no remaining work from this PR that requires new issues
- [x] The changes do not introduce dead code such as unused imports or functions

### DIFF
--- a/backend/api.test/Controllers/RobotControllerTests.cs
+++ b/backend/api.test/Controllers/RobotControllerTests.cs
@@ -7,8 +7,10 @@ using System.Text.Json;
 using System.Threading.Tasks;
 using Api.Controllers.Models;
 using Api.Database.Models;
+using Api.Services;
 using Api.Test.Database;
 using Microsoft.Extensions.DependencyInjection;
+using Moq;
 using Testcontainers.PostgreSql;
 using Xunit;
 
@@ -20,11 +22,13 @@ namespace Api.Test.Controllers
         public required PostgreSqlContainer Container;
         public required HttpClient Client;
         public required JsonSerializerOptions SerializerOptions;
+        public required string PostgreSqlConnectionString;
 
         public async ValueTask InitializeAsync()
         {
             (Container, string connectionString, var connection) =
                 await TestSetupHelpers.ConfigurePostgreSqlDatabase();
+            PostgreSqlConnectionString = connectionString;
             var factory = TestSetupHelpers.ConfigureWebApplicationFactory(
                 postgreSqlConnectionString: connectionString
             );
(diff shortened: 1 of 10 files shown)


backend/api.test/Controllers/RobotControllerTests.cs
```diff
@@ -91,5 +95,60 @@ public async Task CheckThatReadRobotByIdEndpointIsSuccessful()
...
+            // Arrange
+            const string ErrorMessage = "Internal database details";
+            var robotService = new Mock<IRobotService>();
+            robotService
+                .Setup(service => service.ReadAll(It.IsAny<bool>()))
+                .ThrowsAsync(new InvalidOperationException(ErrorMessage));
+            using var factory = TestSetupHelpers.ConfigureWebApplicationFactory(
+                PostgreSqlConnectionString,
+                services => services.AddScoped(_ => robotService.Object)
+            );
```
andchiind, line 111: What is the point of redefining factory here instead of using the one already defined on line 28?

backend/api.test/Controllers/RobotControllerTests.cs
```diff
@@ -91,5 +95,60 @@ public async Task CheckThatReadRobotByIdEndpointIsSuccessful()
...
+            using var client = TestSetupHelpers.ConfigureHttpClient(factory);
```
andchiind, line 112: Likewise here

PR #2935, opened 2026-09-01
Await access role lookups instead of blocking on .Result

Christdej:
## Why

Every installation-scoped query blocked a thread-pool thread on the access-role lookup. Introduced in `d32ed017` (2023-11-23) and copy-pasted into **11 sites across 7 services**.

| Service | Sites |
|---|---|
| RobotService | 5 |
| InstallationService, MissionRunService, MissionDefinitionService, PlantService, InspectionAreaService, ExclusionAreaService | 1 each |

Two shapes. Most started the lookup without awaiting and put `.Result` inside the LINQ expression tree, so EF evaluated it during parameter extraction — confirmed by stack trace, not assumed:

```
Task`1.GetResultCore                      <- .Result blocks the request thread
ExpressionTreeFuncletizer.Evaluate
ExpressionTreeFuncletizer.ExtractParameters
QueryCompiler.ExecuteCore(query, async: true)
EntityQueryable`1.GetAsyncEnumerator
ToListAsync
```

The other three (`RobotService:361`, `ExclusionAreaService:283`, `InspectionAreaService:299`) blocked eagerly. Both hold a thread across a database round-trip.

## Why it went unnoticed for two years

It never throws. A single request works perfectly — I verified this before assuming otherwise. The cost only appears under concurrency, and it compounds: slower database responses hold threads longer, which starves the pool, which queues more requests. It turns a *degraded* database into an *unavailable* backend rather than a slow one.

Worth being explicit: **this did not cause the recent production outage**, and there is no evidence it has caused a user-visible incident. It is a fragility multiplier, not an active fire.

## Measured

Standalone EF Core reproduction, 150ms simulated round-trip, **default** thread pool (no artificial constraint):

| | Before | After |
|---|---|---|
| 200 concurrent requests | **30,830ms** | **169ms** |
| Worker threads held during I/O | 2 | 0 |
| Single request | works | works |

For context, the prod backend runs 1 replica with no CPU limit on an 8-core node, so `ThreadPool` min threads is 8. At 2 threads per request, roughly 4 concurrent requests exhaust the min pool; beyond that .NET injects threads at ~1–2/sec.

## The change

Await the lookup before composing the query. The private query builders become `async Task<IQueryable<T>>`, so call sites read `(await GetInstallations()).ToListAsync()`. Noisier at the call site, but the blocking is gone and **filtering is unchanged** — the same installation codes reach the same `Where` clause.

`TryFindInspectionAreaForMissionTasks` was synchronous while doing database work, so it becomes async; its single caller in `MissionDefinitionController` is updated. No other public signatures change, and all query builders are private with no callers outside their own file.

## Verification

- `dotnet build -warnaserror` — 0 warnings, 0 errors.
- `dotnet csharpier check .` — clean.
- `dotnet test` — **113/113 pass**.
- Compiler-driven: making the builders async turned all 48 call sites into compile errors, so none could be silently missed.

## Deliberately out of scope

- `MqttService.cs:226` also blocks (`.Wait()`), but on a broker subscribe rather than the request path. Separate change.
- The two `GetAwaiter().GetResult()` calls in `CustomServiceConfigurations` are startup-path and correct as they are.

### DIFF
--- a/backend/api/Controllers/MissionDefinitionController.cs
+++ b/backend/api/Controllers/MissionDefinitionController.cs
@@ -189,7 +189,7 @@ namespace Api.Controllers
             try
             {
                 var inspectionAreaForMission =
-                    inspectionAreaService.TryFindInspectionAreaForMissionTasks(
+                    await inspectionAreaService.TryFindInspectionAreaForMissionTasks(
                         missionTasks,
                         customMissionQuery.InstallationCode
                     );
--- a/backend/api/Services/ExclusionAreaService.cs
+++ b/backend/api/Services/ExclusionAreaService.cs
@@ -72,13 +72,14 @@ namespace Api.Services
     {
         public async Task<IEnumerable<ExclusionArea>> ReadAll(bool readOnly = true)
         {
-            return await GetExclusionAreas(readOnly: readOnly).ToListAsync();
+            var query = await GetExclusionAreas(readOnly: readOnly);
+            return await query.ToListAsync();
         }
 
         public async Task<ExclusionArea?> ReadById(string id, bool readOnly = true)
         {
-            return await GetExclusionAreas(readOnly: readOnly)
-                .FirstOrDefaultAsync(a => a.Id.Equals(id));
+            var query = await GetExclusionAreas(readOnly: readOnly);
+            return await query.FirstOrDefaultAsync(a => a.Id.Equals(id));
         }
 
         public async Task<IEnumerable<ExclusionArea>> ReadByInstallationCode(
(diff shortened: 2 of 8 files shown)


Christdej:
Pushed `ae5194a7` addressing the nested-await readability, since `await (await ...)` reads badly even though it is correct.

To be explicit about what the two awaits were: they are two different operations, not the same one awaited twice. The inner resolves the access-role lookup, the outer runs the query. Both round-trips happened before this PR too — the only change is that the first one no longer blocks a thread-pool thread while it waits.

Call sites now bind the query to a local:

```csharp
var query = await GetInstallations(readOnly: readOnly);
return await query.ToListAsync();
```

41 sites hoisted, no behaviour change. One name collision in `MissionDefinitionService.ReadAll` was caught by the compiler and fixed by hand.

**Deliberately not changed:** the query builders still return `Task<IQueryable<T>>` rather than taking the permitted installation codes as a parameter. The parameter version has tidier types, but it moves the access filter out to ~50 call sites where forgetting it would silently widen access rather than fail to compile. For a security control I'd rather keep the slightly awkward return type and make it impossible to forget. Happy to revisit if reviewers disagree.

Re-verified: `dotnet build -warnaserror` clean, `csharpier check` clean, **113/113 tests pass**.

PR #2940, opened 2026-09-02
Do not add aborted missions to the queue

andchiind:
This PR makes it so that we don't restart missions after leaving maintenance mode.

This ISAR PR makes it so that we don't abort missions when going to lockdown: https://github.com/equinor/isar/pull/1174

Previously, when adding aborted missions back to the queue we would run the risk of users deleting them from the queue. This would make the mission deprecated, thereby not allowing the user to see the mission anymore, including the data it would already have gathered at that point.

The PR also cleans up unused code.

- [ ] A test has been written


backend/api/Services/MissionSchedulingService.cs
```diff
@@ -257,30 +255,8 @@ public async Task<MissionRun> MoveMissionRunBackToQueue(
...
-                missionRun.Id,
-                "Status",
-                MissionStatus.Queued
-            );
-            _ = signalRService.SendMessageAsync(
-                "Mission run created",
-                missionRun.InspectionArea.Installation,
-                new MissionRunResponse(missionRun)
-            );
+            await AbortMissionRun(missionRun, stopReason ?? "");
```
Christdej, line 259: Abort mission doesnt not emit message, so frontend might not be updated when ISAR aborts a mission or when ISAR restarts mid-mission.

Christdej (changes requested):

backend/api/Services/MissionSchedulingService.cs
andchiind: It does emit a message. It's inside the Update function, inside UpdateMissionRunProperty, inside AbortMissionRun.

Christdej: Yes you are right, I didnt notice that, all good then

Christdej (approved):
LGTM

PR #2943, opened 2026-09-02
Avoid extra database call on robot info message

andchiind:
We don't need to fetch the database in most cases, so this waits to make that database call until we know it's needed. This event is expected every 30 seconds for each robot.

- [ ] A test has been written


backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -266,8 +263,30 @@ private async void OnIsarRobotInfo(IsarRobotInfoMessage isarRobotInfo)
 
                 UpdatePortIfChanged(isarRobotInfo.Port, ref robot, ref updatedFields);
 
-                if (isarRobotInfo.CurrentInstallation is not null)
+                if (
+                    isarRobotInfo.CurrentInstallation is not null
+                    && robot.CurrentInstallation.InstallationCode
```
Christdej: Could be worth comparing installation code on "ToLower" or similar

Christdej (approved):
LGTM

PR #2948, opened 2026-09-02
Fix task numbering for inspection and analysis report

### LINKED ISSUE #2916: Task number in inspection view is wrong
**Describe the bug**
The task number in header of inspection view is wrong

**To Reproduce**
1. Go to mission page for mission with >1 inspection
2. Click on task to open inspection view
3. Note number in header of the dialog

**Expected behavior**

**Screenshots**
The mission tasks:
Image

The inspection view of task 1:
Image

K-S-Bakken:
- [ ] A test has been written

Closes #2916. 
It was previously using the table index from sara not the mission task table in flotilla.


frontend/src/pages/MissionPage/AnalysisResultView.tsx
```diff
@@ -20,9 +21,10 @@ import { InspectionData } from 'models/InspectionRecord'
 import { AnalysisFeedback } from './AnalysisFeedback'
 import { AnalysisValueDisplay } from 'components/Displays/TaskDisplay'
 
-interface InspectionDialogViewProps {
+interface AnalysisResultDialogViewProps {
     selectedInspectionId: string
     inspectionData: InspectionData[]
+    tasks: Task[]
```
marianards, line 27:
```suggestion
    tasks?: Task[]
```

marianards: Making this optional keeps DataView.tsx's existing call (which has no task list to pass) compiling, instead of breaking it (:

frontend/src/pages/InspectionReportPage/InspectionView.tsx
```diff
@@ -22,16 +23,22 @@ import { InspectionData } from 'models/InspectionRecord'
 interface InspectionDialogViewProps {
     selectedInspectionId: string
     inspectionData: InspectionData[]
+    tasks: Task[]
```
marianards, line 26:
```suggestion
    tasks?: Task[]
```

marianards: Same comment as before (:

frontend/src/pages/MissionPage/AnalysisResultView.tsx
```diff
@@ -128,7 +135,7 @@ export const AnalysisResultDialogView = ({ selectedInspectionId, inspectionData
             <StyledDialogContent>
                 <StyledDialogHeader>
                     <Typography variant="accordion_header" group="ui">
-                        {TranslateText('Analysis result for task') + ' ' + (inspectionIndex + 1)}
+                        {TranslateText('Analysis result for task') + ' ' + (taskNumber + 1)}
```
marianards, line 138: tasks.findIndex(...) will return -1 if the id isnt in tasks (which will since tasks is now optional), so we need to handle both cases where we have taskNumber or not


```suggestion
                        {TranslateText('Analysis result for task') + (taskNumber !== -1 ? ' ' + (taskIndex + 1) : '')}
```

frontend/src/pages/InspectionReportPage/InspectionView.tsx
```diff
@@ -78,7 +85,7 @@ export const InspectionTaskDialogView = ({ selectedInspectionId, inspectionData
             <StyledDialogContent>
                 <StyledDialogHeader>
                     <Typography variant="accordion_header" group="ui">
-                        {TranslateText('Inspection report for task') + ' ' + (inspectionIndex + 1)}
+                        {TranslateText('Inspection report for task') + ' ' + (taskNumber + 1)}
```
marianards, line 88: Same as the other comment (:

frontend/src/pages/InspectionReportPage/InspectionView.tsx
marianards, line 145: Don't know if this is out of scope for this PR but this one is still using inspectionIndex instad of taskNumber (:

frontend/src/pages/InspectionReportPage/InspectionView.tsx
```diff
@@ -22,16 +23,22 @@ import { InspectionData } from 'models/InspectionRecord'
 interface InspectionDialogViewProps {
```
marianards, line 23:
```suggestion
interface InspectionTaskDialogViewProps {
```

Minor detail but this is the props for the InspectionTaskDialogView component so we should have the same name, i think this might be from an old PR but would be nice if you could fix it in here too 🫶🏻

marianards (approved):
LGTM :)

PR #2950, opened 2026-09-03
Separate banners and notifications

### LINKED ISSUE #1836: Improve notification and banner feature
Currently notification (the bell in the header) and banner features are duplicates of each other. Suggested new solution is that banner will handle information that is relevant for the user while using flotilla, like if a mission fails to get scheduled. Notifications should be use for things that occur when the user might not be actively using flotilla that is use full to know when checking at a later time.

Notification should:
* Have timestamp and show newest first
* Should have these kind of notifications:
  * Failed missions
  * Aborted mission. Eg. If safe zone gets activated during mission or robot health (like low battery) causing the mission to be aborted

Banner should:
* Only be used for things that happened now because of something the user did or something that is actively happening 
* Should either: Not overwrite each other OR have such generic text that it doesn't matter that they overwrite each other
  * Example: If failing to schedule 2 missions only the last to fail will get banner as the first gets overwritten. Either have several or a generic enough message 
* Should have these kind of banners:
  * "Send to dock" being active and robots being released from dock
  * Failing to schedule a mission
  * Failure to fetch requested info (like echo missions or should this always be handled as a popup or something instead?)

Currently we have these alert types:
    MissionFail -> Should be notification (should also partially successful give alert?)
    RequestFail -> Should be banner
    DockFail -> Is this a failed mission? Might be both banner and notification if the robot gets stuck in the field and require assistance
    RequestDock -> Should be banner (if aborting a mission because of this it will get covered by mission fail as well)
    DismissDock -> Should be banner
    DockSuccess  -> Should be banner

TODO:
1. Remove existing notifications (not banners)
2. Implement notification for failed missions with timestamp, sorted by newest first
3. Remove banner for failed missions

nicholasdalhaug:


nicholasdalhaug:
Next steps:
1. Check all banners and notifications and see that things work
2. Consider changing how the backend sends alert to make it simpler and more consistent.

nicholasdalhaug:
Looks like this:
image

nicholasdalhaug:
I think we might want to simplify some behavior still, but I am not sure how. Maybe related to local storage and handling of recently failed missions.

frontend/src/components/Alerts/AlertNotification.tsx
```diff
@@ -0,0 +1,82 @@
...
+    cursor: ${(props) => (props.$clickable ? 'pointer' : 'auto')};
+`
+
+const SpaceBetween = styled.div`
+    display: flex;
+    justify-content: space-between;
+    align-items: flex-start;
+`
+
+const SmallButton = styled(Button)`
```
Eddasol, line 21: I think there are some built in buttonIcons if we use standard sizes 16 or 24. Can be considered

frontend/src/components/Alerts/AlertsBanner.tsx
```diff
@@ -1,16 +1,16 @@
...
 import { phone_width } from 'utils/constants'
+import { Alert } from 'models/Alert'
 
 const StyledCard = styled.div`
     display: flex;
     width: 100vw - 10px;
     height: auto;
     min-height: 65px;
-    padding: 6px 15px 2px 32px;
+    padding: 0.5rem 1.5rem 0.5rem 1.5rem;
```
Eddasol: Mixing between rem and px in same style. Consider using one of them

Eddasol: Can simplify to
`padding: 0.5rem 1.5rem;`

frontend/src/components/Header/AlertIcon.tsx
```diff
@@ -14,30 +16,20 @@ const Circle = styled.div`
...
+    flex-direction: column;
+    gap: 10px;
 `
 
-export const AlertIcon = () => {
-    const { listAlerts } = useAlertContext()
+interface Props {
+    installation: Installation
+}
+export const AlertIcon = ({ installation }: Props) => {
```
Eddasol, line 29: Do we need props when its just 1?

frontend/src/components/Header/AlertIcon.tsx
```diff
@@ -50,6 +42,16 @@ export const AlertIcon = () => {
         setIsAlertDialogOpen(false)
     }
 
+    const onNotificationClose = (index: number) => {
+        removeNotification(index)
+    }
+
+    const handleMissionClick = (missionId: string) => {
+        // The mission route is nested under :installationCode, so the prefix is required.
```
Eddasol: Is this comment necessary?

frontend/src/contexts/AlertContext.tsx
```diff
@@ -0,0 +1,281 @@
...
+                addNotification,
+                removeNotification,
+                clearAllNotifications,
+            }}
+        >
+            {children}
+        </AlertContext.Provider>
+    )
+}
+
```
Eddasol, line 269: This file have lots of comments. Are all of them necessary

Eddasol (approved):
LGTM :)

PR #2952, opened 2026-09-04
Used cached robot data when handling info update

andchiind:
This is not only a way to reduce repetitive database calls, but also allows us to add more info to these telemetry messages without having that affect performance/costs. 

Note that the cache is only used in this eventhandler, as well as the pressure and battery eventhandlers, so it won't affect the robot values read in elsewhere.

- [ ] A test has been written

### DIFF
--- a/backend/api/EventHandlers/MqttEventHandler.cs
+++ b/backend/api/EventHandlers/MqttEventHandler.cs
@@ -245,11 +245,38 @@ namespace Api.EventHandlers
             }
         }
 
+        private async Task<Robot?> GetCachedRobotByIsarId(string isarId)
+        {
+            if (!_cache.TryGetValue(isarId, out Robot? cachedRobot) || cachedRobot == null)
+            {
+                var robot = await RobotService.ReadByIsarId(isarId);
+
+                if (robot == null)
+                    return null;
+                var cacheEntryOptions = new MemoryCacheEntryOptions
+                {
+                    AbsoluteExpirationRelativeToNow = TimeSpan.FromDays(1),
+                };
+                _cache.Set(isarId, robot, cacheEntryOptions);
+                cachedRobot = robot;
+            }
+            return cachedRobot;
+        }
+
+        private async void UpdateCachedRobot(Robot robot)
+        {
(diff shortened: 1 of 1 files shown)


backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -664,7 +693,7 @@ private async void OnIsarTaskUpdate(IsarTaskMessage task)
 
         private async Task<(string, string)?> GetRobotInstallationCodeAndId(string robotIsarId)
         {
-            if (!_cache.TryGetValue(robotIsarId, out (string, string)? installationCodeAndId))
+            if (!_cache.TryGetValue(robotIsarId, out Robot? cachedRobot) || cachedRobot == null)
```
markusmoen: Is robotIsarId and isarId the same? If it is, is there a reason to not use GetCachedRobotByIsarId here?

backend/api/EventHandlers/MqttEventHandler.cs
```diff
@@ -296,6 +323,8 @@ ref updatedFields
                 if (updatedFields.Count < 1)
                     return;
 
+                UpdateCachedRobot(robot);
+
                 await RobotService.Update(robot);
```
markusmoen: Do we risk the db update failing resulting in mismatched data between cache and db?

markusmoen (approved):

backend/api/EventHandlers/MqttEventHandler.cs
andchiind: Not really, but I think I prefer it this way tbh, since I don't see a need to couple the functions and the number of lines saved is quite small. But it's a good observation.

backend/api/EventHandlers/MqttEventHandler.cs
andchiind: Yes, potentially. I will rewrite it so that the database update comes first.

backend/api/EventHandlers/MqttEventHandler.cs
andchiind: Good point, that looks cleaner

PR #2970, opened 2026-09-08
Fix livestream activation and bounded camera recovery

Christdej:
## Summary

Always fetch fresh media configuration when a camera viewing session starts or retries. The backend endpoint activates the robot publisher; connecting to LiveKit with a still-valid cached token alone does not. Remove frontend use/persistence of the legacy `mediaConfigs` credential cache.

- Give each viewed robot one reference-counted connection, released when its last camera view leaves. Keep room creation outside React state updaters and guard stale config/connection completions, duplicate viewers, and StrictMode replay.
- Bound activation, room connection, and waiting for video to 30 seconds. Retry up to three attempts with 2s/4s backoff, then expose manual Retry. Replenish the budget after 30 seconds of uninterrupted video so later independent outages recover without making rapidly flapping streams retry forever.
- Recover from lost tracks, disconnection, and both media and signaling reconnection events. Track subscriptions by LiveKit publication SID instead of comparing publication IDs with browser media-track IDs. Preserve other working cameras when one disappears.
- Keep the Camera section visible on both mission and robot pages with English/Norwegian Connecting, Reconnecting, and Stream unavailable states and a Retry action.

## Diagnosis and scope

The cached-token activation bypass was identified in code; it was **not directly proven from the affected user's browser**. This change closes that failure path and the independently identified no-track/loading and lifecycle gaps. It does not claim that browser cache alone caused the incident.

Related publisher-side recovery: equinor/isar-anymal#55. The PRs complement each other and require no API contract change. This PR changes only the frontend. No production changes, deployment, or robot commands were performed.

## Validation

- `pnpm test --run tests/components/Contexts/MediaStreamContext.test.tsx`: 25 passing regression cases covering cached tokens, empty connected rooms, finite/manual retries, later independent outages, partial/all track loss, both native reconnect paths, repeated reconnect events preserving deadlines, rejected/pending config and connections, stale callbacks, robot changes, shared ownership, provider cleanup, StrictMode, and localization.
- `pnpm build`: passes, including TypeScript compilation. Existing dependency/chunk-size warnings remain (dashjs ESM/CommonJS and large bundles).
- ESLint on changed TypeScript files: no errors; the existing context/hook Fast Refresh export warning remains.
- Changed-file Prettier and `git diff --check`: pass.
- Code review identified the separate LiveKit signaling-reconnect event; it is now handled and covered.

## Review follow-up

Moved the livestream tests into `frontend/tests` and included that directory in TypeScript, lint, and formatting coverage. Simplified `startAttempt` by separating retry scheduling and attempt-bound event handling, clarified deadline helper names and timer types, and replaced duplicated README timings with links to named code constants. Recovery behavior is unchanged.

## Behavior and limitations

The frontend deliberately stops automatic retries after exhaustion; longer outages require Retry or leaving/re-entering the camera view. Receiving a video track is the readiness signal, not decoded-frame monitoring or an inventory of every expected camera. Partial camera loss therefore does not tear down remaining working feeds. Legacy stored credentials are ignored, not migrated or deleted. Pending backend requests cannot be aborted by the existing API client, but late results are ignored and attempts are bounded.

- [x] A test has been written


frontend/tests/components/Contexts/MediaStreamContext.test.tsx
```diff
@@ -0,0 +1,412 @@
+import { act, StrictMode } from 'react'
```
andchiind, line 1: I think we should keep tests in folders separate from the source code

frontend/src/components/Contexts/MediaStreamManager.ts
```diff
@@ -0,0 +1,195 @@
...
+                () => {
+                    connection.retryTimer = undefined
+                    void startAttempt(robotId, connection)
+                },
+                RETRY_DELAY_MS * 2 ** Math.max(0, connection.attempts - 1)
+            )
+        }
+
+        // Bounds config retrieval, room connection, and waiting for a video publisher.
+        const armTimeout = () => {
```
andchiind: I'm not really sure I understand what this function does

frontend/src/components/Contexts/MediaStreamManager.ts
```diff
@@ -0,0 +1,195 @@
...
+
+const ATTEMPT_TIMEOUT_MS = 30_000
+const MAX_ATTEMPTS = 3
+const RETRY_DELAY_MS = 2_000
+const STABLE_VIDEO_MS = 30_000
+
+interface Attempt {
+    room?: Room
+    timeout?: ReturnType<typeof setTimeout>
+    stableTimer?: ReturnType<typeof setTimeout>
```
andchiind: I'm not really sure I understand this type hinting

frontend/src/components/Contexts/MediaStreamManager.ts
```diff
@@ -0,0 +1,195 @@
...
+        clearTimeout(attempt.stableTimer)
+        if (attempt.room) disconnect(attempt.room)
+    }
+
+    const publish = (robotId: string, connection: Connection, status: MediaStreamState['status']) => {
+        connection.state = { status, streams: [...(connection.attempt?.tracks.values() ?? [])] }
+        onChange(robotId, connection.state)
+    }
+
+    const startAttempt = async (robotId: string, connection: Connection) => {
```
andchiind: Is there any way to simplify this function?

frontend/README.md
```diff
@@ -41,6 +41,23 @@ To run the frontend in Docker, run the following from the repository root:
...
+viewing session or retry always fetches fresh media configuration from the backend: this also activates
+the robot's publisher, which connecting with a cached LiveKit token alone does not do. The frontend
+no longer reads or writes the legacy `mediaConfigs` credential cache.
+
+Activation, room connection, and waiting for video have a 30-second deadline. Failed attempts retry
+with 2- and 4-second backoff, up to three attempts, then display **Stream unavailable** with a manual
+**Retry** action. Losing the last camera or a stalled native reconnect also has a 30-second grace
+period; a disconnected room starts recovery after 2 seconds. Partial camera loss does not interrupt
+remaining cameras. After 30 seconds of uninterrupted video, the retry budget is replenished for a
+later independent outage; short-lived tracks do not reset it.
```
andchiind: It's nice to have this in the README, but if we want to refer to values, should we instead make them configurable and refer to the config values? At the moment there is a tight coupling between these magic numbers and the README, which might become inconsistent in the future.

andchiind (changes requested):

frontend/tests/components/Contexts/MediaStreamContext.test.tsx
Christdej: Moved the livestream regression tests to `frontend/tests/components/Contexts/MediaStreamContext.test.tsx` and switched their imports to the existing aliases. TypeScript, lint, formatting scripts, and the Makefile now include `tests`, so relocating the file does not drop tooling coverage. Left the existing unrelated tests in place to keep this PR focused.

frontend/src/components/Contexts/MediaStreamManager.ts
Christdej: Split retry scheduling into `scheduleRetry` and attempt-bound callbacks into `createAttemptHandlers`, with a named `registerRoomEvents` helper. `startAttempt` now handles initialization, deadline startup, config retrieval, handler registration, and connecting. The stale-attempt guards and cleanup behavior remain in place; this is a behavior-preserving refactor rather than a new abstraction framework.

frontend/src/components/Contexts/MediaStreamManager.ts
Christdej: Introduced `type TimerHandle = ReturnType<typeof setTimeout>` and used `TimerHandle` for all three timer fields. `ReturnType` derives the handle type from `setTimeout` (a number in browsers, a different handle in Node), while `?` means no timer may be active. The alias makes the fields easier to read without changing timer behavior.

frontend/src/components/Contexts/MediaStreamManager.ts
Christdej: Renamed this to `startRecoveryDeadlineIfNeeded`, paired with `clearRecoveryDeadline`. It starts a deadline only when one is not already running: repeated reconnect events must not keep postponing failure. Added a comment explaining that invariant and a regression case that emits another reconnect event partway through the deadline.

frontend/README.md
Christdej: Removed the duplicated numerical values from the README and linked to the named recovery-policy constants in `MediaStreamManager.ts`. The README now explains bounded retries, backoff, and budget replenishment without needing to mirror the current timings. Kept these as code-level settings rather than adding environment variables, since there is no current requirement for deployment-specific tuning.

andchiind (approved):
NIce, looks much better now

PR #3000, opened 2026-09-17
Fix mission updates overwriting completed task status

eivindsjovold:
## Summary
Fix a race that can leave a mission marked **Successful** while its task remains **In progress** and the completed-task count shows **0/1**.

MQTT mission and task handlers run concurrently. Previously, saving a mission with `context.Update(missionRun)` wrote its detached entity graph, including stale task statuses and timestamps. A task completion committed between the mission read and save could therefore be overwritten.

- Save tracked changes rather than rewriting the mission graph, and re-read persisted data before broadcasting the mission update.
- Preserve intentional failure/cancellation behavior and surface update exceptions instead of returning apparent success.
- Remove excluded tasks by ID during scheduling rather than replacing tracked tasks with detached copies.
- Document the concurrency constraint and add deterministic PostgreSQL regression coverage.

## Validation
- Reproduced the original behavior: five regression cases failed with **Expected: Successful; Actual: InProgress**.
- With the fix, all 18 targeted concurrency, mission service, MQTT mission/task, and scheduling tests passed.
- CSharpier and `git diff --check` passed.

## Operational note
This prevents future stale writes; it does not automatically repair already affected mission/task records. Any correction to existing records should be based on confirmed task outcomes.

- [x] A test has been written


backend/api/Services/MissionRunService.cs
```diff
@@ -208,22 +213,22 @@ public bool IncludesUnsupportedInspectionType(MissionRun missionRun)
...
+        private async Task<MissionRun> Update(MissionRun missionRun)
         {
-            if (missionRun.Robot is not null)
-            {
-                context.Entry(missionRun.Robot).State = EntityState.Unchanged;
-            }
-            context.Entry(missionRun.InspectionArea).State = EntityState.Unchanged;
-
-            var entry = context.Update(missionRun);
+            // Save only tracked changes; updating the graph can overwrite concurrent task updates.
```
andchiind, line 218: This comment seems redundant

backend/api/Services/MissionRunService.cs
```diff
@@ -558,14 +563,13 @@ public async Task<MissionRun> UpdateMissionRunProperty(
 
             try
             {
-                await Update(missionRun);
+                return await Update(missionRun);
             }
             catch (InvalidOperationException e)
             {
                 logger.LogError(e, "Failed to update {missionRunName}", missionRun.Name);
+                throw;
```
andchiind, line 571: This breaking change doesn't seem relevant to the rest of the commit/PR.

PR #3004, opened 2026-09-18
Render inspection videos from resolved media SAS

aeshub:
## Summary

- **Symptom:** missions producing `.mp4` inspection results (`AcousticMeasurement`, `ThermalVideo`) never rendered in the frontend — no inspection-result button appeared in the task row, and the video dialog showed only a dead play-icon placeholder.
- **Cause:** SARA populates exactly *one* of `anonymizedSAS` / `visualizedSAS` depending on the analysis chain. The `anonymize` chain sets only `anonymizedSAS`; the `copy-raw-to-visualized` passthrough used for videos sets only `visualizedSAS` (the visualized lookup in SARA's `AnalysisDto` filters with `!Equals("anonymizer")`). The video player (`InspectionReportImage.tsx`) and the inspection-result button (`TaskTable.tsx`) both read `anonymizedSAS`, which is always null for videos — even though `inspectionRecordToInspectionData` already used the correct `visualizedSAS ?? anonymizedSAS` precedence to derive the file type. Display and type detection simply disagreed.
- **Fix:** resolve the media URL once into a new `mediaSAS` field, assigned from the *exact same* local expression that already feeds `sasURLToFileType`, so the two can no longer diverge by construction. Display sites now read `mediaSAS`. The analysis dialog additionally branches on `fileType` to render the existing `VideoPlayer` instead of forcing an `.mp4` into an ``, and its thumbnail strip shows the video placeholder rather than a broken-image icon.
- **No regression for images:** for the `anonymize` chain `visualizedSAS` is null, so `mediaSAS === anonymizedSAS` — behaviour is identical for every image path that works today. `anonymizedSAS` and `visualizedSAS` are both retained; the change is purely additive. Verified with `pnpm build`, `pnpm lint` (0 errors), `pnpm prettier_check`, and `pnpm test --run` (40/40 passing).
- ⚠️ **This is only half of the fix.** SARA currently never creates an analysis for these inspections at all, so there is no SAS to render: `DefaultAnalysisByInspectionTypeAndExtension` is dead config — `GetDefaultAnalysis` only consults `DefaultAnalysisByInspectionType` (`{"Image": ["anonymize"]}`), introduced in equinor/sara@5d3255b637e9d359a8e1ddd9346abd725215f105. The companion SARA change is required before videos actually appear end-to-end.


frontend/src/models/InspectionRecord.ts
```diff
@@ -78,6 +78,12 @@ export interface InspectionData {
     visualizedSAS?: string
     fileType: FileType
     anonymizedSAS?: string
+    // The SAS URL the media should actually be rendered from. SARA populates
+    // exactly one of visualizedSAS/anonymizedSAS depending on the analysis
+    // chain: the anonymize chain sets only anonymizedSAS, while the
+    // copy-raw-to-visualized passthrough used for videos sets only
+    // visualizedSAS. Resolving once here keeps display in step with fileType.
```
eivindsjovold, line 85: maybe a bit extensive comment?

eivindsjovold (approved):
SEe comments, but lgtm

frontend/src/models/InspectionRecord.ts
aeshub: Good point — reduced to a single line in efc7a85.

PR #3006, opened 2026-09-19
Fix feedback dialog overflow

### LINKED ISSUE #2999: Feedback text goes outside of dialog in small screens
**Describe the bug**
When you open the inspection report dialog in a laptop screen the feedback message is not visible. 

**To Reproduce**
Open dev in a small laptop screen (mine is 14")

**Screenshots**

Image

yogixlogic:
Fixes #2999

## Summary

- Allow the inspection result dialog content to scroll vertically.
- Keep the feedback section accessible when the dialog content exceeds the available screen height.

## Testing

- Reproduced the issue using an inspection result with feedback.
- Verified the dialog at a 1366 × 768 viewport.
- Confirmed that the feedback text and buttons remain accessible without horizontal overflow.
- Ran Prettier and ESLint.
- Ran `corepack pnpm build` successfully.


Eddasol (approved):
I'm not a fan of how the feedback gets hidden behind scroll on smaller screens but its better then overflow

PR #3012, opened 2026-09-22
Replace default seed data with a deliberate test scenario suite

aeshub:
Rebuilds the data seeded into the Flotilla database at startup, which is also used by the local Tilt stack in `robotics`. Assumes an empty database.

- Keep only Kårstø (`KAA`) and Northern Lights (`NLS`), with deliberate inspection areas instead of numbered placeholders.
- Seed no robots or mission runs: robots register through ISAR, and run history comes from actual missions. Inspection areas must be assigned after robot registration.
- Provide one named mission per scenario covering every SARA analyzer and sensor type. Single-inspection missions avoid grouped analyses, while distinct tags keep SARA alarms separate.
- Include a mission in an area without a robot to exercise the scheduling restriction.
- Remove seed-specific tests and retain only essential source comments; the PR contains one seed-data commit.

Companion PRs: equinor/robotics#63 and equinor/isar-robot#438.

Exercised end-to-end on the local Tilt stack against all three branches: robots registered and received inspection areas, the blocked-area mission was rejected, and the CLOE and media scenarios delivered their inspections. CO2 ran through the timeseries path and does not create a SARA inspection record. Audio includes the duration required by ISAR.

Known limitations: Fencilla and thermal analyzers reject SARA `main`'s payload due to pre-existing version skew. Alarm severities require the result-value model from equinor/sara#574. Three isar-robot fixtures remain missing and fall back to substitute images with warnings.

- [ ] A test has been written


oysand (approved):
LGTM

